Issue #86, opened 2026-09-01
[Epic] Model aliases, access control, and routing posture


kolatts:
Composes with #81 (gateway epic). Sets FoundryGate's routing posture from the September 2026 classifier/router investigation.

**Plan:** `plans/25-model-aliases-routing.md`

## Adopted

1. **Model alias map** — the gateway exposes stable virtual model names (`sonnet`, `haiku`, `opus`, `gpt-codex`); an inbound policy fragment (before `llm-token-limit`) rewrites them to real deployment names and selects the matching backend pool. Deployments rotate underneath via Bicep with no developer env-var changes. Maps live in per-product named values (`fg-model-map-{productId}`), editable by the control plane via Management API — no policy redeploy.
2. **Per-product model access control** — the alias map is also the allowlist: an alias absent/blocked in the product's map returns 403 `model_not_permitted` in the API's native error schema. Deterministic, fails loud, zero added latency.
3. **Failover-only dynamic routing** — backend pools + circuit breakers (#83) remain the *only* routing that changes a request's destination, and only within the same model+version.
4. **Optional content safety** — `llm-content-safety` (`shield-prompt="true"`) as an opt-in per-product fragment for less-trusted populations; default off for internal developer products. Streaming caveat: set `enforce-on-completions="false"` for CLI tenants.

## Explicitly rejected (don't relitigate)

- **Foundry Model Router for CLI traffic** — OpenAI-schema only (unreachable from Claude Code's Anthropic path) and nondeterministic model selection destroys prompt-cache continuity.
- **APIM semantic caching** — agentic conversations have near-zero cross-request semantic duplication, and a false-positive hit injects a stale answer into a tool-use loop.
- **Gateway-side prompt classifiers** — a serial LLM hop taxes every streamed request and breaks the harness contract that the responding model equals the pinned model.

## Acceptance

- [ ] `sonnet` alias resolves and completes through the Anthropic front door
- [ ] Alias absent from product map → 403 `model_not_permitted` (native error schema)
- [ ] Retargeting an alias via named-value update takes effect without policy redeploy
- [ ] `llm-token-limit` still counts tokens correctly after the `set-body` rewrite
- [ ] Pool selection follows the alias map (verified via `ApiManagementGatewayLogs`)

---
_Generated by [Claude Code](https://claude.ai/code)_

kolatts:
[...]

Issue #82, opened 2026-09-01
Apply llm-token-limit policies: per-developer TPM caps and monthly token quotas via tiered APIM products


kolatts:
Implement real-time enforcement in APIM policy (part of #81):

- Model quota tiers as APIM **products** (Standard / Power / Unlimited), each carrying an `llm-token-limit` policy with `counter-key="@(context.Subscription.Id)"`, `token-quota` + `token-quota-period="Monthly"` (403 on exhaustion) and a per-developer `tokens-per-minute` cap (429 + Retry-After) to keep any one developer from saturating shared deployment TPM.
- **PoC first**: verify whether `token-quota` / `tokens-per-minute` accept policy expressions. If yes, per-user arbitrary values can be read from cache (`cache-lookup-value`) and tiers collapse to one product; if no, tier-products are the design and FoundryGate's quota resolution resolves to a tier.
- Emit `remaining-quota-tokens` / `remaining-tokens` headers so CLIs and the `/me` page can show live budget state.
- Verify token counting behavior on the Anthropic Messages path (v2 tiers required), including how cache-read/cache-creation tokens are counted vs Claude's `usage` block — document divergence from Foundry's cache-aware ITPM accounting.
- Policies live in Bicep alongside the APIM instance (see #13 / #43).

Acceptance:
- [ ] Developer exceeding monthly quota gets 403 at the gateway within one request of the limit (no sync lag)
- [ ] Developer within quota but bursting past their TPM cap gets 429 with Retry-After
- [ ] Both behaviors verified on the Anthropic (`/anthropic/v1/messages`) and OpenAI (`/openai/v1/responses`) paths
- [ ] Expression-support PoC result documented in `plans/24-apim-genai-gateway.md`

kolatts:
PoC answered by live validation (PR #87): `token-quota` REJECTS policy expressions — APIM returns "Expression return type 'System.Int32' is not allowed". Per-user arbitrary quotas via one policy are off the table; **tiers-as-products is the design**. Also validated: quota counter survives policy updates, and raising the quota literal mid-period unblocks a 403'd developer instantly with consumed history intact — the approval flow works exactly as this epic assumes.

kolatts:
[...]

Issue #60, opened 2026-06-04
[Epic] Foundry model deployment provisioning


kolatts:
Admins can list, create, and delete Azure AI Foundry model deployments from the FoundryGate UI without touching the Azure portal. Uses `Azure.ResourceManager.CognitiveServices` with the Container App Managed Identity (`Cognitive Services Contributor` on the Foundry resource). Because APIM already covers all Foundry routes, a new deployment is immediately accessible via existing user subscription keys.

**Plan:** `plans/20-foundry-provisioning.md`

## Sub-issues
- [ ] #61 — Implement /foundry/deployments API endpoints using Azure.ResourceManager.CognitiveServices
- [ ] #62 — Build admin /foundry page with MudDataGrid, create dialog, and delete confirmation

kolatts:
**Direction update (Sep 2026 research — see #81 and `fable-refactor.md`):** facts to bake into this epic:

- `sku.capacity` on a deployment = **thousands of TPM** (capacity 25 → 25K TPM) — the create form should say so.
- Deployment capacity is **PATCH-able at runtime** → add a "resize" action alongside create/delete for live TPM rebalancing between deployments.
- **32 deployments per Foundry resource**; creates must be serialized (concurrent creates 409).
- Claude deployments need the `modelProviderData` attestation block (org name, country, industry) — the create dialog needs those fields for `format: 'Anthropic'` models; model *version* selects hosting (`1` = Anthropic-hosted, `2` = Azure-hosted).
- Current catalog to offer: claude-opus-5, claude-opus-4-8, claude-sonnet-5, claude-sonnet-4-6, claude-haiku-4-5; gpt-5-codex family (incl. gpt-5.1-codex-max/-mini, gpt-5.3-codex) for Codex CLI. claude-sonnet-4-5 is legacy — don't default to it.
- Newly created deployments become reachable through the gateway only if they're added to the right **backend pool** (#83) — provisioning flow should update pool membership.

---
_Generated by [Claude Code](https://claude.ai/code)_

kolatts:
Both sub-issues closed: #61 via PR #131 (list/create/delete over ARM with create-once safety; Claude deployments are infra-owned) and #62 via PR #176 (admin `/foundry` page). Follow-ups: #125 (live validation), #126 (Anthropic create via SDK), #130 (capacity PATCH — in the current backend sweep), #173 (model/SKU catalogue). Closing the epic. — orchestrator, 2026-09-02

PR #87, opened 2026-09-01
Gateway data plane Bicep, live validation, plain-language landing page, model-alias plan

kolatts:
## What this is

Session 2 of the September 2026 strategic pass (task 10 of `fable-refactor.md`): the deployable gateway data plane, empirically validated against the Imagile Paid subscription, plus the docs/landing overhaul and the routers/classifiers investigation.

## Changes

- **`infra/`** — subscription-scope Bicep for the full data plane: Foundry accounts per region + model deployments (Anthropic `modelProviderData` attestation, create-once semantics, `createModelDeployments` re-run switch), APIM StandardV2, backend pool with 429 circuit breakers, Anthropic/OpenAI front doors, `llm-token-limit` + `llm-emit-token-metric` policies, LLM logging pipeline, scalability tags (`fg-role`, `fg-region-role`).
- **`plans/25-model-aliases-routing.md`** (issue #86) — adopt model aliases + per-product allowlists + failover-only routing; explicitly reject Foundry Model Router / semantic caching / prompt classifiers for CLI traffic.
- **Landing page** — plain-English explainer with demo dashboard (fake data) and a "This is real" section showing unedited live-gateway captures; new generated nebula (WebP, −94% weight). Progressive docs ladder + wire-verified CLI setup. Repo `CLAUDE.md` pins these as invariants.
- **`fable-refactor-log.md`** — full decision log (D-001…D-009) and empirical findings (E-001…E-010).

## Live validation results (rg-foundrygate-test, torn down after)

| Verified | Result |
|---|---|
| OpenAI front door e2e | 200 via gateway; exact token accounting non-streaming |
| Auth | no/bad key → 401 both paths |
| Per-dev TPM | 429 + Retry-After; per-subscription isolation confirmed |
| **Native monthly quota** | **403 "try again in 29 days…" in real time**; mid-period raise unblocks instantly |
| #82 PoC | `token-quota` rejects expressions → tiers-as-products is the design |
| Codex CLI | real session e2e; requires `env_http_headers = { "api-key" = … }` |
| Claude Code | wire-captured: sends `x-api-key`; passes gateway auth + policy chain |
| Claude deployment e2e | blocked by subscription-wide Anthropic create failures post-first-delete (E-007) — platform-side; documented with reproduction timeline |


kolatts:

*(posted as a comment; GitHub won't let the PR author file a formal Request Changes review)*

The validation work is the real deal — the wire captures alone were worth the PR. But the headline feature is a no-op in the committed template: the monthly quota was proven live through the Management API and never made it back into the policy XML. The docs promise a 403 the infra can't produce. That's technically correct marketing. Which is the worst kind of correct.

---

### Issues

- **Major** `infra/modules/ai-gateway.bicep` ~L140/L195 — `replace(…, '__MONTHLY_TOKEN_QUOTA__', …)` targets a token that exists in neither policy XML, so monthly budgets fail open to unlimited spend while CLAUDE.md, why-foundrygate, and the landing page all present the 403 as shipped behavior.
- **Major** `infra/modules/ai-gateway.bicep` L36/L88 — `listKeys().key1` puts the master Foundry keys in ARM deployment history and backend objects, the system-assigned MI sits unused, and the accounts are publicly reachable with local auth on — anyone holding that key skips every meter this product exists to enforce. Consider `authentication-managed-identity` + `disableLocalAuth`.
- **Minor** `infra/policies/openai-api.xml` L21 — retry against a single backend with no pool just hammers an already-throttled deployment and adds latency to every failure.
- **Minor** `infra/policies/anthropic-api.xml` L20 — `backend-id="foundry-anthropic-pool"` by string convention (hence the manual `dependsOn`) while the OpenAI policy gets a replacement token; pick one mechanism.
- **Minor** `infra/main.bicep` L146 — outputs discard the workspace id, product id, API URLs, and APIM principal id a control plane needs on day one; `ai-gateway.bicep` already computes three of them and they get thrown away.
- **Minor** `infra/modules/apim.bicep` L43 — App Insights at 100% fixed sampling duplicates GatewayLogs ingestion and scales linearly with agent traffic; consider 10% + allErrors.
- **Minor** `infra/modules/monitoring.bicep` L11 — 30-day retention under a calendar-month reconciliation window leaves no margin; a late run has already lost the start of the period.
- **Minor** `infra/parameters/test.bicepparam` — omits `createModelDeployments` (defaults true), so re-running the shipped parameter file re-PUTs Anthropic deployments into the exact Failed state the comments warn about.
- **Minor** `infra/main.bicep` L28-L61 — the model-placement contract ("everywhere" or "index 0", plus one global create bool) has no path to add a model or place one in a subset of regions without restructuring; if deployment lifecycle is intentionally the control plane's job, the params should say so.
- **Minor** `infra/policies/*.xml` — `buffer-request-body="true"` holds ~87KB per in-flight request on a single StandardV2 unit, and response buffering defaults are a question mark for SSE — what happens to streaming latency under load?
- **Minor** `infra/policies/*.xml` — the two policy files are ~85% identical with no shared fragment; the next header rename lands in one and silently drifts the other. Fine to defer to the #86 fragment work — but say so.
- **Minor** `docs-site/…/architecture/feasibility.mdx` L110 — still says the subscription key header is `api-key` and "both CLIs work with native env vars"; this PR's own wire capture says otherwise on both counts.
- **Minor** `docs-site/…/architecture/overview.mdx` L90 — "Quota exhaustion only suspends" survived the rewrite and contradicts the rest of the same page.
- **Nit** `infra/main.bicep` L101 — the region-abbreviation ternary chain needs a source edit per region; a map with a fallback reads better.
- **Nit** `infra/modules/ai-gateway.bicep` L13 — a param suppressed with `#disable-next-line` and a "reserved" comment is a param that wants deleting.
- **Nit** `docs-site/src/pages/index.astro` L671 — the one asset that hardcodes `/foundry-gate/` while everything else in the file uses `${base}`.
- **Nit** `fable-refactor-log.md` L64 — the live subscription GUID reads identically redacted.

kolatts:
## Response to review — all addressed in 4710da1

Fair catches, especially the first one. Point by point:

**Major — quota no-op:** Fixed and live-verified. The policies now carry a `__QUOTA_ATTRS__` marker that Bicep fills with `token-quota` / `token-quota-period="Monthly"` / `remaining-quota-tokens-header-name` when a quota is configured (empty when 0). Redeployed the test gateway from the committed template: `x-fg-remaining-quota: 4988453` comes back on live traffic. Root cause for the record: the quota was proven through a runtime Management API PUT during testing and never folded back into the repo XML — exactly the drift the docs invariants exist to catch, caught by review instead. The `token-quota`-rejects-expressions constraint is now documented next to the var.

**Major — credential posture:** Switched to `authentication-managed-identity` in both policies + a new `foundry-rbac.bicep` granting APIM's system identity Cognitive Services User per account; backends carry no credentials, so nothing lands in ARM deployment history. `disableLocalAuth: true` (new param, default on) closes the bypass — live-verified: the gateway completes requests over MI while a direct call to Foundry with the account key now gets 403. `publicNetworkAccess` stays Enabled deliberately: StandardV2 APIM reaches backends over public endpoints, and with local auth off there's no credential to steal — network isolation is a Premium/VNet conversation for a later tier.

**Minors, in order:**
- OpenAI retry: removed — you're right, it had nowhere to go. `buffer-response="false"` set on both paths for SSE.
- Pool backend-id: now `__ANTHROPIC_POOL_ID__` replaced with the Bicep resource name — implicit dependency, manual `dependsOn` deleted.
- Outputs: `main.bicep` now emits gateway URLs, product id, workspace id, APIM principal id, and RG name.
- App Insights sampling: 10% + allErrors; the 100% pipeline is GatewayLlmLogs, which is the accounting source of truth.
- Retention: 62 days — two full calendar months.
- `test.bicepparam`: pins `createModelDeployments = false` with a first-run note.
- Model-placement contract: documented as day-0-only in `main.bicep`; lifecycle is the control plane's job (#60/#64), which is also why the create bool has no add-a-model path — that's intentional, now stated.
- Buffering: `buffer-request-body="true"` stays on the Anthropic path only (the pool retry needs it) and is documented as a capacity input; response buffering is off.
- Policy duplication: deferring the shared fragment to the #86 alias work, which introduces `policyFragments` anyway — both files now carry a mirror-the-change comment pointing there.
- `feasibility.mdx` / `overview.mdx` / plan 24: stale `api-key` and "only suspends" claims corrected to the wire-verified reality.
- `estimate-prompt-tokens="false"` (your streaming question): deliberate. Streaming traffic is estimated APIM-side regardless of the setting, non-streaming gets exact post-response settlement with `false`, and overshoot is bounded by one request's `max_tokens`. Documented trade-off, happy to revisit with #88's cache-counting data.

**Nits:** region ternary → lookup map with fallback; dead `appInsightsLoggerId` param deleted; nebula path uses `${base}`; GUID redacted in the log (it lives on in git history — it's targeting data, not a secret, so no rotation equivalent applies).

Verification after fixes: docs build 11 pages clean; full `az deployment sub create` from the committed template + params → Succeeded; live checks — MI-authenticated 200 with quota header, direct-key 403, Anthropic path `DeploymentNotFound` (auth chain healthy, deployment tracked in #88).

PR #90, opened 2026-09-01
feat: .NET 10 solution scaffold (#20, #21)

kolatts:
## Summary

- Establishes `FoundryGate.sln` and the nine-project solution layout under `src/` per `CONVENTIONS.md` §Solution structure: `FoundryGate.Domain`, `FoundryGate.Data`, `FoundryGate.Api`, `FoundryGate.Functions`, `FoundryGate.Web`, `FoundryGate.Database`, `FoundryGate.Cli`, `FoundryGate.Tests.Predeployment`, `FoundryGate.Tests.Postdeployment` (CONVENTIONS.md supersedes `plans/01-scaffold.md`'s and issue #20's older seven/four-project lists, per its own stated precedence).
- Every project targets `net10.0`, `ImplicitUsings`/`Nullable` enable, `TreatWarningsAsErrors` true. Project references follow the dependency flow: `Domain` has zero deps; `Data` → `Domain`; `Api`/`Functions`/`Cli` → `Data` + `Domain`; `Web` → `Domain` only (hard boundary); `Tests.Predeployment` → `Api`/`Data`/`Domain`/`Functions`; `Tests.Postdeployment` → `Domain` only.
- Each project is an entry-point skeleton only (minimal `Program.cs`, demo pages/weather/counter removed from the Blazor WASM template) — no entities, endpoints, or feature code, per scope.
- `FoundryGate.Database` is a `.sqlproj` (`Microsoft.Build.Sql/2.2.0`) building a `.dacpac`; confirmed it builds cleanly standalone and as part of the solution on Windows, so it's included in the `.sln`'s Debug/Release build configuration (unlike imagile-app's default it doesn't need to be excluded — no cross-platform break observed).
- Root tooling: `global.json` (pins .NET 10, `rollForward: latestFeature`), `NuGet.config` (nuget.org only), `.editorconfig` (copied from imagile-app's authoritative one, C# style rules intact), `Directory.Build.props` (version-prop-only, modeled on imagile-app — kept minimal so it doesn't leak `TargetFramework`/`Nullable` into the sqlproj), extended `.gitignore` (`.azurite/`), `docker-compose.yml` (SQL Server on 3433, Azurite), minimal `README.md`.
- One trivial placeholder xUnit test per test project so the pipeline has something green.

## Verification

1. `dotnet build FoundryGate.sln` — **zero errors, zero warnings**:
   ```
   Build succeeded.
       0 Warning(s)
       0 Error(s)
   ```
[...]

Issue #25, opened 2026-06-04
Define enums, shared constants, and paged-result wrapper in FoundryGate.Domain


kolatts:
Part of #3.

Add `QuotaRequestStatus` enum (`Pending`, `Approved`, `Rejected`) and any other enums needed by DTOs. Add a generic `PagedResult<T>` wrapper (items + totalCount + page + pageSize) used by all list endpoints. Add `ValidationConstants` with max-length and range constraints shared between API validation attributes and Blazor form validation.

kolatts:
Implemented and merged in PR #91 alongside #24 — PagedResult/PagedRequest, enums, constants, ApiError.

Issue #24, opened 2026-06-04
Define request/response DTOs for all API endpoint groups in FoundryGate.Domain


kolatts:
Part of #3.

In `FoundryGate.Shared/Models/`, create strongly-typed DTOs grouped by domain: `UserDto`, `UserProfileDto` (for /users/me), `SetQuotaRequest`, `GroupDto`, `GroupDetailDto`, `CreateGroupRequest`, `QuotaAllocationDto`, `QuotaIncreaseRequestDto`, `SubmitIncreaseRequest`, `ReviewRequestBody`, `KeyInfoDto`, `SyncResultDto`, `AuditLogEntryDto`, `DashboardSummaryDto`, and `ConfigEntryDto`. Use nullable reference types throughout.

kolatts:
Implemented and merged in PR #91 (full review cycle: 2 Majors + 1 Minor + 4 Nits addressed; 38/38 tests green).

Issue #20, opened 2026-06-04
Initialize .NET 10 solution with five projects and folder structure


kolatts:
Part of #1.

Create `foundrygate.sln` and four .NET 10 class library / web projects: `FoundryGate.Api` (ASP.NET Core 10 Web API), `FoundryGate.Data` (class library), `FoundryGate.Shared` (class library targeting `netstandard2.1` or `net10.0`), and `FoundryGate.Web` (Blazor WASM). Add a `global.json` pinning `sdk.version` to .NET 10. Add project references: Api → Data + Shared; Web → Shared. Verify `dotnet build` is green from the repo root.

kolatts:
Implemented and merged in PR #90 (full review cycle; 9-project .NET 10 solution, all verification gates green).

PR #91, opened 2026-09-01
feat: domain contracts — DTOs, enums, constants, paging (#24, #25)

kolatts:
## Summary

Implements #24 (request/response DTOs) and #25 (enums, shared constants,
paged-result wrapper) — populates `src/FoundryGate.Domain` (zero dependencies,
per CONVENTIONS.md) with every contract type the API and Blazor WASM client
will share.

- **`Common/`** — `PagedResult<T>` + `PagedRequest` (with `Clamp()` for
  bounds-safe paging), and `ApiError` (an RFC 7807 mirror, not a reference to
  `Microsoft.AspNetCore.Http.ProblemDetails`, so `FoundryGate.Web` can
  deserialize error responses without pulling in an ASP.NET Core dependency).
- **Enums** — `QuotaRequestStatusType` (`Requests/`) and `ModelProviderType`
  (`Config/`), each stored `int`-backed per CONVENTIONS.md.
- **`Constants/`** — `SystemConfigurationKeys` (the eight seeded keys from
  plans/02-data-layer.md), `RoleNames`/`PolicyNames` (`FoundryGate.Admin`),
  `ValidationConstants` (max-length/range bounds shared between API validation
  attributes and future Blazor form validation), and `AuditActions` (see
  decisions below).
- **Per-area `<Area>/Contracts/*Contracts.cs`** — `Users`, `Groups`, `Quota`,
  `Requests`, `Keys`, `Config`, `Audit`, `Dashboard`. `UserProfileResponse`
  (`/me`) embeds the quota gauge, masked API key, and a new
  `GatewayConnectionInfo`/`ModelAliasInfo` CLI-config block so `/me` can render
  the "Configure your CLI" panel (gateway base URL, Anthropic/OpenAI paths,
  the model aliases from plans/25-model-aliases-routing.md).

## Decisions

- **Layout deviates from plan-03's one-file-per-DTO list**: mirrors
  `imagile-app`'s `Imagile.App.Domain.<Area>.Contracts` style instead (one
  `Contracts.cs` per area; enums live directly under `<Area>/`), per this
  task's explicit imagile-alignment direction (mandatory reading item #5).
  DTO *names* still match plan-03 (`UserResponse`, `CreateGroupRequest`, etc.)
  — only the file grouping changed.
- **`AuditLog.Action` stays a `string` constants class, not an enum.** Spec
  §3.1 describes it as free-form text (`"quota.approved"`, `"key.rotated"`)
  and the column isn't `int`-backed; an enum would force a Domain change (and
  a DB migration under the "enums stored as int" convention) every time a new
  admin action needs auditing. `Constants/AuditActions.cs` documents this
  reasoning inline and defines one constant per admin-mutating endpoint in the
  spec, with the two exact strings the spec itself uses
  (`quota.approved`, `key.rotated`, `quota.monthly-reset`) called out.
- **No `UserStatusType`/`KeyStatusType` enums, and no `QuotaLevel` enum
  either** (plan-03 lists `QuotaLevel` alongside the enums it expects `#25` to
  add). Spec's `User` entity has no status column (just `IsActive`), key state
  is fully derived from `ApimSubscriptionId` nullability, and spec has no
  "quota level/tier" concept anywhere in §3 — nothing in the domain model
  resolves to a small closed set of named levels the way quota is actually
  computed (§3.2's precedence chain over `IsUnlimited`/`MonthlyTokenQuota`
  values, not a level). All three would be state invented past what the spec
  defines. `ApiKeyResponse.IsProvisioned` is a plain derived `bool` instead.
- **`PagedRequest` carries no `DataAnnotations`.** It's meant to degrade
  gracefully via `Clamp()`, not 400 the caller for an out-of-range page/size —
  putting `[Range]` on it would fight that behavior under ASP.NET Core's
  automatic model-state validation. Every actual domain-input request DTO
  (`CreateGroupRequest`, `SubmitQuotaIncreaseRequest`, ...) does carry
  attributes.
- **Reconciliation vs. enforcement framing.** Per this pass's direction
  (quota enforcement is real-time at the gateway; usage numbers surfaced to
  the UI are reconciliation data), `QuotaAllocationResponse` and
  `DashboardSummaryResponse` carry doc comments making explicit that
  `TokensUsed`/`PercentUsed`/etc. reflect the last Log Analytics sync (spec
  §5.4), not live enforcement — the gateway's own `x-fg-remaining-*` response
  headers are the real-time source of truth. `IsHardStopped` is preserved
  (admin-facing only) since spec §5.3 sets it on key revocation, not on quota
  exhaustion, so it doesn't actually imply suspension-on-exhaustion.
- **Int PK + Guid Unique.** DTOs expose the data layer's `int` PK plus a
  `{Entity}Unique` `Guid` on `User`, `Group`, and `QuotaIncreaseRequest` — the
  entities judged most likely to be referenced externally (audit trails,
  cross-system correlation). `QuotaAllocation`, `ApiKey`, `AuditLog`, and
  `SystemConfiguration` DTOs stick to the int PK (or, for config, its natural
  string key) since nothing external needs to reference them by a stable id.
- **Plaintext key delivery is a distinct type.** `ApiKeyResponse` (masked,
  returned by every *display* endpoint) and the new `ApiKeyRevealResponse`
  (plaintext, returned ONLY by rotate/provision) are separate records rather
  than one type with a sometimes-null `PlaintextKey` field, so the type system
  documents the one-time-delivery semantic instead of relying on callers to
  remember which endpoints populate the sensitive field.

## Review round 2

Addressed all findings from the first review pass
([comment](https://github.com/kolatts/foundry-gate/pull/91#issuecomment-5497995349)):
plaintext key delivery shape (`ApiKeyRevealResponse`), the architecture test
inverted to an exact allowlist, `EntraGroupId` validation + exposure on
`GroupResponse`, `ApiError.Type`/`Title` made nullable, the email-length RFC
citation corrected, `HasApiKey`→`IsApiKeyProvisioned` naming unified, and the
`QuotaLevel` decision documented above. Point-by-point response posted as a
PR comment. Test count below reflects the added coverage (38, up from 31).

## Verification

- [x] `dotnet build FoundryGate.sln` — 0 warnings, 0 errors
  (`TreatWarningsAsErrors` is on repo-wide, including `GenerateDocumentationFile`
  + `CS1591`-suppressed XML-doc analysis)
- [x] `dotnet test src/FoundryGate.Tests.Predeployment` — 38/38 passing, including:
  - `DomainArchitectureTests` — now an exact allowlist: asserts the compiled
    `FoundryGate.Domain` assembly's `GetReferencedAssemblies()` contains only
    `System.Collections`/`System.ComponentModel.Annotations`/`System.Runtime`
    (captured from a clean build), so any new reference at all — not just the
    previous seven-prefix blocklist — fails the test; doc comment now states
    precisely what direct-IL-reference reflection does and doesn't catch
  - `PagedRequestTests` / `PagedResultTests` — clamping and `TotalPages` logic
  - `SystemConfigurationKeysTests` — the eight seeded keys, no duplicates
  - `RequestDtoValidationTests` — `DataAnnotations` actually fire (via
    `Validator.TryValidateObject`) on representative request DTOs, including
    `EntraGroupId`'s new GUID-shape validation
  - `ApiErrorTests` — a payload missing `type`/`title` deserializes to `null`
    without throwing, using `JsonSerializerDefaults.Web` (case-insensitive
    camelCase, matching how `HttpClient` JSON extensions actually deserialize)
- [x] `dotnet format FoundryGate.sln --verify-no-changes` — clean (exit 0)
- plans/03-shared-dtos.md verification:
  - [x] `dotnet build` passes with zero warnings
  - [x] `FoundryGate.Domain` has no dependencies on ASP.NET Core or EF Core packages
  - [x] All request DTOs have at least one validation attribute
  - [x] `PagedResult<T>` is usable from both the API and the Blazor WASM project
    (plain positional record, no non-portable dependencies; `FoundryGate.Web`
    and `FoundryGate.Api` both already reference `FoundryGate.Domain`)

## Scope notes

- Domain + its own Predeployment tests only — no `FoundryGate.Data` entities,
  `FoundryGate.Api`, or `FoundryGate.Web` touched, per instructions (a parallel
  agent owns the `Data` entities). New test files are namespaced under
  `FoundryGate.Tests.Predeployment.Domain` / a `Domain/` subfolder to avoid
  colliding with that agent's own Predeployment test additions.
- `Foundry/` (deployment DTOs, epic #20) intentionally not touched — plan-03
  notes it's out of scope for #24/#25.
- Current branch already contained PR #90's `.NET 10` scaffold merge and
  matched `origin/main` exactly at start, so no additional merge/pull was
  needed.

### DIFF
--- a/src/FoundryGate.Domain/Audit/Contracts/AuditContracts.cs
+++ b/src/FoundryGate.Domain/Audit/Contracts/AuditContracts.cs
@@ -1,0 +1,29 @@
+namespace FoundryGate.Domain.Audit.Contracts;
+
+/// <summary>One audit log entry (spec &#167;3.1 <c>AuditLog</c>). GET /audit (paged, filterable).</summary>
+/// <param name="AuditLogId">Surrogate int PK.</param>
+/// <param name="ActorUserId">The admin (or system job) that performed the action. Null for system-initiated actions with no actor, e.g. an unattended background job.</param>
+/// <param name="ActorDisplayName">Denormalized for display; null when <paramref name="ActorUserId"/> is null.</param>
+/// <param name="Action">One of <see cref="Constants.AuditActions"/>, or a future action not yet constant-ized — the column is free text by design.</param>
+/// <param name="TargetType">e.g. "User" | "Group" | "Request" (spec &#167;3.1). Null when the action has no single target.</param>
+/// <param name="TargetId">Identifier of the affected record, as a string (spec &#167;3.1 keeps this loosely typed since <paramref name="TargetType"/> varies). Null when the action has no single target.</param>
+/// <param name="Details">Opaque JSON blob (spec &#167;3.1); the API does not interpret it, only the UI's audit viewer renders it.</param>
+/// <param name="OccurredDate">When the action happened.</param>
+public record AuditLogEntryResponse(
+    int AuditLogId,
+    int? ActorUserId,
+    string? ActorDisplayName,
+    string Action,
+    string? TargetType,
+    string? TargetId,
+    string? Details,
+    DateTimeOffset OccurredDate);
+
+/// <summary>Filter parameters for GET /audit. Bind alongside <see cref="Common.PagedRequest"/> via a separate <c>[FromQuery]</c> parameter.</summary>
+public record AuditLogQuery(
+    int? ActorUserId,
+    string? Action,
+    string? TargetType,
+    string? TargetId,
+    DateTimeOffset? FromDate,
+    DateTimeOffset? ToDate);
--- a/src/FoundryGate.Domain/Common/ApiError.cs
+++ b/src/FoundryGate.Domain/Common/ApiError.cs
@@ -1,0 +1,23 @@
+namespace FoundryGate.Domain.Common;
+
+/// <summary>
+/// RFC 7807-shaped error envelope. FoundryGate.Api's single <c>IExceptionHandler</c>
+/// (CONVENTIONS.md) serializes ASP.NET Core's <c>ProblemDetails</c> onto the wire in
+/// this shape; Domain defines its own POCO mirror — rather than referencing
+/// <c>Microsoft.AspNetCore.Http.ProblemDetails</c> directly — so FoundryGate.Web (which
+/// references Domain only, per the Blazor WASM hard boundary) can deserialize error
+/// responses without pulling an ASP.NET Core dependency into the WASM client.
+/// </summary>
+/// <param name="Type">A URI reference identifying the problem type (RFC 7807 <c>type</c>).</param>
+/// <param name="Title">Short, human-readable summary of the problem.</param>
+/// <param name="Status">The HTTP status code generated for this occurrence.</param>
+/// <param name="Detail">Human-readable explanation specific to this occurrence.</param>
+/// <param name="Instance">A URI reference identifying the specific occurrence (typically the request path).</param>
+/// <param name="Errors">Per-field validation failures, when <see cref="Status"/> is 400.</param>
+public record ApiError(
+    string Type,
+    string Title,
+    int Status,
+    string? Detail = null,
+    string? Instance = null,
+    IReadOnlyDictionary<string, string[]>? Errors = null);
--- a/src/FoundryGate.Domain/Common/ApiError.cs
+++ b/src/FoundryGate.Domain/Common/ApiError.cs
@@ -8,16 +8,29 @@ namespace FoundryGate.Domain.Common;
 /// references Domain only, per the Blazor WASM hard boundary) can deserialize error
 /// responses without pulling an ASP.NET Core dependency into the WASM client.
 /// </summary>
-/// <param name="Type">A URI reference identifying the problem type (RFC 7807 <c>type</c>).</param>
+/// <remarks>
+/// <see cref="Type"/> and <see cref="Title"/> are nullable even though the API's own
+/// responses always populate them: this record's other job is deserializing whatever
+/// error payload actually comes back over the wire (a foreign/malformed response, a
+/// proxy's own error page, a future API version that dropped a field) — System.Text.Json
+/// materializes a missing JSON property as <c>null</c> regardless of the property's C#
+/// nullable-reference-type annotation, so declaring them non-nullable here would just be
+/// a compile-time promise the deserializer doesn't keep at runtime.
+/// </remarks>
+/// <param name="Type">A URI reference identifying the problem type (RFC 7807 <c>type</c>). RFC 7807's own default is <see cref="DefaultType"/> when no more specific value is known.</param>
 /// <param name="Title">Short, human-readable summary of the problem.</param>
 /// <param name="Status">The HTTP status code generated for this occurrence.</param>
 /// <param name="Detail">Human-readable explanation specific to this occurrence.</param>
 /// <param name="Instance">A URI reference identifying the specific occurrence (typically the request path).</param>
 /// <param name="Errors">Per-field validation failures, when <see cref="Status"/> is 400.</param>
 public record ApiError(
-    string Type,
-    string Title,
+    string? Type,
+    string? Title,
     int Status,
     string? Detail = null,
     string? Instance = null,
-    IReadOnlyDictionary<string, string[]>? Errors = null);
+    IReadOnlyDictionary<string, string[]>? Errors = null)
+{
+    /// <summary>RFC 7807's own default <c>type</c> value when no more specific problem type URI applies.</summary>
+    public const string DefaultType = "about:blank";
+}
--- a/src/FoundryGate.Domain/Common/PagedRequest.cs
+++ b/src/FoundryGate.Domain/Common/PagedRequest.cs
@@ -1,0 +1,34 @@
+namespace FoundryGate.Domain.Common;
+
+/// <summary>
+/// Paging parameters accepted by every "(paged)" list endpoint (spec &#167;4). Bind
+/// directly from the query string via <c>[FromQuery]</c>.
+/// </summary>
+/// <remarks>
+/// Deliberately carries no <c>System.ComponentModel.DataAnnotations</c> attributes.
+/// An out-of-range <see cref="Page"/> or <see cref="PageSize"/> here is a client being
+/// sloppy about pagination, not a validation failure worth a 400 — call
+/// <see cref="Clamp"/> before using the values in a query and the request degrades to a
+/// safe default instead. Request/response DTOs that represent actual domain input (a
+/// quota justification, a group name, ...) DO carry validation attributes; see the
+/// per-area <c>Contracts</c> types.
+/// </remarks>
+/// <param name="Page">1-based page number. Values below 1 clamp up to 1.</param>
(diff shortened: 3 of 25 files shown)


kolatts:

Solid pass overall — coverage against spec §4 is complete, the reconciliation framing is exactly right, and all six declared decisions hold up (including Clamp-not-validate on `PagedRequest`, which is the correct call as long as the eventual `ToPagedAsync` helper is the thing that calls `Clamp()` so no controller has to remember). Two things block it: the key contract can never actually hand a developer their key, and the zero-dependency test enforces less than its own doc comment claims.

### Issues

- **Major** src/FoundryGate.Domain/Keys/Contracts/ApiKeyContracts.cs — `ApiKeyResponse` declares itself the response for rotate and provision but carries only `MaskedKey`, so no contract in the PR can ever deliver the plaintext key that cli-setup.mdx's "Configure your CLI" panel requires; the moment key provisioning is implemented you'll be forced into positional-record churn — add a `PlaintextKey` (null except on rotate/provision) or a separate `KeyRotationResponse` now.
- **Major** src/FoundryGate.Tests.Predeployment/Domain/DomainArchitectureTests.cs — `GetReferencedAssemblies()` returns only the assemblies Domain's own IL directly references, so the doc comment's claim of catching transitive sneak-ins is false, and the seven-prefix blocklist misses everything else (Newtonsoft, FluentValidation, whatever); invert it to an allowlist (fail on any reference outside `System.*`/`netstandard`/`mscorlib`/`Microsoft.CSharp`) and "zero dependencies" is actually enforced instead of "not these seven".
- **Minor** src/FoundryGate.Domain/Groups/Contracts/GroupContracts.cs — `CreateGroupRequest.EntraGroupId` is the one string in the PR with no `[StringLength]` (while `ValidationConstants.EntraObjectIdMaxLength` sits right there), and no response DTO ever returns `EntraGroupId`, so an admin can link a group to Entra and then never see or verify which group it's linked to.
- **Nit** src/FoundryGate.Domain/Common/ApiError.cs — `Type`/`Title` are non-nullable but System.Text.Json will happily materialize null into them from a payload missing those fields, and the WASM client deserializing foreign error responses is exactly the consumer this type exists for; either make them nullable or give them defaults.
- **Nit** src/FoundryGate.Domain/Constants/ValidationConstants.cs — RFC 5321 §4.5.3.1.3 caps the forward-path at 256 octets, making the real address maximum 254; 320 is the 64+1+255 folk number, harmless as a bound but the citation oversells it.
- **Nit** src/FoundryGate.Domain/Users/Contracts/UserContracts.cs — `UserResponse.HasApiKey` and `ApiKeyResponse.IsProvisioned` are the same derived fact under two names; pick one before Web bakes both into components.
- **Nit** — the PR body's decisions cover dropping plan-03's `UserStatus`/`KeyStatus` enums but silently drops `QuotaLevel` too; the drop is right (spec has no such concept), it just deserved a line.

kolatts:
Pushed `da1b4c3` addressing every finding. Point-by-point:

**Major — plaintext key delivery (`ApiKeyContracts.cs`)**
Added `ApiKeyRevealResponse` (`PlaintextKey`, `MaskedKey`, `ApimSubscriptionId`, `IssuedDate`) as a distinct record, returned only by `POST /keys/me/rotate`, `POST /keys/{userId}/rotate`, and `POST /keys/{userId}/provision`. `ApiKeyResponse` stays masked-only and its doc comment now says explicitly it's never the response for those three endpoints. Doc comments on both types spell out the one-time-delivery contract: the API must not persist the plaintext value anywhere re-servable, and the UI should show it once and fall back to the masked form after. No positional-record churn needed later — the shape's there now.

**Major — allowlist inversion (`DomainArchitectureTests.cs`)**
Replaced the seven-prefix blocklist with an exact allowlist. Captured the real set via a throwaway `ITestOutputHelper` probe against a clean build of `FoundryGate.Domain.dll`: `System.Collections`, `System.ComponentModel.Annotations`, `System.Runtime` (all `10.0.0.0`, standard BCL token) — that's genuinely everything Domain's IL references today. The test now asserts `GetReferencedAssemblies()` is a subset of exactly that list, so Newtonsoft, FluentValidation, or anything else fails just as loudly as EF Core would've. Deliberately not a `"System."` prefix match, per your point that a prefix still lets things through — went with the tighter reading over your own suggested prefix set.

Also rewrote the doc comment to stop overclaiming: it now says plainly that this only catches assemblies Domain's own IL *directly* references (the AssemblyRef metadata table), that a `PackageReference` sitting unused in the project file with zero call sites won't show up here at all, and that the empty `.csproj` is the actual "zero dependencies" guarantee — this test is the CI-enforced cross-check that code never starts exercising something Domain didn't declare, not a substitute for reading the `.csproj`.

**Minor — `EntraGroupId` (`GroupContracts.cs`)**
Added `[StringLength(ValidationConstants.EntraObjectIdMaxLength)]` + `[RegularExpression(ValidationConstants.GuidPattern)]` on `CreateGroupRequest.EntraGroupId` (new `GuidPattern` const in `ValidationConstants`). Kept it as a validated string rather than switching to `Guid?`, to stay consistent with `User.EntraObjectId`, which is also a GUID-shaped string elsewhere in this domain — didn't want two different representations of "Entra object id" in the same PR. `GroupResponse` now carries `EntraGroupId` too (used by `GroupDetailResponse` as well, since it wraps `GroupResponse`), so an admin can actually see the link they set. Added validation test cases (malformed strings fail, a real GUID passes).

**Nit — `ApiError.Type`/`Title` nullable**
Made both `string?`. Added `ApiError.DefaultType = "about:blank"` (RFC 7807's own default) since making `Type` nullable removed the obvious place to look for that convention. Added `ApiErrorTests` proving a payload missing both fields deserializes to `null` without throwing, using `JsonSerializerDefaults.Web` (case-insensitive camelCase — what `HttpClient`'s JSON extensions actually use), plus a full round-trip test.

**Nit — email length RFC citation**
Corrected: RFC 5321 §4.5.3.1.3 caps the total reverse-/forward-path at 256 octets, netting out to a 254-char deliverable address, not 320. Rewrote the comment to attribute 320 to the actual source — the 64 (local-part, §4.5.3.1.1) + 1 + 255 (domain) folk sum — and note it's deliberately looser than the real RFC bound so it never rejects a valid address. Value unchanged (320 is still a fine bound), only the citation was wrong.

**Nit — `HasApiKey`/`IsProvisioned` naming**
Renamed `UserResponse.HasApiKey` → `UserResponse.IsApiKeyProvisioned` to match `ApiKeyResponse.IsProvisioned` exactly. Grepped for other call sites first — none existed yet (Web isn't built), so this was a clean rename.

**Nit — `QuotaLevel` drop undocumented**
Added to the PR body's "No `UserStatusType`/`KeyStatusType` enums" decision (now also covers `QuotaLevel`): plan-03 lists it alongside the two others, spec §3 has no quota-tier concept for any of the three — quota is computed via the §3.2 precedence chain over `IsUnlimited`/`MonthlyTokenQuota` values, never a named level.

---

Verification re-run clean: `dotnet build FoundryGate.sln` (0 warnings/errors), `dotnet test src/FoundryGate.Tests.Predeployment` (38/38, up from 31 — added `ApiErrorTests` plus new `EntraGroupId` validation cases), `dotnet format FoundryGate.sln --verify-no-changes` (exit 0). PR body updated with the new decisions and refreshed verification counts. Not merging.

Issue #77, opened 2026-06-04
Add FoundryGate.Database .sqlproj with initial SQL table files for all entities


kolatts:
Part of #76.

Create `src/FoundryGate.Database/FoundryGate.Database.sqlproj` using SDK `Microsoft.Build.Sql/2.0.0` targeting `SqlAzureV12DatabaseSchemaProvider`. Add one `.sql` file per entity table under `dbo/Tables/` — contents must exactly match the EF schema from epic #2. The `.sqlproj` is never hand-edited after initial creation; it is kept in sync with EF migrations via `foundrygate db compare`. Output dacpac to `artifacts/`. Add project to `FoundryGate.sln`. The `build-dacpac` CI job (plan #22) runs `dotnet build` on this project on a `windows-latest` runner.

Issue #32, opened 2026-06-04
Implement five-level quota resolution logic and write to QuotaAllocation


kolatts:
Part of #7.

Extract a `QuotaResolutionService` that implements the five-level precedence: (1) `User.IsUnlimited` → unlimited; (2) `User.MonthlyTokenQuota` is set → use it; (3) any group with `IsUnlimited` → unlimited; (4) max of all group `MonthlyTokenQuota` values; (5) `SystemConfiguration["DefaultMonthlyTokenQuota"]`. The service writes or upserts a `QuotaAllocation` row for the target period. Called from user/group quota changes, monthly reset, and first-login provisioning.

kolatts:
Claimed — implementing together with #33 on branch `claude/quota-resolution-32-33` (quota resolution service → tier-product mapping per the #7 direction update, `QuotaAllocation` schema additions, `IGatewayTierSync` seam).

PR #92, opened 2026-09-01
feat: data layer — entities, DbContext, interceptors, seeding (#22, #23)

### LINKED ISSUE #22: Define EF Core 10 entities and configure all relationships in DbContext
Part of #2.

In `FoundryGate.Data/Entities/`, create the seven entity classes exactly as specified: `User`, `Group`, `GroupMember` (composite PK), `QuotaAllocation` (unique constraint on UserId+Year+Month), `QuotaIncreaseRequest`, `SystemConfiguration` (string PK), and `AuditLog`. In `FoundryGateDbContext` configure all relationships using Fluent API: `GroupMember` composite key, `QuotaAllocation` unique index, and appropriate cascade delete rules. Include `DateTimeOffset`-typed audit fields.

### LINKED ISSUE #23: Create initial EF migration and seed the eight SystemConfiguration defaults
Part of #2.

Run `dotnet ef migrations add InitialSchema` to produce the baseline migration. Add a `HasData` seed in `OnModelCreating` (or a custom `IEntityTypeConfiguration`) for the seven `SystemConfiguration` rows: `DefaultMonthlyTokenQuota`, `ApimResourceId`, `FoundryResourceId`, `EntraTenantId`, `EntraGroupSyncEnabled`, `ResetDayOfMonth`, and `ApimProductId`. Seed values are placeholder strings that forks must override via the admin config UI.

kolatts:
Closes #22
Closes #23

## Summary

Implements #22 (EF Core 10 entities + DbContext) and #23 (SystemConfiguration seeding), adapted from foundrygate-spec.md §3.1 to CONVENTIONS.md and the gateway-centric direction updates on #7/#10.

- **Entities** (`src/FoundryGate.Data/Entities/`): `User`, `Group`, `GroupMember` (composite PK), `QuotaAllocation`, `QuotaIncreaseRequest`, `SystemConfiguration`, `AuditLog` — int identity `{Entity}Id` PKs, `Guid {Entity}Unique` on `User`, `Group`, and `QuotaIncreaseRequest`, `DateTimeOffset` fields suffixed `Date`, enums suffixed `Type`, non-nullable strings defaulting `string.Empty` (one justified exception: `User.EmployeeId`).
- **`AppDbContext`**: plain `DbContext` registration, `ApplyConfigurationsFromAssembly`, co-located `internal sealed IEntityTypeConfiguration` classes only where delete behavior needs Fluent API. Cascade policy: at most one cascade edge per entity (`Group→GroupMember`, `User→QuotaAllocation`, `User→QuotaIncreaseRequest.UserId`); every other FK (including the two extra `User` FKs on `GroupMember` and the two extra on `QuotaIncreaseRequest`) is `NoAction` with a comment.
- **`TimestampInterceptor`**: `SaveChangesInterceptor` using injected `TimeProvider`, driven by small opt-in `ICreatedDate`/`IModifiedDate` interfaces. Registered via `AddFoundryGateData(IServiceCollection, string connectionString)`, wired into `FoundryGate.Api/Program.cs` reading `ConnectionStrings:FoundryGate` (default in `appsettings.Development.json`, pointed at the docker-compose local SQL instance — no secrets, same dev-only password already committed in `docker-compose.yml`).
- **Seeding**: `SystemConfiguration`'s 8 spec-defined defaults via a locally-implemented `IReferenceDataEntity`/`SyncReferenceDataAsync` pattern (see below for why), plus a Bogus-driven `TestDataSeeder` matching the landing page's demo shape (varied quota tiers, one unlimited developer, one pending increase request).
- **Direction-update adaptations**: `QuotaAllocation.TokensUsed` is documented as reconciliation state only (populated from the Log Analytics sync, never read on the request path); `IsHardStopped` is offboarding-only. Enforcement itself is APIM's `llm-token-limit` policy — nothing in this PR gates requests.
- **Tests** (`src/FoundryGate.Tests.Predeployment`): SQLite in-memory harness (`InMemoryDatabaseTest`), `FoundryGateConventionTests` (inherits `Imagile.Framework.EntityFrameworkCore.Testing`'s `DbContextConventionTests`, 3 justified exclusions), local `NamingTests`/`DesignTests` for what the framework base doesn't cover, plus behavior tests: interceptor sets `CreatedDate`/`ModifiedDate` (advancing a fake `TimeProvider` between insert and update so the assertion can actually fail), and `SystemConfiguration` seeding is idempotent and never clobbers an operator's edited value.

## Imagile.Framework package fit (signal requested by the task)

- `Imagile.Framework.Core` / `Imagile.Framework.Configuration` / `Imagile.Framework.EntityFrameworkCore` / `Imagile.Framework.EntityFrameworkCore.Testing`, all pinned to the required **1.0.12**.
- `DbContextConventionTests` (Testing package) matches CONVENTIONS.md well — used as-is, only 3 exclusions needed (`SystemConfiguration`'s string PK, `User.EmployeeId`'s nullable string, `AuditLog.Details`'s unbounded length — see review-fix notes below).
- `Imagile.Framework.EntityFrameworkCore`'s public `IReferenceEntity<T>`/`AddOrUpdateReferenceEntities` **does not match** imagile-app's internal `IReferenceDataEntity<TSelf,TId>`/`SyncReferenceDataAsync` pattern the conventions doc points at: its upsert does a blind `CurrentValues.SetValues(...)` that overwrites every column and ignores `[DoNotUpdate]`. That's wrong for `SystemConfiguration`, whose entire point is that fork operators edit seeded defaults via the admin UI and re-seeding must never clobber that edit. **Implemented locally** in `FoundryGate.Data/Seeding/` instead, reusing only the public, unmodified `Imagile.Framework.Core.Attributes.DoNotUpdateAttribute`.

## Review-fix + #91 reconciliation pass

The first review round ([comment](https://github.com/kolatts/foundry-gate/pull/92#issuecomment-5498081091), 3 Major / 5 Minor / 2 Nits) landed right as #91 (Domain contracts) merged to `main`. This PR now merges `origin/main` and reconciles against #91's now-authoritative DTO shapes, plus fixes every review finding:

- `GroupMember` implements `ICreatedDate` via an explicit interface implementation forwarding to `AddedDate` (kept that name — matching #91's already-merged `GroupMemberResponse.AddedDate` — instead of renaming to the generic `CreatedDate`), so the column is actually populated now.
- `TimestampInterceptorTests`' update-path assertion advances a mutable fake `TimeProvider` between the insert and the update, so it can actually fail if the interceptor stops firing on `Modified`.
- Added `Group.GroupUnique` (+ unique index) to match `GroupResponse`; deleted the duplicate `FoundryGate.Domain.Enums.QuotaRequestStatusType` in favor of #91's `FoundryGate.Domain.Requests.QuotaRequestStatusType`; `SystemConfiguration.Key`/`Value` now use `ValidationConstants.ConfigKeyMaxLength`/`ConfigValueMaxLength` (200/4000); `SystemConfiguration`'s `ModifiedDate` renamed to `UpdatedDate` to match `SystemConfigEntryResponse` (still backs `IModifiedDate` via an explicit interface implementation); `SystemConfigEntryResponse.UpdatedByUserId` and `GroupMemberResponse.AddedByUserId` are now nullable **in Domain** (`main`, via this PR) to match the entities' actual nullability truth — entity wins, DTO follows.
- `QuotaIncreaseRequestUnique` gets a unique index, matching `UserUnique`.
- `ReferenceDataSeeder` passes a `deleteFilter` scoped to `SystemConfigurationKeys.All`, so a fork-added config key survives re-seeding instead of being deleted as an "orphan".
- `ReferenceDataExtensions.SyncReferenceDataAsync` no longer fetches `AsNoTracking()` + re-attaches via `Update()` (which throws once the context already tracks that key); it fetches tracked rows and mutates them in place. The property-copy loop now walks EF's own `IEntityType.GetProperties()` metadata instead of raw CLR reflection, so it can no longer touch a navigation property.
- `TestDataSeeder` takes an injected `TimeProvider` instead of `DateTimeOffset.UtcNow`, and seeds its own `Faker`'s random state via `UseSeed()` instead of mutating the process-global `Bogus.Randomizer.Seed`.
- `AuditLog.Details` is unbounded (`nvarchar(max)` on SQL Server) instead of `StringLength(4000)` — no `[StringLength]`/`HasColumnType` override at all, since `HasColumnType("nvarchar(max)")` is SQL-Server-only syntax that breaks the SQLite test harness.
- Fixed the `usedFraction` comment in `TestDataSeeder` that mislabeled index 4 as the unlimited tier when it's actually index 5.

Follow-ups endorsed in the review, filed per CLAUDE.md ("everything is a GitHub issue"):
- #94 — verify the cascade-delete graph against real Azure SQL once #77's dacpac pipeline exists (SQLite doesn't enforce SQL Server's multiple-cascade-paths restriction).
- #95 — encrypt `ApimSubscriptionKey` at rest (spec §11) before real APIM keys land.

## Verification

- [x] `dotnet build FoundryGate.sln` — 0 Warning(s), 0 Error(s)
- [x] `dotnet test src/FoundryGate.Tests.Predeployment` — Passed: 65, Failed: 0 (28 from this PR + #91's Domain tests merged in, all green together)
- [x] `dotnet format FoundryGate.sln --verify-no-changes` — clean

### plans/02-data-layer.md checklist
- [x] `dotnet build` passes
- [ ] `dotnet ef migrations list` shows `InitialCreate` as applied — **N/A by design**: CONVENTIONS.md explicitly bans EF migrations ("no migrations folder... dacpac pipeline comes in #77"); schema is `EnsureCreated`/dacpac-driven instead. Superseded by the build+test gates above.
- [x] All eight `SystemConfiguration` rows are present after seeding (issue #22/CONVENTIONS.md corrected the spec's "seven" to the actual eight keys listed in both the issue table and spec §3.1 comment block; test asserts this explicitly)
- [x] No orphaned FK constraints or missing indexes — verified via `EnsureCreated` schema build (fails fast on invalid FK/cascade config) + convention tests; no migration output to inspect since there is no migration. Real-DDL cascade verification tracked separately as #94.

## Follow-ups

- No API endpoints, no quota-resolution service logic (explicitly out of scope; #32).
- #94 — Azure SQL cascade-delete verification once #77 lands.
- #95 — `ApimSubscriptionKey` encryption at rest before real keys land.


kolatts:

Score: 78/100. The core is solid — entities track spec §3.1 with the CONVENTIONS adaptations applied correctly, the cascade design is reasoned through and actually avoids the multiple-cascade-path trap (Group→GroupMember, User→QuotaAllocation, User→QuotaIncreaseRequest.UserId, NoAction everywhere else with comments — I traced it, it's clean), and the local `SyncReferenceDataAsync` genuinely honors `[DoNotUpdate]` where the framework package wouldn't. But `GroupMember.AddedDate` is a column nothing ever populates, one interceptor test can't fail, and this PR and #91 are about to merge two different `QuotaRequestStatusType` enums into the same project. Fix the Majors and this ships.

### Issues

- **Major** `src/FoundryGate.Data/Entities/GroupMember.cs` — `AddedDate` is never populated: the entity isn't `ICreatedDate`, no interceptor covers it, and `TestDataSeeder` inserts memberships without setting it, so every row lands as `0001-01-01`. The dacpac flow bakes this in and #91's `GroupMemberResponse` exposes `AddedDate` straight to the UI — either make it `ICreatedDate` (rename to `CreatedDate`; CONVENTIONS wins over the spec's `AddedAt`) or stamp it everywhere.
- **Major** `src/FoundryGate.Tests.Predeployment/Data/TimestampInterceptorTests.cs` — `SaveChanges_sets_ModifiedDate_on_insert_and_update_via_TimeProvider` is vacuous for the update half: the fake clock never moves, so the second assert passes even if the interceptor never fires on Modified. Advance the fake between saves so the test can actually fail.
- **Major** cross-PR (#91 convergence) — two structural collisions: (a) #91's `GroupResponse` carries `Guid GroupUnique` (added per its own review) but this PR's `Group` entity has no `GroupUnique` column; (b) this PR adds `FoundryGate.Domain.Enums.QuotaRequestStatusType` while #91 adds `FoundryGate.Domain.Requests.QuotaRequestStatusType` — two distinct CLR enums for the same concept. Whichever merges second must add the Group column and delete one enum; decide the canonical namespace now.
- **Minor** `src/FoundryGate.Data/Entities/QuotaIncreaseRequest.cs` — `QuotaIncreaseRequestUnique` is documented as a deep-link lookup id but has no `[Index(IsUnique = true)]`, unlike `User.UserUnique`. Lookups scan and uniqueness isn't constrained.
- **Minor** cross-PR (#91 convergence, field-level) — `SystemConfiguration.Key` is `StringLength(100)` / `Value` is `StringLength(500)` vs #91's `ConfigKeyMaxLength = 200` / `ConfigValueMaxLength = 4000` (so `UpdateSystemConfigRequest` validates values the column will reject); entity `ModifiedDate` + `int? UpdatedByUserId` vs DTO `UpdatedDate` + non-nullable `int UpdatedByUserId` (seeded rows have null); `GroupMemberResponse.AddedByUserId` non-nullable vs entity `int?`. Pick one side per field and converge.
- **Minor** `src/FoundryGate.Data/Seeding/ReferenceDataExtensions.cs` — the sync deletes any row not present in seed data by default, so a fork-added `SystemConfiguration` key dies on the next deploy. Pass a `deleteFilter` from `ReferenceDataSeeder` (or make SystemConfiguration insert/update-only) before an operator finds out the hard way.
- **Minor** `src/FoundryGate.Data/Seeding/ReferenceDataExtensions.cs` — latent identity-conflict: `Update(current)` attaches a detached copy and throws if the context already tracks that key; you only dodge it today because every `SystemConfiguration` property is `[DoNotUpdate]` so the branch is unreachable. The reflection copy also includes navigation properties, which happens to be harmless now. Worth hardening before entity #2 uses this.
- **Minor** `src/FoundryGate.Data/Seeding/TestDataSeeder.cs` — naked `DateTimeOffset.UtcNow`, which CONVENTIONS bans outside the interceptor's `TimeProvider`. Dev-only doesn't exempt it; take a `TimeProvider` parameter.
- **Nit** `src/FoundryGate.Data/Entities/AuditLog.cs` — `Details` is a JSON blob capped at `StringLength(4000)`; consider `nvarchar(max)` before something interesting gets truncated.
- **Nit** `src/FoundryGate.Data/Seeding/TestDataSeeder.cs` — the `usedFraction` comment labels index 4 as the unlimited user, but the unlimited tier is index 5; also `Randomizer.Seed` is a process-global mutation.

### Follow-ups (endorsed)

- Azure SQL cascade sanity check post-#77 — endorsed; SQLite won't surface multiple-cascade-path errors, but the design as written should pass real DDL.
- `ApimSubscriptionKey` encryption before real keys land — endorsed as a follow-up, but per CONVENTIONS follow-ups are labeled issues, not PR prose. File it and link it here.
- Nothing wires either seeder into a prod path (Program.cs calls no seeder; CLI epic owns `db deploy → seed-reference → seed-test`) — correct as-is.

kolatts:
Thanks for the thorough pass — merged `origin/main` (#91) first, then fixed everything below.

## Majors

1. **`GroupMember.AddedDate` never populated** — `GroupMember` now implements `ICreatedDate`, but via an explicit interface implementation that forwards to `AddedDate` rather than renaming the property to `CreatedDate`. #91's `GroupMemberResponse.AddedDate` merged to `main` in the meantime and is now the authoritative DTO field name, so keeping the entity's public name aligned with it (while still getting full `TimestampInterceptor` coverage through the explicit `ICreatedDate.CreatedDate` accessor) avoided an unnecessary rename ripple into Domain. Documented on the entity.
2. **Vacuous interceptor update test** — rewrote `SaveChanges_sets_ModifiedDate_on_insert_and_update_via_TimeProvider` into `SaveChanges_advances_ModifiedDate_on_update_but_leaves_insert_time_CreatedDate_alone`: the fake `TimeProvider` is now mutable, advances 3 hours between the insert and the update, and the test asserts `UpdatedDate` (see #3c) moved to the new time while `CreatedDate` stayed at the original insert time. Confirmed it fails if I comment out the interceptor's Modified-state branch.
3. **Cross-PR convergence with #91:**
   - (a) Added `Group.GroupUnique` (+ `[Index(..., IsUnique = true)]`) matching `GroupResponse.GroupUnique`.
   - (b) Deleted `FoundryGate.Domain.Enums.QuotaRequestStatusType`; `QuotaIncreaseRequest.StatusType` now uses #91's `FoundryGate.Domain.Requests.QuotaRequestStatusType` (values line up: Pending=0/Approved=1/Rejected=2).
   - (c) `SystemConfiguration.Key`/`Value` now use `ValidationConstants.ConfigKeyMaxLength`/`ConfigValueMaxLength` (200/4000) instead of local 100/500 literals. `ModifiedDate` renamed to `UpdatedDate` to match `SystemConfigEntryResponse.UpdatedDate` (still backs `IModifiedDate` via an explicit interface implementation, same trick as #1). `UpdatedByUserId` nullability: entity keeps `int?` (a seeded row genuinely has no editor), and I updated `SystemConfigEntryResponse.UpdatedByUserId` to `int?` in Domain (on `main` via #91) to match — noted inline that this is the reconciliation from #92.
   - (e) Same for `GroupMemberResponse.AddedByUserId` → `int?`, matching the entity.

## Minors

- Added `[Index(nameof(QuotaIncreaseRequestUnique), IsUnique = true)]`.
- `ReferenceDataSeeder.SeedAsync` now passes `deleteFilter: c => SystemConfigurationKeys.All.Contains(c.Key)`, so a fork-added config key survives re-seeding.
- `SyncReferenceDataAsync` no longer does the `AsNoTracking()` fetch + `Update(current)` re-attach dance — it fetches tracked rows and mutates in place (EF's change tracker picks up the diff automatically), so there's no detached-copy identity conflict waiting for entity #2. `UpdateProperties` now walks `IEntityType.GetProperties()` (EF's scalar-property metadata) instead of raw CLR reflection over every public property, so navigation properties are structurally excluded rather than "happens to be harmless."
- `TestDataSeeder.SeedAsync` now takes a `TimeProvider` parameter and uses `Faker<T>.UseSeed(randomSeed)` instead of mutating the global `Bogus.Randomizer.Seed`.

## Nits

- `AuditLog.Details` has no `[StringLength]` at all now (which is what makes it map to `nvarchar(max)` on SQL Server) — tried `HasColumnType("nvarchar(max)")` first per your suggestion, but that's SQL-Server-only syntax and SQLite's DDL parser chokes on the literal `max`, which broke the whole convention-test suite (SQLite in-memory harness). Leaving the column unconfigured gets the same unbounded result on both providers. Added a `StringsMustHaveMaxLengthRule` exclusion with that explanation.
- Fixed the `usedFraction` switch comment (index 4 → index 5 is actually the unlimited tier) and scoped the Bogus seed via `UseSeed()` (see Minors).

## Follow-ups (filed)

- #94 — Azure SQL cascade-delete verification once #77's dacpac pipeline exists.
- #95 — `ApimSubscriptionKey` encryption at rest (spec §11) before real keys land.

## Verification

- `dotnet build FoundryGate.sln` — 0 Warning(s), 0 Error(s)
- `dotnet test src/FoundryGate.Tests.Predeployment` — Passed: 65, Failed: 0 (28 from this PR + #91's Domain tests, all green together post-merge)

PR body updated with `Closes #22` / `Closes #23` in the body text (not just commit trailers) and the review-fix summary above.

Issue #83, opened 2026-09-01
Configure backend pools with priority groups and 429-tripped circuit breakers across Foundry deployments


kolatts:
Transient 429 smoothing at the gateway (part of #81):

- Define APIM **backend pools** per model family (e.g. `claude-sonnet-pool`, `gpt-codex-pool`): primary deployment at priority 1, secondary deployment(s) — other region or DataZone Standard — at priority 2.
- **Circuit breaker** rule per backend tripping on 429/5xx with `acceptRetryAfter: true` so trip duration honors the backend's `Retry-After`.
- `retry` policy in the backend section retrying onto the pool on 429 (retry-across-backends, not wait-and-retry) — only between identical model+version deployments.
- Multi-region pools are the Claude substitute for PTU/spillover (both OpenAI-only). Claude quota pools are subscription-level and shared across regions for GlobalStandard, so cross-region pooling multiplies *availability*, and DataZone/Anthropic-hosted variants diversify the pool — document what it does and doesn't buy.
- Bicep: `Microsoft.ApiManagement/service/backends` (single + pool) with circuit breaker rules.

Acceptance:
- [ ] Saturating the priority-1 deployment causes traffic to flow to priority-2 without client-visible errors
- [ ] Tripped breaker recovers per Retry-After; 503 returned only when all pool members are tripped
- [ ] Pool topology documented in the architecture docs

kolatts:
Implemented and merged in PR #93: priority-grouped anthropic pool (region[0] P1, later regions P2, per-account overrides) with the validated 429 circuit breakers. Live failover confirmation consolidated into #88.

Issue #67, opened 2026-06-04
[Epic] CI/CD pipelines — infra deploy, code deploy, and destruction workflows


kolatts:
Seven GitHub Actions workflows covering separated infrastructure and code pipelines, dev→prod promotion gates, and manual-only destruction workflows with typed confirmation. All OIDC, no long-lived secrets. See `plans/22-cicd-pipelines.md` for the full pipeline reference.

**Plan:** `plans/14-github-actions.md` + `plans/22-cicd-pipelines.md`

## Sub-issues
- [ ] #68 — Configure GitHub Environments: dev, production, dev-destroy, prod-destroy
- [ ] #69 — Write infra-deploy.yml: validate → what-if PR comment → deploy dev → deploy prod
- [ ] #70 — Write infra-destroy.yml: typed confirmation guard and environment gate
- [ ] #71 — Write api-deploy.yml: build/test/Docker/Container App (separated from infra)
- [ ] #72 — Write functions-deploy.yml: build/test/publish/Functions deploy
- [ ] #73 — Write ui-deploy.yml: build/publish/Static Web Apps with PR preview
- [ ] #74 — Write docs-deploy.yml: Astro build and GitHub Pages deploy
- [ ] #75 — Write deploy-all.yml: full-stack ordered deploy via workflow_call
- [ ] #79 — Write db-deploy.yml: reusable dacpac deploy via foundrygate CLI

kolatts:
Partial progress: minimal PR CI landed on main (ci.yml — build with warnings-as-errors, Predeployment tests, dotnet format, conditional docs build) as the required status check anchoring the new main branch protection (1 required review, stale-dismiss, no force pushes). The full pipeline shape from plans/22 (deploy workflows, environments, artifact-fed Claude review job) remains this epic's scope.

kolatts:
[...]

PR #93, opened 2026-09-01
feat(infra): tiered quota products, priority pools, model aliases (#82, #83, #86)

kolatts:
Completes the APIM gateway enforcement layer that PR #87 left half-built. Infra/policy only — no `src/` changes.

Closes nothing yet: every issue's acceptance criteria include live-traffic checks, and no APIM instance currently exists (the test stack was torn down at the end of the strategic pass, D-010). See **What still needs a live deploy** below.

## #82 — quota tiers as products

The single `foundrygate` product is gone, replaced by a parameterized `quotaTiers` set (default: `standard` 5M/20K TPM, `power` 20M/40K, `unlimited` no monthly quota/100K). Migration is a clean rename rather than an alias: products *are* tiers now, and the control plane assigns a developer's quota by issuing their APIM subscription against the matching tier product.

`token-quota` accepts literals only (validated live 2026-09-01), so each tier renders its own product policy from a shared template, `infra/policies/product-policy.xml`.

### The product / API split, and why

APIM evaluates inbound as **global → product → API** (an API policy's `<base />` expands the *product* policy). That ordering decides where each concern has to live:

| Scope | Owns | Why |
|---|---|---|
| **Product** (rendered per tier) | model allowlist, `tokens-per-minute`, `token-quota` | **Entitlement** — what this subscription may use and how much. `token-quota` must be a literal, so it has to be rendered per tier anyway; and anything that must run *before* token counting has to sit in the outer scope. |
| **API** (`anthropic-api.xml`, `openai-api.xml`) | credential stripping, managed-identity backend auth, default backend, retry/streaming, token metrics | **Mechanism** — identical for every tier, so it renders once per API. |

`llm-token-limit` is declared in **exactly one** scope. Declaring it at both would silently double-count every request; `scripts/validate-policies.ps1` now asserts that statically.

## #83 — priority-grouped pools

The Anthropic pool derives routing from the `foundryRegions` index: **first region priority 1, every later region priority 2** — spillover shape, not round-robin, so normal traffic stays in-region (latency + Claude prompt-cache affinity) and other regions are standing headroom. Per-account `priority`/`weight` overrides are supported. Circuit breakers (`acceptRetryAfter: true`) are untouched, and the pool resource still carries no `url`/`protocol` (E-009).

Documented in plan 24: multi-region pooling multiplies *availability* against deployment-level throttling, **not** the subscription's token budget — Claude GlobalStandard quota is pooled per-subscription per-model across regions.

## #86 — model aliases + per-product allowlist

New `fg-model-alias` fragment: parse `model` from the body → look it up in the tier's `fg-model-map-{tier}` named value → `403 model_not_permitted` if absent/blocked, otherwise `set-body` to the real deployment name + `set-backend-service` to its pool.

Two design calls worth reviewing:

**The fragment is included from the product policy, not the API policies.** `{{named-value}}` tokens resolve by *literal* name, so a shared fragment cannot compute `fg-model-map-{productId}` at runtime — only the per-tier product policy (rendered once per tier by Bicep) can hand the map in, which it does via a `fgModelMap` variable set immediately before the include. This also satisfies "before `llm-token-limit`" for free, since `llm-token-limit` is itself product-scope: **a blocked model costs the developer no quota.**

**One fragment with a schema branch, not two fragments.** Only the ~6 lines that build the 403 body differ (Anthropic `permission_error` envelope vs the OpenAI `error` object); parse/lookup/rewrite/route is identical. The fragment branches on `context.Api.Id`, and Bicep substitutes the real API name into the comparison so it never depends on a hand-copied magic string. Two fragments would have reintroduced exactly the duplication this PR was asked to remove. The 403 body is built with `JObject` rather than string concatenation, so the attacker-controlled alias is JSON-escaped (and truncated to 128 chars before it is echoed).

The API policies' static `set-backend-service` is **demoted, not deleted**: it now sits *before* `<base />`, making the pool the default route that the alias map overrides per-model. A request the map never touches (e.g. `GET /openai/v1/models`, no `model` in the body) still routes sensibly instead of falling back to a single-region `serviceUrl`.

**Day-0 provisioning change:** `claude-sonnet-4-5` and `claude-opus-4-5` join `pooledModelDeployments`. Every Claude model reachable through an alias must exist in **every** pool member, or a 429 failover turns into a 404. Names/versions were read from `az cognitiveservices model list` in both default regions; capacities sit well inside the subscription's GlobalStandard quota (sonnet-4-5 200, haiku-4-5 80, opus-4-5 40). This does mean more create-once Anthropic deployments on first run (E-007) — `createModelDeployments=false` on every re-run remains mandatory.

## Deferred cleanup from the PR #87 review

The duplication between the two API policy XMLs is folded into `fg-backend-auth` (credential stripping + MI auth) and `fg-token-metrics` (`llm-emit-token-metric`, now carrying a `Product ID` dimension). What stayed duplicated is genuinely different — backend topology and retry semantics — and was left alone on purpose.

## Static verification evidence

No live APIM exists, so verification is static but strict. All four checks below were run on this branch.

**1. Bicep compiles clean**
```
$ az bicep build --file infra/main.bicep
exit=0
```
Only the three expected `BCP081` warnings (no published types for `Microsoft.CognitiveServices/accounts@2026-07-01`, which is pinned deliberately per E-005).

**2. Policy XML — new `scripts/validate-policies.ps1`**

Substitutes every `__TOKEN__` exactly the way `modules/ai-gateway.bicep` does, then asserts: well-formed XML, no surviving placeholder, no *unknown* placeholder, correct root element (`<policies>` vs `<fragment>`), every `include-fragment` id resolves to a fragment file, and `llm-token-limit` declared in exactly one scope.

```
$ pwsh ./scripts/validate-policies.ps1
Checked 6 policy document(s)
  - anthropic-api.xml
  - backend-auth-fragment.xml
  - model-alias-fragment.xml
  - openai-api.xml
  - product-policy.xml
  - token-metrics-fragment.xml

All policy documents are well-formed and fully substituted.
```

The script was negative-tested (injected malformed XML, an unknown token, a second `llm-token-limit`, and a dangling `fragment-id`) — it exits 1 and names each problem.

**3. ARM validate — subscription scope, read-only**
```
$ az deployment sub validate --location eastus2 \
    --template-file infra/main.bicep \
    --parameters infra/parameters/test.bicepparam   # createModelDeployments=false
provisioningState: Succeeded
```

**4. ARM validate + what-if — gateway module, read-only**

Subscription-scope validate does not expand the gateway module's APIM children (they depend on runtime outputs), so the module was validated directly with representative parameters:

- `az deployment group validate` — **Succeeded**, all **30** APIM child resources pass ARM preflight: 3 `policyFragments`, 3 `namedValues`, 3 tier products + 3 product policies + 6 apiLinks, the priority pool, both APIs and their policies.
- `az deployment group what-if` — inspected the rendered property values:
  - pool renders `priority: 1` (eastus2) / `priority: 2` (swedencentral), `weight: 1`
  - `fg-model-map-standard` / `-power` → `sonnet`, `haiku`, `gpt`; `fg-model-map-unlimited` additionally → `opus`; every logical `pool` resolved to a real backend id
  - `standard` product policy renders `tokens-per-minute="20000"` + `token-quota="5000000" token-quota-period="Monthly"`; `unlimited` renders the TPM cap with **no** quota attributes
  - all 8 ARM-rendered policy documents re-parse as well-formed XML
- Neither command left a deployment record behind (checked `az deployment group list` / `sub list`). Nothing was created.

## What still needs a live deploy

None of the three issues' acceptance criteria can close on static evidence. Marked "pending next live deploy" in `plans/24` and `plans/25`:

**#82**
- Monthly quota 403 and TPM 429 + `Retry-After` were proved live on the *pre-tier, API-scope* policy (T4/T5) — they need re-proof now that `llm-token-limit` moved to product scope.
- Each tier product enforces its own literal quota; moving a subscription between tier products changes the enforced budget.
- `llm-emit-token-metric` accepts the `Product ID` dimension.

**#83**
- Saturate priority-1, confirm traffic continues via priority-2 with no client-visible error; breaker recovers per `Retry-After`; 503 only when the whole pool is tripped.

**#86**
- `sonnet` resolves and completes through the Anthropic front door.
- Blocked alias → 403 `model_not_permitted` in the correct schema on each front door.
- **The one mechanism with no offline proof:** that a named value whose JSON contains double quotes substitutes cleanly into the `set-variable value="{{...}}"` attribute. XML parsing happens before substitution, so this should be fine, but it is the highest-risk line in the change and worth checking first on the next deploy.
- Retargeting an alias via a named-value update takes effect without a policy redeploy.
- `llm-token-limit` still counts correctly after the `set-body` rewrite.
- Pool selection follows the alias map (via `ApiManagementGatewayLogs`).

Also unproven offline: APIM's policy *compiler* only runs on a real PUT, so expression-level errors in `fg-model-alias` (allowed-type violations, etc.) would first surface there.

## Follow-ups (not in this PR)

- **Docs** still describe a single `foundrygate` product: `docs-site/.../getting-started/fork-guide.mdx`, `reference/configuration.mdx` (`ApimProductId`), `architecture/overview.mdx`. Left alone to avoid colliding with parallel agents; needs a docs pass once tier naming settles.
- **`cli-setup.mdx` alias documentation** is deliberately deferred — CLAUDE.md requires that page to carry only empirically verified configuration, and aliases have not been exercised against a live gateway.
- **Control plane**: `ApimProductId` (single value) needs to become a tier reference, and the admin endpoint for editing alias maps via the Management API is still unbuilt (#86's last bullet).
- The template output `productId` became `productIds` + `defaultProductId` — any pipeline consuming the old output needs updating.

Refs #82 #83 #86


kolatts:
## Code review — APPROVE

The scope split holds up under an actual trace, which is the thing that mattered here. Effective inbound for a tiered request: subscription validated (pre-policy, `x-api-key`) → API-scope static `set-backend-service` (default) → `<base />` expands the product policy → `fgModelMap` set → alias fragment (403-if-blocked, else body rewrite + `set-backend-service`, making the alias the last writer over the static default) → `llm-token-limit` → back in API scope: credential strip + MI auth → metrics → forward. A blocked model burns no quota, the strip happens after product scope needs nothing from those headers and before forward, and the alias wins routing. I also compiled the module: Bicep emits implicit `dependsOn` from the `.name` references, so both API policies depend on their backends and each product policy on its fragment + named value + product — no first-deploy race. The 403 body is JObject-built with the alias truncated before lookup and echo, so neither injection nor truncation-to-a-permitted-prefix works. Alias map ↔ pool coverage is complete both directions, capacities sit inside the logged quota facts, E-007 is documented where a fork will actually read it, and the leftover `products/foundrygate` references are docs-only (overview.mdx, fork-guide.mdx, foundrygate-spec.md — already flagged as follow-ups; no script, workflow, or output consumes the old id). Plans 24/25 static-vs-pending splits match what was actually verified.

### Issues

- **Minor** `infra/policies/model-alias-fragment.xml` — the alias map is per-tier, not per-API, so a permitted Claude alias sent through the OpenAI front door rewrites and routes `/chat/completions` at the Anthropic pool (and `gpt` through the Anthropic door lands on the OpenAI backend): a confusing backend 404 instead of a clean 4xx. Consider checking the entry's backend family against `context.Api.Id` before the rewrite branch.
- **Minor** `infra/policies/product-policy.xml` — moving `llm-token-limit` and the allowlist to product scope means any subscription not scoped to a tier product (including APIM's built-in all-access subscription) now bypasses both entirely; the old API-scope policy caught everything. Disable the built-in subscription at deploy time or record it as a control-plane invariant.
- **Nit** `scripts/validate-policies.ps1` line 44 — only the populated `__QUOTA_ATTRS__` shape is ever parsed; loop both renderings so the unlimited tier's empty-string render is validated too instead of asserted safe in a comment.
- **Nit** `infra/policies/model-alias-fragment.xml` line 94 — `error.code` isn't part of Anthropic's error envelope (`{type, message}`); harmless extra field, but the comment claims "native error schema", and `x-fg-error` already carries the machine-readable code.

kolatts:
All four items addressed in `8674255`, on top of a merge of latest `main` (`2186524`) so CI runs against current code.

### 1. Minor — non-product subscription bypass ✅

New `infra/policies/require-product-fragment.xml` (`fg-require-product`), included in **both** API policies before their `<base />` — ahead of where product scope would have run — returning 403 when `context.Product == null`, in each API's native envelope (Anthropic `permission_error`, OpenAI `error` object), with `x-fg-error: plan_required`.

**Master subscription decision: policy guard only — I did not add a Bicep resource.** Deactivating it means PUTing `Microsoft.ApiManagement/service/subscriptions/master`, and `scope` is required on that resource. I have no way to confirm the correct `scope` value for a *built-in* subscription without a live APIM, and a wrong guess doesn't fail loudly — it silently rewrites a live built-in subscription's scope on every deploy. That's a worse failure mode than the gap it closes, and the gap is already closed by the guard, which covers strictly more cases (any subscription created outside a tier product later, not just `master`). Recorded as a revisit-on-live-deploy note in `modules/ai-gateway.bicep` and plan 24, plus a live checklist item asserting the master key gets a 403.

The check is `context.Product == null` rather than a tier-name membership test: every product this template creates *is* a tier, and a name list would need substituting into a shared fragment for no added safety.

### 2. Minor — cross-schema alias routing ✅

Alias entries now carry `provider` (`"anthropic"` | `"openai"`) alongside `deployment` and `pool`; all three are required or the entry is treated as blocked. The fragment computes the current front door from the substituted API id and refuses a mismatch:

> `Model 'sonnet' is served by the anthropic front door on this gateway. Send this request to /anthropic instead.`

Both front-door paths are substituted from Bicep vars (`__ANTHROPIC_API_PATH__` / `__OPENAI_API_PATH__`) rather than hard-coded, and those same vars now feed the APIs' `path` properties so they can't drift.

`provider` is deliberately a **separate field from `pool`** rather than derived: they coincide today, but an Anthropic DataZone or secondary pool splits them. Both refusal reasons now share a single deny path (`fgDenyMessage`), so the two error bodies differ only by wire schema, not by logic. Bicep param defaults updated for all three tiers.

### 3. Nit — unlimited-tier render ✅

`scripts/validate-policies.ps1` now runs every document through **both** render variants — `quota-configured` and `quota-disabled (unlimited tier)` (empty `__QUOTA_ATTRS__`) — 14 renders across 7 files. It also asserts the `token-quota` attribute is present in exactly the variant that should carry it, so a silent swap can't pass.

### 4. Nit — Anthropic 403 body ✅

Dropped `error.code` from the Anthropic-shaped body; the native envelope is `{type, error:{type, message}}` with no `code` field. The machine-readable code now travels in the `x-fg-error` header on **both** front doors, which is the only place it can appear consistently across the two schemas. Applied to the new guard fragment too. Verified in the ARM-rendered output: exactly one `code` property per fragment, in the OpenAI branch only.

---

## Re-run static verification

**`az bicep build --file infra/main.bicep`** — exit 0, only the three expected `BCP081` warnings.

**`pwsh ./scripts/validate-policies.ps1`**
```
Checked 7 policy document(s)
  - anthropic-api.xml
  - backend-auth-fragment.xml
  - model-alias-fragment.xml
  - openai-api.xml
  - product-policy.xml
  - require-product-fragment.xml
  - token-metrics-fragment.xml
Render variants: quota-configured, quota-disabled (unlimited tier)  (14 renders)

All policy documents are well-formed and fully substituted.
```

Negative-tested again, including the new assertions:
- Hardcoded a `token-quota` into the template → caught twice: `duplicate attribute name` in the configured render, and `token-quota attribute present=True, expected=False` in the unlimited render.
- Unknown token + dangling `fragment-id` + a stray API-scope `llm-token-limit` → 6 problems reported across both variants, exit 1.

**`az deployment sub validate`** (test.bicepparam, `createModelDeployments=false`) — **Succeeded**.

**`az deployment group validate`** on the module — **Succeeded**, now **31** APIM child resources through ARM preflight (was 30; +`fg-require-product`).

**`az deployment group what-if`** — inspected rendered values:
- every alias entry carries `provider`, e.g. `fg-model-map-unlimited` → `{"gpt":{...,"provider":"openai"},"haiku":{...,"provider":"anthropic"},"opus":{...,"provider":"anthropic"},"sonnet":{...,"provider":"anthropic"}}`
- all **9** ARM-rendered policy documents re-parse as well-formed XML (was 8; +the guard fragment)
- `fg-require-product` renders first in both API policies, before `<base />`, ahead of the default `set-backend-service`
- no placeholder survives in any rendered document; API id and both paths substituted

Neither `validate` nor `what-if` left a deployment record behind (`az deployment group list` / `sub list` re-checked). Nothing was created.

Everything still gated on a live deploy is listed in plan 24 / plan 25 under "pending next live deploy", now including the master-subscription 403 and the wrong-front-door 403.

kolatts:
**CI note:** branch protection on `main` requires the `build-test` status context, but no workflow in the repo emits it — `.github/workflows/` currently holds only `claude-review.yml`, `claude-triage.yml` and `docs.yml`, and `grep -r build-test .github/` finds nothing. Check runs on this head commit are just `review | skipped`, so the PR sits at `mergeStateStatus: BLOCKED` with nothing pending to wait on. That will affect every PR, not this one.

Branch is merged up to current `main` (`6fbc6d5`). I have not added a CI workflow here — that's the #67 CI/CD epic's territory (CONVENTIONS.md specs `ci.yml` as build → Predeployment tests → format check → Claude review) and likely belongs to whoever owns that work, not an infra/policy PR.

Issue #103, opened 2026-09-01
Implement foundrygate db compare (DacFx schema-compare, Windows-only) to regenerate dbo/Tables/*.sql from local docker SQL


kolatts:
Deferred from #100 (option 2). #100 was closed by extending `SchemaParityTests` (option 1), which now checks column types/lengths/precision, IDENTITY, PK shape, FK shape (columns, principal, ON DELETE, composite), index composition/order/direction, and extra indexes/FKs/files — cross-platform, no database required.

What that test can't do is *write* the `.sql` files. Today `src/FoundryGate.Database/dbo/Tables/*.sql` are hand-authored to mirror the EF model, and the parity test is the alarm when they drift. imagile-app's `SchemaComparisonHelpers` workflow (plans/23-database-tooling.md, original design) would instead regenerate them mechanically:

- source: `SchemaCompareDatabaseEndpoint(local docker SQL)` after `foundrygate local setup` (`EnsureCreated`)
- target: `SchemaCompareProjectEndpoint(src/FoundryGate.Database/FoundryGate.Database.sqlproj)`
- `Compare()` + `PublishChangesToProject()`, discarding pure column-reordering noise

## Scope
- `foundrygate db compare` command in `src/FoundryGate.Cli` (`System.CommandLine`), Windows-only guard with a clear message on Linux/macOS
- Uses the already-pinned `Microsoft.SqlServer.DacFx` package
- Not a CI gate — `SchemaParityTests` stays the cross-platform backstop; this is a developer convenience for the "I changed an entity, regenerate the script for me" step

## Why not now
No concrete need yet: seven tables, hand-authored once, and the parity test now catches every drift class we've seen. Pick this up when adding/altering tables by hand becomes a recurring chore.

kolatts:
Picking this up: implementing `foundrygate db compare [--connection-string] [--apply] [--check]` per plans/23-database-tooling.md and the issue scope. Branch `claude/cli-db-compare-103`.

PR #104, opened 2026-09-01
test(data): check SQL types, IDENTITY, PK/FK/index shape in SchemaParityTests

### LINKED ISSUE #100: Close SchemaParityTests' documented gaps and/or implement foundrygate db compare
Part of #76.

## Context

#77 added \`src/FoundryGate.Tests.Predeployment/Data/Conventions/SchemaParityTests.cs\` as a drift alarm between the EF entity model and the checked-in \`FoundryGate.Database/dbo/Tables/*.sql\` files, standing in for real DacFx schema-compare (\`foundrygate db compare\` was never implemented — see plans/23-database-tooling.md's status note). It is a deliberately pragmatic regex-level parser, and its own XML doc lists the gaps:

- No SQL data type/length/precision checking — a column that silently shrinks from \`nvarchar(200)\` to \`nvarchar(50)\` in the \`.sql\` file without a matching entity change is NOT caught.
- No composite-FK support (none exist in the model today, so untested).
- No index column-order/composition validation beyond the leading \`UNIQUE\` flag.
- No detection of an extra index present in the \`.sql\` file but absent from the model.
- Regex parsing over the repo's existing SQL style, not a real T-SQL parser — hand-authored SQL that strays from that style (different bracket/whitespace conventions, multi-table files, etc.) can produce false positives/negatives instead of a clean parser error.

## Ask

Pick one (or both) once there's a concrete need:

1. **Extend the parity test** to also check column data types/lengths (parse the type token between the column name and NULL/NOT NULL, compare against the EF property's configured `[StringLength]`/CLR type → SQL Server type mapping). This closes the biggest real gap (silent truncation) without needing Windows-only tooling.
2. **Implement `foundrygate db compare`** for real (DacFx `SchemaComparison` API, Windows-only — see plans/23-database-tooling.md's original design and imagile-app's `SchemaComparisonHelpers.cs`) as the authoritative alternative, with the parity test staying as the cross-platform CI backstop regardless.

Neither is blocking anything today — the seven current tables were hand-verified column-by-column against the entities during #77's review, and the regex parser already catches column existence/nullability/index/FK drift, which is the failure mode most likely to actually happen (someone edits an entity and forgets the `.sql` file, or vice versa).

kolatts:
## Summary

Closes #100 via **option 1** (extend the cross-platform parity test). `SchemaParityTests.CheckedInSqlFiles_ShouldMatchEfModel` now builds the `AppDbContext` model with the **SQL Server provider** (design-time `IModel` via `IDesignTimeModel` — no connection is ever opened; the connection string host is a deliberately unresolvable `.invalid`) instead of the SQLite in-memory harness, so real store types and provider annotations are available. Per table it now checks:

| Area | Before | Now |
|---|---|---|
| Columns | name set (both directions), NULL/NOT NULL | + store type incl. length/precision/scale (`nvarchar(200)` vs `(50)`, `nvarchar(max)`, `decimal(p,s)`, `bigint` vs `int`), case/whitespace-insensitive, `DATETIMEOFFSET (7)` folded to `datetimeoffset`; + `IDENTITY` presence vs `SqlServerValueGenerationStrategy` |
| Primary key | — | constraint name, column composition/order, CLUSTERED/NONCLUSTERED |
| Foreign keys | single-column only: name exists, `ON DELETE CASCADE` flag | composite-capable: dependent columns, principal table + columns, `ON DELETE` action derived from `DeleteBehavior` (`CASCADE` / `SET NULL` / `NO ACTION`); + FKs in the .sql absent from the model |
| Indexes | name exists, `UNIQUE` flag | + column composition, order, ASC/DESC; + indexes in the .sql absent from the model |
| Files | one .sql per entity | + orphaned .sql files with no mapped entity |

Single aggregate bullet-list assertion kept; XML doc rewritten into "what is checked" + "deliberate remaining limits" (regex parser tied to the repo's SQL style; no default/check constraints, computed columns, collations, filtered/INCLUDE indexes — none exist in the model today; compares to scripts, not a live DB). No exclusions were needed — the seven checked-in scripts already matched the model at type level, so **no .sql changes** in this PR.

`plans/23-database-tooling.md` status note + verification checklist updated accordingly.

## Verification

- `dotnet test src/FoundryGate.Tests.Predeployment -c Release --no-build` — **82/82** passing (same count as main: this extends the existing aggregate fact rather than adding a second one)
- Cross-platform: no SQLite/DacFx/Windows dependency added; `Microsoft.EntityFrameworkCore.SqlServer` already flows transitively from `FoundryGate.Data`

### Proof the new checks bite (then reverted)

Injected five deliberate drifts into the .sql files at once:

1. `Users.DisplayName` shrunk `NVARCHAR (200)` → `NVARCHAR (50)`
2. Stray `CREATE NONCLUSTERED INDEX [IX_Users_Email]` appended to `Users.sql`
3. `IX_QuotaAllocations_UserId_PeriodYear_PeriodMonth` columns reordered to `([PeriodYear] ASC, [UserId] ASC, [PeriodMonth] DESC)`
4. `IDENTITY (1, 1)` removed from `Groups.GroupId`
5. `ON DELETE CASCADE` removed from `FK_QuotaAllocations_Users_UserId`

The single assertion failed with exactly these five bullets:

```
Found 5 schema drift violation(s) between the EF model and FoundryGate.Database/dbo/Tables/*.sql:
  - Groups.GroupId: model is an IDENTITY column but Groups.sql does not declare IDENTITY
  - QuotaAllocations: FK FK_QuotaAllocations_Users_UserId DeleteBehavior is Cascade (ON DELETE CASCADE) but QuotaAllocations.sql declares ON DELETE NO ACTION
  - QuotaAllocations: index IX_QuotaAllocations_UserId_PeriodYear_PeriodMonth columns are ([UserId] ASC, [PeriodYear] ASC, [PeriodMonth] ASC) in the model but ([PeriodYear] ASC, [UserId] ASC, [PeriodMonth] DESC) in QuotaAllocations.sql
  - Users.DisplayName: model type is nvarchar(200) but Users.sql declares nvarchar(50)
  - Users.sql declares index IX_Users_Email that the EF model does not have
```

`git checkout -- src/FoundryGate.Database` restored the scripts; 82/82 green again.

## Deferred (option 2)

`foundrygate db compare` (DacFx `SchemaComparison`, Windows-only) was **not** implemented — it's a developer convenience for *regenerating* the scripts, not a CI gate, and the parity test now covers every drift class we've seen without needing it. Tracked as #103 for when hand-authoring tables becomes a recurring chore.

### DIFF
--- a/plans/23-database-tooling.md
+++ b/plans/23-database-tooling.md
@@ -24,10 +24,14 @@ Table file names below also use each table's actual (plural) name —
 `AppDbContext`'s `DbSet` names and the file-name-equals-table-name convention the parity test
 depends on.
 
-The parity test's documented gaps (no data type/length/precision checking, no composite-FK
-support, no index-composition validation beyond the `UNIQUE` flag) and the `db compare` deferral
-above are tracked as follow-up work in
-[#100](https://github.com/kolatts/foundry-gate/issues/100) rather than left as inline TODOs, per
+The parity test's originally documented gaps (no data type/length/precision checking, no
+composite-FK support, no index-composition validation beyond the `UNIQUE` flag) were closed by
+[#100](https://github.com/kolatts/foundry-gate/issues/100): the test now builds the EF model with
+the SQL Server provider (design-time model, no connection opened) and compares store types incl.
+length/precision, `IDENTITY`, PK name/columns/clustering, FK columns/principal/`ON DELETE`
+(composite-capable), index `UNIQUE`/columns/order/direction, plus extra indexes, FKs, and orphaned
+`.sql` files. The `db compare` deferral is tracked separately in
+[#103](https://github.com/kolatts/foundry-gate/issues/103) rather than left as an inline TODO, per
 CLAUDE.md's "everything is a GitHub issue" rule.
 
 ## Overview
@@ -241,7 +245,14 @@ A separate `db-deploy.yml` workflow (reusable, `workflow_call`) downloads the da
 - [ ] `Foundry Gate db compare` — **not implemented** (see the Status update note above): DacFx
       schema-compare is Windows-only and there is no live-database-to-compare-from in the
       EnsureCreated-based pipeline this repo actually uses. The `SchemaParityTests` Predeployment
-      test is the substitute drift check, and it runs cross-platform in CI.
+      test is the substitute drift check, and it runs cross-platform in CI. Tracked as a
+      developer-convenience follow-up in #103.
+- [x] `SchemaParityTests` type-level parity (#100) — verified it bites, not just passes: injected
+      five deliberate drifts at once (`Users.DisplayName` shrunk `NVARCHAR (200)` → `(50)`, a stray
+      `IX_Users_Email` index, `IX_QuotaAllocations_*` columns reordered + one `DESC`, `IDENTITY`
+      dropped from `Groups.GroupId`, `ON DELETE CASCADE` removed from
+      `FK_QuotaAllocations_Users_UserId`) and the single aggregate assertion reported exactly five
+      bullets naming each; reverted, 82/82 green.
 - [x] `foundrygate ip setup` is a documented stub (issues #96 tracks the real implementation) that
       prints guidance and exits 1 rather than doing anything; the reusable
       `.github/workflows/_deploy-database.yml` (#79) calls it with `continue-on-error: true` and a
--- a/src/FoundryGate.Tests.Predeployment/Data/Conventions/SchemaParityTests.cs
+++ b/src/FoundryGate.Tests.Predeployment/Data/Conventions/SchemaParityTests.cs
@@ -31,17 +31,26 @@ namespace FoundryGate.Tests.Predeployment.Data.Conventions;
 /// including length/precision/scale — <c>nvarchar(200)</c> vs <c>nvarchar(50)</c>,
 /// <c>nvarchar(max)</c>, <c>decimal(p,s)</c>, <c>bigint</c> vs <c>int</c>, etc. Comparison is
 /// case- and whitespace-insensitive, and SQL Server's implicit default precision is normalised
-/// (<c>DATETIMEOFFSET (7)</c> in the script equals EF's bare <c>datetimeoffset</c>).</item>
+/// (<c>DATETIMEOFFSET (7)</c> in the script equals EF's bare <c>datetimeoffset</c>). A column
+/// line that does not parse is reported as unparseable (quoting the line), never as
+/// "missing".</item>
+/// <item><c>DEFAULT</c> constraint <i>presence</i> per column matches the model
+/// (<c>HasDefaultValue</c>/<c>HasDefaultValueSql</c>); both the DacFx/SSDT placement
+/// (<c>CONSTRAINT [DF_x] DEFAULT ((1)) NOT NULL</c>) and EF's (<c>NOT NULL DEFAULT (...)</c>) are
+/// accepted. The default <i>expression</i> is not compared.</item>
 /// <item><c>IDENTITY</c> presence per column matches the model's SQL Server value-generation
 /// strategy (int <c>{Entity}Id</c> PKs are identity; composite/natural keys are not).</item>
-/// <item>Primary key constraint name, column composition and order, and clustering.</item>
+/// <item>Primary key constraint name, column composition and order, and clustering — declared
+/// either as <c>ALTER TABLE ... ADD CONSTRAINT [PK_*]</c> (this repo's style) or inline inside
+/// <c>CREATE TABLE</c> (what DacFx/SSDT scripting emits).</item>
 /// <item>Every foreign key (single- or multi-column) by constraint name: dependent columns,
 /// principal table and columns, and the <c>ON DELETE</c> action implied by its configured
 /// <see cref="DeleteBehavior"/>. Also flags FK constraints in the .sql file that the model does not
 /// declare.</item>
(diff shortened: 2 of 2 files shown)


kolatts:
**APPROVE**

This is the right way to do it — the design-time `IModel` trick is exactly what makes store types, `IDENTITY` and `IsClustered` reachable without a database, and I verified it's load-bearing, not cosmetic: the runtime model's `IsDescending` always returns an empty array (which this file's encoding would read as "all DESC"), so `Context.Model` would have been quietly wrong. Hand-checked all seven scripts against the model, built and ran it on the branch (82/82), and confirmed the "SQLite makes `GetValueGenerationStrategy` ambiguous" comment is true in EF 10. Ship it.

Non-blocking, but one is worth a quick look:

---

**Issues**

- **Minor** `src/FoundryGate.Tests.Predeployment/Data/Conventions/SchemaParityTests.cs` line 69 — `ColumnLineRegex` is now stricter than the `.+?` on main: a DacFx/SSDT-style `[IsActive] BIT CONSTRAINT [DF_Users_IsActive] DEFAULT ((1)) NOT NULL` (default *before* nullability, which is what schema-compare emits) fails to match and surfaces as `Users.IsActive: column missing from Users.sql` — I injected it and got exactly that. The limits doc does say defaults mean "extend the test", but the first `HasDefaultValue` will send someone hunting for a column that's right there; consider either tolerating an optional `CONSTRAINT [..] DEFAULT (..)` group or making the message say "column line didn't parse" instead of "missing". Same family: an inline `CONSTRAINT [PK_x] PRIMARY KEY CLUSTERED (...)` inside `CREATE TABLE` reports "no `ADD CONSTRAINT [PK_*]` statement found".

- **Nit** `SchemaParityTests.cs` line 82 — `IndexRegex` names a `clustered` capture group nothing reads (main had it non-capturing), while the doc lists per-index clustering as unchecked. Either compare it against `index.IsClustered()` (the model API is right there) or drop back to `(?:...)`.

---

Checked and fine, so nobody re-litigates: no connection opened and no disposal issue (model comes from the cached root provider, outlives the context); `IsDescending` null/empty/per-column encoding matches `MigrationsSqlGenerator`; `ToSqlOnDeleteAction` folding everything but Cascade/SetNull to `NO ACTION` matches what EF's SQL Server generator emits; `IsClustered() ?? true` is the right default; IDENTITY detection is correct for int PKs, the `GroupMembers` composite key and the `SystemConfigurations` string key; single aggregate assertion preserved; #103 exists for the deferred `db compare`.

kolatts:
Thanks for the review — and for catching that the runtime model's `IsDescending` would have read as all-DESC; good to have it on record that the design-time model is load-bearing. Fix pass is in 78f5065; both findings addressed, nothing deferred.

**Minor — `ColumnLineRegex` too strict for SSDT-style defaults / inline PK → fixed**

- `ColumnLineRegex` now accepts an optional `CONSTRAINT [DF_x] DEFAULT (expr)` clause **either** before the nullability keyword (what DacFx/SSDT emits: `BIT CONSTRAINT [DF_Users_IsActive] DEFAULT ((1)) NOT NULL`) **or** after it (what EF's own DDL emits: `NOT NULL DEFAULT (...)`). Nested parens like `((1))`/`(getutcdate())` are matched with .NET balancing groups, not `[^)]*`.
- Rather than just tolerating the clause, default *presence* is now compared against the model (`RelationalAnnotationNames.DefaultValue` annotation or `GetDefaultValueSql()`), so a `DEFAULT` showing up in a script with no matching `HasDefaultValue*` in the entity is flagged — `Users.IsActive: model has no default value but Users.sql declares a DEFAULT constraint`. Expressions are still not compared; the doc says so. (Gotcha worth noting in the code comment: `GetDefaultValue()` itself returns the CLR default for every non-nullable value type in EF 10, so the first cut flagged every `int`/`bool`/`Guid` column — hence the annotation check.)
- Silent skip is gone: any line that starts like a column (`[Name]`) but doesn't parse is reported as `column line for [TargetType] could not be parsed by the parity regex (expected '[Name] TYPE [IDENTITY (..)] [CONSTRAINT [DF_x] DEFAULT (..)] NULL|NOT NULL'): '<the line>'`, and the "column missing" bullet is suppressed for that name so nobody hunts for a column that's right there.
- `PrimaryKeyRegex` now matches an inline `CONSTRAINT [PK_*] PRIMARY KEY CLUSTERED (...)` table constraint inside `CREATE TABLE` as well as this repo's `ALTER TABLE ... ADD CONSTRAINT`. The not-found message names both accepted forms.

**Nit — unused `clustered` capture in `IndexRegex` → fixed (compared, not dropped)**

Compared against `index.IsClustered() ?? false` (SQL Server's default for a secondary index is nonclustered, mirroring the `?? true` on the PK). Doc updated: per-index clustering moved from the "unchecked" list to "checked".

**Proof (injected, verified, reverted):** SSDT-style default on `Users.IsActive`; `Groups` PK moved inline into `CREATE TABLE` with the `ALTER TABLE` removed; `AuditLogs` PK removed entirely plus a `COLLATE` clause on `TargetType`; `IX_QuotaAllocations_*` made `CLUSTERED`. Result: exactly four bullets — the unparseable-line message for `TargetType` (not "missing"), the no-PK message for `AuditLogs`, the clustering mismatch, and the default-presence mismatch — and **no** bullet for `Groups`, i.e. the inline PK parsed and matched. Reverted; 82/82 green, 0 warnings, `dotnet format` clean, CI green on 78f5065.

Issue #102, opened 2026-09-01
E2E assertions for 401/403 auth behavior on /api/v1 (needs deployed environment)


kolatts:
> Parent: #220 — Live validation of the dev environment

Deferred from PR #99 (issues #26/#27): integration assertions that /api/v1 endpoints return 401 without a bearer token and 403 without the FoundryGate.Admin role require a real Entra tenant + deployed environment (WebApplicationFactory can't mint real Entra tokens). Belongs with FoundryGate.Tests.Postdeployment once an environment exists (#67/#43). Filed per CLAUDE.md's everything-is-an-issue rule during session handoff.

kolatts:
## Live validation — PASS, and it found a blocker

Run against dev (`ca-foundrygate-api-dev.wittygrass-b608ecfa.eastus2.azurecontainerapps.io`), tenant `67fdff67-…`, real Entra tokens.

### The token

The Azure CLI's public client is now pre-authorized for `access_as_user` on `FoundryGate.Api (dev)`, alongside the SPA (`preAuthorizedApplications` is replaced wholesale, so both had to be listed):

```bash
az rest --method PATCH \
  --uri "https://graph.microsoft.com/v1.0/applications(appId='7e7d0561-…')" \
  --body '{"api":{"preAuthorizedApplications":[
     {"appId":"21b82312-…","delegatedPermissionIds":["92cffb8c-…"]},
     {"appId":"04b07795-8ddb-461a-bbee-02f9e1bf7b46","delegatedPermissionIds":["92cffb8c-…"]}]}}'
```

**`--resource` does not work; `--scope` does.** `az account get-access-token --resource api://7e7d0561-…` fails `Status_InteractionRequired` and tells you to `az login --scope …`; `--scope api://7e7d0561-…/access_as_user` succeeds against the existing session. Claims (values elided):

```
aud  7e7d0561-0973-411d-ba62-a667cbfec1d9      ver  2.0
iss  https://login.microsoftonline.com/67fdff67-…/v2.0
azp  04b07795-8ddb-461a-bbee-02f9e1bf7b46      scp  access_as_user
roles ["FoundryGate.Admin"]
```

[...]

Issue #66, opened 2026-06-04
Implement re-activation with full APIM re-provision and orphan subscription detection


kolatts:
Part of #64.

`POST /users/{id}/activate` must call `IUserLifecycleService.ReactivateAsync`: set `IsActive = true`, run quota resolution for the current period, then provision a new APIM subscription. Before calling APIM to create, query the Management plane for an existing subscription named `foundrygate-{userId}` — if found (orphan from a failed previous deprovision), reuse it and update DB fields rather than creating a duplicate. `DELETE /keys/{userId}` (key-only revocation without full deactivation) must only call APIM delete + audit log, not the full deprovision sequence — user stays active and can request a new key.

kolatts:
Claimed by an implementation agent on branch claude/users-lifecycle-28-29-65-66 (users lifecycle wave: IUserLifecycleService, GET /users/me, admin user endpoints, ApimGatewayTierSync).

Issue #65, opened 2026-06-04
Wire IUserLifecycleService as single pipeline orchestrator; fix Entra departure deprovision


kolatts:
Part of #64.

Create `IUserLifecycleService` with `ProvisionAsync`, `DeprovisionAsync`, and `ReactivateAsync`. Wire it into `UsersController` (activate/deactivate), `EntraUserSyncService` (departure detection — currently only sets `IsActive=false`, missing APIM subscription deletion), and `KeysController` (provision/revoke). The service holds the canonical step sequence from plan #21. Use a DB transaction for the DB steps; APIM calls happen outside the transaction with documented compensation on failure. Add `SuspendAsync`, `ReenableAsync`, and `DeleteAsync` to `IApimKeyService` to make the distinction between quota suspension and account deactivation explicit.

kolatts:
Note from the keys wave (PR #133): `IApimKeyService` now ships the deprovision building block in two flavours — `RevokeAsync(User)` (attributes `key.revoked` to the current HTTP caller; used by `DELETE /keys/{userId}` and suitable for `POST /users/{id}/deactivate`) and `RevokeAsSystemAsync(User, reason)` (no `ICurrentUserAccessor` involved; writes the audit row via `IAuditWriter.AddSystem` with the reason in the details). Trigger B — Entra departure detected by the sync — should call the system variant; both are no-ops returning `false` for a user with no key, so they can be called unconditionally. Also: `ProvisionAsync(User, tier)` claims the row inside `dbContext.Database.CurrentTransaction` when the orchestrator has one open (it only opens its own when none exists), so wrapping the pipeline in a single transaction works as plan 21 describes. There is no `SuspendAsync`/`ReenableAsync`/`DeleteAsync` to add — per the #116 ruling the lifecycle has no suspended state, and `RevokeAsync` is the delete.

kolatts:
Claimed by an implementation agent on branch claude/users-lifecycle-28-29-65-66 (users lifecycle wave: IUserLifecycleService, GET /users/me, admin user endpoints, ApimGatewayTierSync).

Issue #61, opened 2026-06-04
Implement /foundry/deployments (admin) and /foundry/models (any user) API endpoints


kolatts:
Part of #60.

Create `FoundryDeploymentsController` with three admin-only endpoints: `GET /foundry/deployments` lists all deployments on the configured Foundry resource; `POST /foundry/deployments` provisions a new deployment (model name, version, SKU, capacity); `DELETE /foundry/deployments/{name}` removes it. Back the controller with `IFoundryDeploymentService` using `ArmClient` + `DefaultAzureCredential`. The Foundry resource ID is read from `SystemConfiguration["FoundryResourceId"]`. Write `foundry.deployment.created` and `foundry.deployment.deleted` audit log entries. Add `Azure.ResourceManager.CognitiveServices` package to `FoundryGate.Api.csproj`.

kolatts:
Claiming #61 — implementing `/api/v1/foundry/deployments` (admin) + `/api/v1/foundry/models` (any user) on branch `claude/foundry-deployments-61`. ARM calls go behind a thin `IFoundryManagementClient` with a fake in tests; create checks existence first (409, never re-PUT — E-007); Anthropic path decided against the SDK's ability to express `modelProviderData` (see #107).

PR #112, opened 2026-09-01
feat(api): audit service, GET /audit, and cross-cutting API scaffolding

### LINKED ISSUE #42: Wire IAuditService writes to all mutating endpoints and key lifecycle actions
Part of #12.

Create an `IAuditService` with a single `LogAsync(actorId, action, targetType, targetId, details)` method backed by an EF INSERT into `AuditLog`. Inject it into every controller and service that mutates state: quota changes (`quota.set`, `quota.approved`, `quota.rejected`), key events (`key.provisioned`, `key.rotated`, `key.revoked`), user lifecycle (`user.activated`, `user.deactivated`), group changes (`group.member-added`, `group.member-removed`), resets (`quota.monthly-reset`), and sync runs. The `details` field is a JSON blob with before/after values for audit reviewers. Expose `GET /audit` (admin) as a paged, filterable endpoint.

kolatts:
## Summary

First `/api/v1` controller, plus the cross-cutting scaffolding the seven endpoint waves that follow (users, groups, quota, requests, keys, Entra sync, foundry deployments) build on.

**Issue #42 — audit**
- `Services/Audit/IAuditService` + `AuditService`: `LogAsync(action, targetType, targetId, details, ct)` attributed to the current caller, and `LogAsync(int? actorUserId, ...)` for system actors (Functions/sync jobs pass `null`). `details` is any object, JSON-serialized (camelCase) into `AuditLog.Details`; `OccurredDate` from the injected `TimeProvider`. Returns the added `AuditLog`.
- `Controllers/AuditController`: `GET /api/v1/audit`, `[Authorize(Policy = PolicyNames.AdminOnly)]`, paged + filterable via the existing `AuditLogQuery` / `PagedRequest` records -> `PagedResult<AuditLogEntryResponse>`. Projection-to-record inside the query, `AsNoTracking`, newest first (`OccurredDate` desc, `AuditLogId` desc).
- `AuditActions`: added `UserProvisioned`, `QuotaIncreaseSubmitted`, `QuotaAllocationReset`, `FoundryDeploymentCreated/Deleted`, `KeyRevealed`, `UsageSynced` (existing ones untouched). New `AuditTargetTypes` constants so call sites never spell `"User"`/`"Group"`/`"Request"` inline.

**Cross-cutting scaffolding**
- `Services/Identity/ICurrentUserAccessor` + `CurrentUserAccessor` (scoped): `EntraObjectId` (accepts both `oid` and the long `.../objectidentifier` claim type via Microsoft.Identity.Web's `GetObjectId()`), `IsAdmin` (`IsInRole` — the same check as the policy), `DisplayName`/`Email` (for #28's auto-provision), `TryGetUserAsync` (null = not provisioned, a first-class path; misses are not cached so a provisioning caller sees the new row), `GetRequiredUserAsync` (-> `KeyNotFoundException`/404). Returned `User` is tracked. No oid -> `UnauthorizedAccessException`/403.
- `Controllers/ApiControllerBase` — `[ApiController]`, `[Route("api/v1/[controller]")]`, `[Produces("application/json")]`.
- DI convention: `Services/ServiceCollectionExtensions.AddFoundryGateApiServices()` called **once** from `Program.cs`; each area owns `Services/<Area>/<Area>ServiceCollectionExtensions.Add<Area>Services()`. Later waves add one file + one line and never touch `Program.cs`.
- `FoundryGate.Data.Extensions.QueryableExtensions.ToPagedAsync(PagedRequest, ct)` -> `PagedResult<T>` (CONVENTIONS.md asked for this; it didn't exist yet). Clamps paging itself.
- `GatewayTiers` (coordinator-requested): APIM tier product ids `standard`/`power`/`unlimited`, `Default = Standard`, `All`; XML-doc'd as APIM products because `token-quota` accepts literals only, and as must-match-`infra/main.bicep`. `SystemConfigurationKeys.ApimProductId` doc amended to "legacy single-product model, superseded by `GatewayTiers`" (constant kept — seed data references it). `GatewayTiersTests` parses the bicep `quotaTiers` default and asserts parity + order.
- Test harness: `Api/ApiTestFactory` (`WebApplicationFactory<Program>`, env `local`, SQL Server `AppDbContext` swapped for SQLite in-memory on one kept-open `SqliteConnection`, `EnsureCreated` + `ReferenceDataSeeder`, `TestAuthHandler` as default authenticate/challenge scheme, `AzureAd` placeholders supplied in-memory so it never needs Azure) with `CreateClientAs(oid, isAdmin, name, email)`, `SeedUserAsync(...)`, `CreateDbContext()`, `TimeProvider` (a `Support/MutableTimeProvider`). `HealthEndpointTests` unchanged and still green.
- CONVENTIONS.md: new **"API service/controller conventions"** section documenting all of the above for the next agents.

## Design decisions (please scrutinize)

1. **Audit rows are NOT fire-and-forget.** `LogAsync` adds to the request's `AppDbContext`; the caller's `SaveChangesAsync` commits mutation + audit atomically. The plan's "fire-and-log" verification item is reversed with rationale: a fire-and-log audit can leave a mutation with no audit row, or an audit row for a rolled-back mutation — for an audit trail, failing the request is the correct outcome. Pattern at every call site: mutate -> `await audit.LogAsync(...)` -> `await dbContext.SaveChangesAsync(ct)`.
2. **A caller with no `User` row cannot write a human-attributed audit row** — the current-user `LogAsync` overload throws `UnauthorizedAccessException` (403) rather than silently logging `ActorUserId = null`. Admins are provisioned by `GET /users/me` (#28) before they act, so this only bites an admin who calls a mutating endpoint before ever loading the UI. Alternative considered: fall back to a null actor with the oid in `details` — rejected because it makes "system action" and "unprovisioned human" indistinguishable in the log.
3. **SQLite-only `DateTimeOffset` -> UTC-ticks value conversion in `AppDbContext.OnModelCreating`** (gated on `Database.ProviderName`). EF's SQLite provider refuses to translate ordering/comparison on `DateTimeOffset`, which would have broken the audit date-range filter (and every future quota-period/review-date query) in the test harness. Normalizing to `UtcTicks` (not EF's built-in `DateTimeOffsetToBinaryConverter`) means a query parameter with a non-UTC offset still compares by instant — pinned by `Date_filter_with_a_non_utc_offset_compares_by_instant_not_by_wall_clock`. SQL Server model and dacpac untouched; column names/nullability unchanged so convention/parity tests are unaffected.
4. **Empty-string <-> null translation happens in the projection.** Entities keep the non-nullable-string convention (`TargetType = ""` for "no target"); the pre-existing `AuditLogEntryResponse` contract exposes `null`. The query maps `""` -> `null` so the UI never sees `""`.
5. `IsAdmin` uses `ClaimsPrincipal.IsInRole` rather than inspecting claims directly, so it can never disagree with `PolicyNames.AdminOnly`. `TestAuthHandler` builds identities with role claim type `roles`, matching Microsoft.Identity.Web's real tokens.
6. `ICurrentUserAccessor` also exposes `DisplayName`/`Email` — slightly beyond the brief, but #28 needs both for auto-provisioning and this avoids a second claim-parsing site.

## Verification

```
dotnet build FoundryGate.sln -c Release                                 -> 0 Warning(s), 0 Error(s)
dotnet test src/FoundryGate.Tests.Predeployment -c Release --no-build   -> Passed! Failed: 0, Passed: 119, Skipped: 0, Total: 119  (82 on main + 37 new)
dotnet format FoundryGate.sln --verify-no-changes                       -> exit 0
```

New tests: `AuditServiceTests` (9 — actor attribution, camelCase JSON, `TimeProvider` timestamp incl. SQLite round-trip, adds-without-saving, null details -> `""`, 403 on unprovisioned caller, system-actor overload, blank action rejected, newest-first + `""`->null projection), `CurrentUserAccessorTests` (10 — both oid claim types, 403 on missing oid / unauthenticated / no HttpContext, `IsAdmin`, name/email claims, null miss, 404, tracked + cached, miss not cached), `AuditEndpointTests` (11 — 401 anonymous, 403 non-admin, 200 admin + default paging, newest first + actor display name, page/pageSize/totals, each filter, inclusive date range, non-UTC offset), `QueryableExtensionsTests` (3), `GatewayTiersTests` (2).

plans/12-audit-logging.md Verification checklist updated; "every mutation produces a row" is explicitly deferred to each endpoint wave's own PR (the endpoints don't exist yet — #28–#41, #61, #65–#66 each wire their own `LogAsync` and assert it).

Also closes the two plans/04 deferrals in-process: unauthenticated -> 401 and wrong role -> 403 are now asserted against a real `/api/v1` endpoint (#102 remains for the deployed-environment version).

## Not in this PR
- Wiring `LogAsync` into mutating endpoints — they don't exist yet; each wave owns its call sites (tracked by the existing per-endpoint issues).
- No `.github/`, `infra/`, `docs-site/`, or `SchemaParityTests.cs` changes (other agents own those). Landing page / architecture docs unaffected: no enforcement-semantics or topology change.

Closes #42

### DIFF
--- a/CONVENTIONS.md
+++ b/CONVENTIONS.md
@@ -74,10 +74,14 @@ Landed with #42 (the first `/api/v1` controller); every endpoint wave builds on 
 
 - **Controllers** inherit `Controllers/ApiControllerBase` (`[ApiController]`,
   `[Route("api/v1/[controller]")]`, `[Produces("application/json")]`) and declare only
-  their `[Authorize(Policy = PolicyNames.AdminOnly)]` (class- or action-level) plus thin
-  actions that delegate to a service. Authentication is already global (the
-  `AuthorizeFilter` in Program.cs): anonymous → 401, non-admin on an admin-only route →
-  403, with no per-controller code.
+  their `[Authorize(Policy = PolicyNames.AdminOnly)]` (class- or action-level) plus
+  **expression-bodied** actions that delegate to a service. No `ThrowIfNull` on bound
+  parameters: `[ApiController]` never hands an action a null `[FromQuery]`/`[FromBody]`
+  record (an empty query string yields an instance with defaults; a missing body is a
+  400 before the action runs), so the guard is unreachable — the first-line
+  `ThrowIfNull` rule in §Style applies to services, not controller actions.
+  Authentication is already global (the `AuthorizeFilter` in Program.cs): anonymous →
+  401, non-admin on an admin-only route → 403, with no per-controller code.
 - **Services** live in `Services/<Area>/` with the single-implementation interface
   co-located (`IAuditService.cs` + `AuditService.cs`); primary constructors; take
   `AppDbContext` directly; throw `KeyNotFoundException` (404) / `ArgumentException` (400)
@@ -85,39 +89,59 @@ Landed with #42 (the first `/api/v1` controller); every endpoint wave builds on 
   status codes themselves.
 - **DI — one line per area, one file.** Each area owns
   `Services/<Area>/<Area>ServiceCollectionExtensions.cs` exposing `Add<Area>Services()`;
-  register it by adding ONE line to `Services/ServiceCollectionExtensions
+  register it by adding ONE line to `Services/ApiServiceCollectionExtensions
   .AddFoundryGateApiServices()`, which Program.cs calls once. Never register application
-  services in Program.cs — parallel waves would collide there.
+  services in Program.cs. (Parallel waves will still conflict on adjacent lines in that
+  one method — a trivially-resolved conflict, versus registrations-plus-usings churn in
+  Program.cs.) Data-layer services every host shares (`IAuditWriter`, the interceptor,
+  `TimeProvider`) are registered in Data's `AddFoundryGateData`, not here.
 - **Current user**: inject `ICurrentUserAccessor` (scoped) rather than reading
   `HttpContext.User`. `EntraObjectId` (accepts both `oid` and the long
   `.../objectidentifier` claim type), `IsAdmin` (same `IsInRole` check as the policy),
   `DisplayName`/`Email` (token claims, for auto-provisioning), `TryGetUserAsync(ct)`
-  (null = not provisioned yet — a first-class path; `GET /users/me` provisions on it),
-  `GetRequiredUserAsync(ct)` (→ 404). The returned `User` is tracked by the request's
-  context, so mutate it and `SaveChangesAsync`. No oid on an authenticated principal →
-  `UnauthorizedAccessException` (403).
-- **Audit**: inject `IAuditService` and call
-  `await audit.LogAsync(AuditActions.X, AuditTargetTypes.Y, id.ToString(), new { before, after }, ct)`
-  *before* the caller's own `SaveChangesAsync` — the row is added to the SAME
-  `AppDbContext` and commits atomically with the mutation. No fire-and-forget audit: a
-  mutation without its audit row (or vice versa) is worse than a failed request. System
-  actors (Functions, sync jobs) use the `LogAsync(int? actorUserId, ...)` overload with
-  `null`. New action/target kinds are constants in Domain's `AuditActions` /
-  `AuditTargetTypes`, never inline strings. `details` is JSON-serialized (camelCase) —
-  never put a key or token in it.
+  (null = not provisioned yet — a first-class path; `GET /users/me` provisions on it; it
+  checks the change tracker first, so a `User` you just `Add`ed is found before you
+  save), `GetRequiredUserAsync(ct)`. The returned `User` is tracked by the request's
+  context, so mutate it and `SaveChangesAsync`. **"No `User` row for this caller" is 403
+  everywhere, never 404** (`GetRequiredUserAsync` and `IAuditService.LogAsync` both throw
+  `UnauthorizedAccessException` saying "call `GET /users/me` first"): an authenticated
+  principal without a row is an authorization-state problem, not a missing resource. No
+  oid on an authenticated principal → `UnauthorizedAccessException` (403) too.
+- **Audit — one writer, two entry points.** The row builder is
+  `FoundryGate.Data.Audit.IAuditWriter` (Data, so Functions/Cli use the same one):
+  `Add(User actor, …)` (attributes via the `ActorUser` navigation — works for an actor
+  that is itself unsaved), `Add(int actorUserId, …)`, `AddSystem(…)` (null actor; the
+  monthly reset / usage sync / Entra sync jobs). Api code acting as the current caller
+  uses the thin wrapper `IAuditService.LogAsync(AuditActions.X, AuditTargetTypes.Y,
+  id.ToString(), new { before, after }, ct)`; Api code that already holds a `UserId` or
+  acts as the system uses `IAuditWriter` directly. Either way call it *before* your own
+  `SaveChangesAsync` — the row is added to the SAME `AppDbContext` and commits atomically
+  with the mutation; nothing in the audit path saves. No fire-and-forget audit: a
+  mutation without its audit row (or vice versa) is worse than a failed request. For
+  auto-provisioning that means `Add(user)` → `LogAsync(UserProvisioned…)` → one save.
+  New action/target kinds are constants in Domain's `AuditActions` / `AuditTargetTypes`,
+  never inline strings. `details` is JSON-serialized (camelCase, cycles ignored) — pass
+  the fields you mean rather than a tracked entity, and never a key or token.
 - **Paging**: bind `[FromQuery] PagedRequest paging` (alongside any `[FromQuery]` filter
   record) and finish the query with `.OrderBy(...).Select(projection).ToPagedAsync(paging, ct)`
   (`FoundryGate.Data.Extensions`) → `PagedResult<T>`. Order deterministically first;
-  the helper clamps page/size itself.
+  the helper clamps page/size itself. Index the columns your default ordering and
+  filters use (`AuditLog.OccurredDate` is the precedent) and mirror the index in
+  `FoundryGate.Database/dbo/Tables/*.sql` — `SchemaParityTests` checks name, uniqueness,
+  clustering and column composition.
 - **Integration tests** use `IClassFixture<ApiTestFactory>` (real pipeline, SQLite
   in-memory, header-driven `TestAuthHandler`): `factory.CreateClient()` is anonymous,
   `factory.CreateClientAs(oid, isAdmin: true)` is an admin, `factory.SeedUserAsync(...)`
   and `factory.CreateDbContext()` arrange/assert rows, `factory.TimeProvider` moves the
   clock. One database per test class — seed with unique markers, never assert absolute
   counts. Service-level tests use `InMemoryDatabaseTest` + `Support/MutableTimeProvider`
-  + hand-rolled stubs (no mocking library). Under SQLite, `AppDbContext` stores
-  `DateTimeOffset` as UTC ticks (provider-gated) so ordering and date-range filters
-  translate; the SQL Server model is untouched.
+  + `Support/FixedHttpContextAccessor` + hand-rolled stubs (no mocking library; the
+  framework's `HttpContextAccessor` is a static `AsyncLocal` — never use it in tests).
+  Under SQLite, `AppDbContext` stores `DateTimeOffset` as UTC ticks (provider-gated) so
+  ordering and date-range filters translate; the SQL Server model is untouched. Known
+  test-only divergence: values read back from SQLite always carry `+00:00` (the instant
+  is preserved, the original offset is not), where SQL Server's `datetimeoffset` keeps
+  it — assert on instants, never on `.Offset`.
 
 ## Schema pipeline (no EF migrations)
 
--- a/CONVENTIONS.md
+++ b/CONVENTIONS.md
@@ -68,6 +68,57 @@
   (404/400/403/409 mapping), not per-controller try/catch. Controllers are thin
   expression-bodied delegations with class-level `[Authorize]`.
 
+## API service/controller conventions
+
+Landed with #42 (the first `/api/v1` controller); every endpoint wave builds on these.
+
+- **Controllers** inherit `Controllers/ApiControllerBase` (`[ApiController]`,
+  `[Route("api/v1/[controller]")]`, `[Produces("application/json")]`) and declare only
+  their `[Authorize(Policy = PolicyNames.AdminOnly)]` (class- or action-level) plus thin
+  actions that delegate to a service. Authentication is already global (the
+  `AuthorizeFilter` in Program.cs): anonymous → 401, non-admin on an admin-only route →
+  403, with no per-controller code.
+- **Services** live in `Services/<Area>/` with the single-implementation interface
+  co-located (`IAuditService.cs` + `AuditService.cs`); primary constructors; take
+  `AppDbContext` directly; throw `KeyNotFoundException` (404) / `ArgumentException` (400)
+  / `ConflictException` (409) / `UnauthorizedAccessException` (403) and never touch HTTP
+  status codes themselves.
+- **DI — one line per area, one file.** Each area owns
+  `Services/<Area>/<Area>ServiceCollectionExtensions.cs` exposing `Add<Area>Services()`;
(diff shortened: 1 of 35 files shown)


kolatts:
**REQUEST CHANGES**

This is the scaffolding I'd want seven waves to build on — atomic audit rows, tracked `User` per request, one-file-one-line DI, and a test harness that exercises the real pipeline. Two things need deciding before it sets in concrete: the current-user `LogAsync` cannot attribute a user that was `Add`ed but not yet saved, which means the exact `mutate → LogAsync → SaveChanges` pattern the epic prescribes for `GET /users/me` (#28, wave one) throws 403; and the system-actor overload built for Functions lives in a project Functions can't reference.

Verified on the branch: build clean for Api/Data/Domain/Web/Tests, 119/119 green, `dotnet format` clean. (`FoundryGate.Functions` hit MSB3030 in the WorkerExtensions inner build in my sandboxed worktree — the PR doesn't touch Functions and main has the same artifacts built, so that's my environment, not you.) Everything below was checked with a probe test, not guessed.

---

**Issues**

- **Major** `src/FoundryGate.Api/Services/Identity/CurrentUserAccessor.cs` line 494 — `SingleOrDefaultAsync` goes to the database, so a `User` the calling service has just `Add`ed is invisible until `SaveChanges`; the "miss is not cached so a provisioning caller sees the new row" promise only holds after a save, and `AuditService.LogAsync` therefore 403s on the auto-provision path. Consider checking `dbContext.Users.Local` first and having `Add` set the `ActorUser` navigation rather than `ActorUserId` — then #28 gets user + `user.provisioned` row in one atomic save, as the interface remarks promise.

- **Major** `src/FoundryGate.Api/Services/Audit/IAuditService.cs` line 410 — the `LogAsync(int? actorUserId, …)` overload is documented for "the monthly reset, usage sync, Entra sync jobs" (#38/#39), but `FoundryGate.Functions` references only Data and Domain, so it can never call this. The actor-agnostic writer wants to live in Data (with the Api service wrapping it for the current-user path) — decide now, before two hosts grow two audit writers.

- **Minor** `src/FoundryGate.Api/Services/Audit/AuditService.cs` line 250 — same state ("no `User` row for this oid") is a 404 from `GetRequiredUserAsync` and a 403 here; an admin who holds the role but hasn't hit `/users/me` yet will read "Forbidden" as a role problem and open the wrong ticket. Pick one status — or, since #28 exists, provision on demand and make the question moot.

- **Minor** `src/FoundryGate.Database/dbo/Tables/AuditLogs.sql` line 20 — the only index is `ActorUserId`, and the shipped query orders every page by `OccurredDate DESC, AuditLogId DESC`, so an unfiltered admin view is a full sort of the table forever. Needs `[Index]` on the entity plus the .sql; a follow-up issue is fine per the repo rules.

- **Minor** `src/FoundryGate.Api/Services/Audit/AuditService.cs` line 333 — `details` is "any object" and the default `System.Text.Json` cycle handling throws, so someone passing a tracked `Group` with its membership collection gets an unmapped 500 with the mutation lost (probe-verified). `ReferenceHandler.IgnoreCycles`, or say "never pass an entity" in the docs. Also, `JsonSerializerDefaults.Web` does not enable relaxed escaping — the comment on line 240 says it does.

- **Minor** `src/FoundryGate.Data/AppDbContext.cs` line 677 — `new DateTimeOffset(ticks, TimeSpan.Zero)` means the SQLite harness always reads back `+00:00` where SQL Server would preserve `-05:00` (probe-verified; instant-equality assertions hide it). Fine as a test-only divergence, but worth one sentence in the summary so nobody chases a phantom bug in a future `ReviewedDate` test.

- **Minor** `src/FoundryGate.Tests.Predeployment/Domain/GatewayTiersTests.cs` line 29 — this pins the bicep parameter *default*, not what a `.bicepparam` deploys, and `^\s*name:` will happily match a nested `name:` inside a tier entry; both fine today, but say so in the comment so the first person who adds a nested block knows why the test broke.

- **Nit** `src/FoundryGate.Api/Services/ServiceCollectionExtensions.cs` line 21 — `FoundryGate.Api.Services.ServiceCollectionExtensions` and `FoundryGate.Data.ServiceCollectionExtensions` are both imported in `Program.cs`; it compiles, but `ApiServiceCollectionExtensions` costs nothing and saves someone a confused minute.

- **Nit** `src/FoundryGate.Api/Controllers/AuditController.cs` line 185 — `[ApiController]` never binds a `[FromQuery]` record to null, so the two `ThrowIfNull`s are ceremony that keeps the action from being the expression-bodied delegation CONVENTIONS.md still asks for (the new section quietly relaxes that rule — either amend the old line or drop the guards).

- **Nit** `src/FoundryGate.Api/Services/Audit/AuditService.cs` line 264 — `ThrowIfCancellationRequested` in a non-`async` method throws synchronously instead of returning a faulted task; make it `async` or drop the check.

- **Nit** `CONVENTIONS.md` line 46 — the DI file is still a seven-way adjacent-line merge point; the conflicts are trivial, just don't sell it as hotspot-free.

---

Design decisions 1 (atomic audit), 3 (UTC-ticks converter, gated correctly — `HealthEndpointTests` proves the SQL Server model never sees it), 4, 5 and 6 all hold up. Decision 2 is where the bootstrap question lives: there's no deadlock *if* #28 can provision atomically, which is exactly what the first Major blocks today.

kolatts:
Author response to the review — fix pass in `5655596` (after merging `origin/main` for the index-composition parity checks). Every finding below is answered with the change made or the reason it was deferred; nothing was skipped.

### Majors

**Major 1 — `TryGetUserAsync` blind to an `Add`ed-but-unsaved `User` → 403 on #28's auto-provision.** Fixed as suggested, both halves:
- `CurrentUserAccessor.TryGetUserAsync` now consults `dbContext.Users.Local` (change tracker) before `SingleOrDefaultAsync`, so a `User` the calling service has just `Add`ed is found before the save.
- The writer attributes via the **`ActorUser` navigation**, not `ActorUserId`, so EF fixes up the FK at `SaveChanges` even when the actor's `UserId` is still `0`.
- Tests reproduce the exact #28 sequence (Add user → `LogAsync` → `SaveChanges` → row carries the new `UserId`) at two levels: `AuditWriterTests.Add_with_an_unsaved_actor_attributes_the_row_once_the_caller_saves…` and `AuditServiceTests.LogAsync_attributes_a_caller_whose_User_was_added_but_not_yet_saved…` (the latter through the **real** `CurrentUserAccessor` + `AuditWriter`, not a stub). Plus `CurrentUserAccessorTests.TryGetUserAsync_finds_a_User_that_was_added_to_the_context_but_not_yet_saved`.

**Major 2 — system-actor writer unusable from `FoundryGate.Functions`.** Moved. The actor-agnostic writer is now `FoundryGate.Data.Audit.IAuditWriter` / `AuditWriter`, registered scoped in `AddFoundryGateData` (so Api, Functions, Cli all get the same one):
- `Add(User actor, …)` — navigation-attributed (the auto-provision path);
- `Add(int actorUserId, …)` — already-persisted actor by id;
- `AddSystem(…)` — null actor for the reset / usage-sync / Entra-sync jobs.
- Methods are **synchronous** (they only touch the change tracker — nothing to await), which also removes the `ThrowIfCancellationRequested` nit outright. Three named overloads rather than `int?`/`User?` so `Add(null, …)` can't be ambiguous.
- Api's `IAuditService` is now the thin wrapper: `LogAsync(action, targetType, targetId, details, ct)` resolves the caller via `ICurrentUserAccessor` and calls `writer.Add(actor, …)`; plus `QueryAsync`. Api code that already holds a `UserId` or acts as the system uses `IAuditWriter` directly. `AddFoundryGateApiServices()` shape unchanged for the next waves.

### Minors

**404 vs 403 for "no `User` row".** Made consistent as 403 (orchestrator's preference): `GetRequiredUserAsync` now throws `UnauthorizedAccessException` with the same message `LogAsync` uses — "No FoundryGate user exists for the caller (oid …). Call GET /users/me first — it provisions the user on first login." Rationale documented in the `ICurrentUserAccessor` type remarks and the CONVENTIONS "Current user" bullet: an authenticated principal without a row is an authorization-*state* problem, not a missing resource. Tests updated (`GetRequiredUserAsync_throws_UnauthorizedAccessException_…telling_the_caller_to_provision_first`, and `LogAsync_throws…` asserts the `GET /users/me` text).

**Missing `OccurredDate` index.** Added now: `[Index(nameof(OccurredDate))]` on `AuditLog` (ascending — SQL Server scans it backwards for the DESC order) and `IX_AuditLogs_OccurredDate` in `dbo/Tables/AuditLogs.sql`. Merged `origin/main` first so `SchemaParityTests`' new index-composition check covers the pair; 126/126 green including it.

**JSON cycles → unmapped 500.** `DetailsSerializerOptions` now sets `ReferenceHandler.IgnoreCycles`; the wrong "relaxed escaping" comment is gone (it now says camelCase + cycle-tolerant, and why). `AuditWriterTests.Add_tolerates_object_cycles_in_details_instead_of_throwing` passes a `Group ↔ GroupMember ↔ Group` graph. The `IAuditWriter` remarks also say "prefer projecting the fields you mean over passing a tracked entity".

**SQLite UTC-normalisation read-back divergence.** Documented in two places: a `<remarks>` on `AppDbContext.ApplySqliteDateTimeOffsetConversion` (instant preserved, offset always reads back `+00:00`, never assert on `.Offset`) and the CONVENTIONS integration-test bullet.

**`GatewayTiersTests` regex.** Tightened: the block match anchors `^param quotaTiers array = [` and the `^]` at column 0; items are parsed as top-level `^  {` … `^  }` blocks (two-space indent) and the tier's `name:` must sit at exactly four spaces — a nested object's `name:` (six+) can't match. Added `Assert.NotEmpty` so a regex miss fails loudly rather than comparing empty lists. The method's XML doc now states the scope explicitly: it pins the bicep parameter *default*, not what a `.bicepparam` deploys.

### Nits

- **`ServiceCollectionExtensions` name collision** — renamed to `ApiServiceCollectionExtensions` (method still `AddFoundryGateApiServices()`); crefs updated.
- **`ThrowIfNull` on `[FromQuery]` records** — dropped; `AuditController.ListAsync` is expression-bodied again. Rule picked and written into CONVENTIONS: `[ApiController]` never binds a `[FromQuery]`/`[FromBody]` record to null, so controller actions carry no guards and stay expression-bodied; the §Style first-line `ThrowIfNull` rule applies to services. A `<remarks>` on the action says the same.
- **Sync `ThrowIfCancellationRequested`** — gone (the method that had it no longer exists; the writer is synchronous).
- **"Never a merge hotspot"** — reworded in both the `ApiServiceCollectionExtensions` summary and CONVENTIONS: parallel waves will still conflict on adjacent lines in that one method; it's a one-line trivially-resolved conflict versus registrations-plus-usings churn in `Program.cs`.

### Also
- New shared test double `Support/FixedHttpContextAccessor` (the framework's `HttpContextAccessor` keeps its context in a static `AsyncLocal`, which silently made side-by-side accessor assertions test the same principal); CONVENTIONS says never to use the framework one in tests.
- plans/12 verification list and deviations updated for the writer-in-Data split, 403 consistency, and the index.

### Verification (after fix pass, on top of merged `origin/main`)
```
dotnet build FoundryGate.sln -c Release                               → 0 Warning(s), 0 Error(s)
dotnet test src/FoundryGate.Tests.Predeployment -c Release --no-build → Passed! 126/126
dotnet format FoundryGate.sln --verify-no-changes                     → exit 0
```

### Final API shapes for the next waves (changed since the first push)
- `FoundryGate.Data.Audit.IAuditWriter` (**new**, Data): `AuditLog Add(User actor, string action, string targetType, string targetId, object? details)` · `AuditLog Add(int actorUserId, …)` · `AuditLog AddSystem(string action, string targetType, string targetId, object? details)`. Registered by `AddFoundryGateData`.
- `FoundryGate.Api.Services.Audit.IAuditService`: `Task<AuditLog> LogAsync(string action, string targetType, string targetId, object? details, CancellationToken ct)` · `Task<PagedResult<AuditLogEntryResponse>> QueryAsync(AuditLogQuery, PagedRequest, ct)`. (The `int? actorUserId` overload is **removed** — use `IAuditWriter`.)
- `ICurrentUserAccessor`: unchanged members; `GetRequiredUserAsync` now throws `UnauthorizedAccessException` (403) instead of `KeyNotFoundException` (404); `TryGetUserAsync` sees unsaved `Add`ed users.
- `Services/ApiServiceCollectionExtensions.AddFoundryGateApiServices()` (file/class renamed; method name unchanged).
- `ApiTestFactory` / `TestAuthHandler`: unchanged.

Issue #123, opened 2026-09-01
SystemConfiguration EntraTenantId is unused by identity-based Graph sync — remove or repurpose


kolatts:
Small cleanup discovered while implementing #40 / #110.

`SystemConfigurationKeys.EntraTenantId` (seeded reference data; documented in `reference/configuration.mdx` as "Azure AD tenant ID used by the Microsoft Graph SDK for bulk user sync") is **not read by anything**. With identity-based Graph auth (#110) the tenant is implied by the identity itself — the managed identity's home tenant in Azure, the Azure CLI login locally — and the only tenant setting the API consumes is `AzureAd:TenantId` for bearer-token validation.

## Ask

Either remove the key (reference-data seeder + `SystemConfigurationKeys` + the `configuration.mdx` row + any `SystemConfigurationKeysTests` expectation), or repurpose/re-describe it if a future feature (e.g. a cross-tenant B2B scenario) actually needs a tenant id distinct from `AzureAd:TenantId`. Today its documented purpose is misleading to a fork owner configuring the system.

kolatts:
Claiming this on branch `claude/hardening-entra-groups-config`, folded into #164's key retirement.

Issue #122, opened 2026-09-01
Entra user sync: should accountEnabled=false count as departed?


kolatts:
> Parent: #223 — Product decisions pending

Design question surfaced while implementing #40.

Spec §7.2 defines "departed" as *absent from the directory result*. In practice an off-boarded employee's account is usually **disabled** (`accountEnabled = false`) on their last day and only hard-deleted (which is what removes it from `appRoleAssignedTo`) weeks later — often 30 days. Under the current rule such a user stays `IsActive = true` (and, once #65 wires the full deprovision, keeps a working APIM subscription) for that whole window.

## Options

1. Treat `accountEnabled = false` as departed: add `accountEnabled` to the `$select` on `GET /users?$filter=id in (...)`, surface it on `EntraUser`, and have `EntraUserSyncService` route disabled accounts through the same deactivation branch (→ #65's `DeprovisionAsync(EntraDeparture, …)`).
2. Keep spec semantics but expose `accountEnabled` on `EntraUser` so `GET /users/me` (#28) can refuse to provision a disabled account.
3. Leave as is and document the lag.

Option 1 matches the security intent ("access follows the org chart") and costs one extra selected field; the reactivation rule stays unchanged (an admin must re-activate). Needs an owner decision before #65 finalises the deprovision trigger, since it changes which rows that trigger fires for.

kolatts:
Status from the users/lifecycle wave (#65): the deprovision trigger this question is about now exists and is **live**, so the lag this issue describes is no longer theoretical.

`EntraUserSyncService` calls `IUserLifecycleService.DeprovisionAsync(EntraDeparture, userId)` for every active user absent from the assigned-user list: APIM subscription deleted, `IsActive = false`, allocation hard-stopped, pending requests rejected. It still fires on **spec §7.2's rule only** — absent from `appRoleAssignedTo` — because that is all `EntraUser` carries. So option 3 ("leave as is and document the lag") is what shipped, by omission rather than by decision.

Concretely, with the pipeline now wired: an off-boarded employee whose account is *disabled* but whose app-role assignment survives keeps a **working gateway key** for the whole disable-to-delete window (often 30 days). That is a bigger deal than it was when departure only flipped a flag.

[...]

Issue #121, opened 2026-09-01
Entra user sync: expand group-based app role assignments (principalType=Group) to their user members


kolatts:
Discovered while implementing #40.

`GraphEntraDirectoryClient.ListAssignedUsersAsync` enumerates `servicePrincipals/{id}/appRoleAssignedTo` and keeps only assignments with `principalType = "User"`. Assignments made to a **group** (the common enterprise pattern: assign `SG_AI_Developers` to the FoundryGate enterprise application) appear as a single `principalType = "Group"` row and are currently **skipped with a warning log** (`Skipped {Count} app-role assignment(s) … whose principal is a group or service principal`). Members of such groups therefore never appear in `POST /users/sync` results and, worse, would be flagged `IsActive = false` if they had previously logged in via `GET /users/me`.

## Ask

- For each `principalType = "Group"` assignment, expand the group to its user members with `IEntraDirectoryClient.ListGroupMemberIdsAsync(groupObjectId, transitive: true, ct)` (already implemented for #41) and merge the ids into the assigned-user set before hydration.
- Decide whether `transitive: true` is the right default (nested groups) — Entra's own "assigned users" evaluation for app role assignments is transitive for security groups.
- Keep de-duplication (a user assigned directly *and* via a group counts once) and add a fake-directory test plus a stubbed-HTTP test for the expanded path.
- Requires `GroupMember.Read.All` on the API identity — already part of the role set documented in the live-validation issue.

Until this lands, forks must assign developers **individually** to the enterprise application, or accept that group-assigned developers are managed only through `GET /users/me` auto-provisioning (#28).

kolatts:
Update from the PR #124 review: until this lands, `EntraUserSyncService` **suspends departure detection** whenever the directory reports one or more group-principal app-role assignments — nobody is deactivated, adds/updates still happen, and the run returns `UserSyncResult.SkippedGroupAssignmentCount > 0` with the groups named in a Warning log and in the `users.synced` audit row (`departureDetectionSuspended: true`). The Graph client already returns the skipped groups as `EntraAssignedUsers.SkippedGroupAssignments` (`GroupObjectId`, `DisplayName`), so this issue's job is: expand each via `ListGroupMemberIdsAsync(groupObjectId, transitive: true, ct)`, merge into the user set, and **remove the suspension** so departure detection is re-enabled for group-assigned tenants. Re-point the fake-directory test that currently asserts zero deactivations.

kolatts:
[...]

Issue #120, opened 2026-09-01
Live-validate Entra user sync against a real tenant (Graph app roles for the API managed identity, POST /users/sync)


kolatts:
> Parent: #220 — Live validation of the dev environment

Follow-up to #40 / #110 (PR `claude/entra-graph-user-sync-40`). The Entra sync is fully covered by hermetic tests (in-memory directory fake for the sync service; stubbed HTTP for the real `GraphServiceClient` pipeline), but it has **not** been run against a real tenant. Owner action + one manual pass are needed before `Entra:Enabled = true` goes into a deployed environment.

## Owner setup (manual, not Bicep-able)

The API calls Graph as its own identity — no client secret (#110). Grant the Graph **application** roles to the API's user-assigned managed identity `id-foundrygate-api-{env}` (app-role assignments on the Microsoft Graph service principal, appId `00000003-0000-0000-c000-000000000000`; `az ad app permission` does **not** apply to managed identities):

| Graph app role | Why |
|---|---|
| `Application.Read.All` | resolve the FoundryGate service principal from `AzureAd:ClientId` and read `servicePrincipals/{id}/appRoleAssignedTo` (least privilege per Graph docs) |
| `User.Read.All` | `GET /users?$filter=id in (...)&$select=id,displayName,mail,userPrincipalName,employeeId` and `GET /users/{id}` |
| `GroupMember.ReadBasic.All` | `GET /groups/{id}/members` / `transitiveMembers` with `$select=id` (needed by #41) — the least-privileged application role per the Graph reference; `GroupMember.Read.All` is one notch over and unnecessary since only `id` is selected |

Example (Microsoft Graph PowerShell, as a Privileged Role Administrator / Global Administrator):

```powershell
Connect-MgGraph -Scopes AppRoleAssignment.ReadWrite.All, Application.Read.All
$graphSp = Get-MgServicePrincipal -Filter "appId eq '00000003-0000-0000-c000-000000000000'"
$apiMi   = Get-MgServicePrincipal -Filter "displayName eq 'id-foundrygate-api-dev'"
foreach ($role in 'Application.Read.All','User.Read.All','GroupMember.ReadBasic.All') {
  $appRole = $graphSp.AppRoles | Where-Object { $_.Value -eq $role -and $_.AllowedMemberTypes -contains 'Application' }
[...]

PR #124, opened 2026-09-01
feat(api): Entra bulk user sync over Microsoft Graph with identity-based auth (POST /users/sync)

### LINKED ISSUE #40: Implement bulk Entra user sync via Microsoft Graph SDK (POST /users/sync)
Part of #11.

Register an app-only `GraphServiceClient` using client credentials stored in Key Vault (not appsettings). `POST /users/sync` pages through `GET /users?$select=id,displayName,mail,employeeId` and for each Entra user: upserts `DisplayName`, `Email`, `EmployeeId`, `LastSyncedAt`; inserts new users with the default quota (no APIM key — key only provisioned on actual login). Users present in DB but absent from Entra are flagged `IsActive = false`. Returns `{ added, updated, deactivated }`. Write a `sync.bulk-users` audit log entry.

### LINKED ISSUE #110: Entra sync: use the API managed identity Graph application permissions instead of a GraphClientSecret
Design decision from #43/#44 that #40/#41 (Entra sync) should build on.

Spec §9.1/§11 and the #43 body list a `GraphClientSecret` Key Vault secret. The control-plane Bicep (`infra/modules/key-vault.bicep`) deliberately creates **no** secrets:

- a Bicep-managed secret is re-PUT with its template value on every deploy, wiping whatever an operator set out of band;
- the API already runs as a user-assigned managed identity (`id-foundrygate-api-{env}`, `AZURE_CLIENT_ID` set on the Container App), and Microsoft Graph application permissions (`User.Read.All`, `Group.Read.All`, `GroupMember.Read.All`) can be granted directly to that service principal — no client secret, nothing to rotate.

## Ask

- Implement the Graph client in the Entra sync using the app's `TokenCredential` (Imagile.Framework `AppTokenCredential`, already registered in `Program.cs`): `new GraphServiceClient(credential, ["https://graph.microsoft.com/.default"])`.
- Document the owner step: grant the Graph app roles to the API identity's service principal (app-role assignments against the Microsoft Graph service principal `00000003-0000-0000-c000-000000000000`, e.g. `New-MgServicePrincipalAppRoleAssignment` or the Graph `appRoleAssignments` endpoint; `az ad app permission` does not apply to managed identities). Locally the Azure CLI credential is used, so the developer's own delegated Graph consent applies.
- If a client secret ever becomes unavoidable, set it out of band in `kv-fg-{env}-{suffix}` and reference it as `@KeyVault(GraphClientSecret)` in appsettings — never add it to Bicep.

kolatts:
## Summary

Graph plumbing for every Entra-facing feature plus the first consumer, the admin bulk user sync.

- **`Services/Entra/IEntraDirectoryClient`** — the single seam to Microsoft Graph, with the shapes the users (#28) and groups (#41) waves will call:
  ```csharp
  Task<EntraUser?> GetUserAsync(string objectId, CancellationToken ct);                       // null on Graph 404
  Task<EntraAssignedUsers> ListAssignedUsersAsync(CancellationToken ct);                      // user assignees (hydrated, de-duplicated) + group assignees (not expanded, #121)
  IAsyncEnumerable<string> ListGroupMemberIdsAsync(string groupObjectId, bool transitive, CancellationToken ct); // user members only, every page
  public sealed record EntraUser(string ObjectId, string DisplayName, string Email, string? EmployeeId);
  public sealed record EntraAssignedUsers(IReadOnlyList<EntraUser> Users, IReadOnlyList<EntraGroupAssignment> SkippedGroupAssignments);
  public sealed record EntraGroupAssignment(string GroupObjectId, string DisplayName);
  ```
- **`GraphEntraDirectoryClient`** over `Microsoft.Graph` 6.5.0 (Graph.Core 4.0.1 / Kiota 2.0; `Azure.Core` already pinned, no `Azure.Identity` needed). Authenticates with the app's registered `TokenCredential` (`AppTokenCredential`: managed identity in cloud, Azure CLI locally) using `{GraphBaseUrl authority}/.default` — **no client secret, nothing in Key Vault** (#110).
- **`DisabledEntraDirectoryClient`** when `Entra:Enabled` is false (the default; local dev and the test host): every call → `ArgumentException` → `400` whose detail names `Entra:Enabled` and the three Graph roles to grant.
- **`EntraUserSyncService.SyncUsersAsync` → `UserSyncResult(AddedCount, UpdatedCount, DeactivatedCount, SkippedGroupAssignmentCount)`**, one `SaveChangesAsync`, one `users.synced` audit row via `IAuditService` (attributed to the calling admin). **When any app-role assignment is granted to a group, departure detection is suspended for the run** (adds/updates still happen, nobody is deactivated, groups named in a Warning log and the audit row) — see decision 1.
- **`Controllers/UsersController`** with only `POST /api/v1/users/sync` (admin, action-level policy so #28's `/me` can join the class).
- **`Configuration/EntraOptions`** (`Enabled`, `ServicePrincipalObjectId?`, `GraphBaseUrl`) — validated fail-fast only when enabled; `AppSettings.Entra` is the one added property. `appsettings.json` carries the disabled default.

Closes #40
Closes #110

## Design decisions (for the reviewer)

1. **Who is a FoundryGate user** = principals with an app-role assignment on the FoundryGate service principal (`servicePrincipals/{id}/appRoleAssignedTo`), per the #40 body — not the whole tenant as spec §7.2 step 1 literally says. Only `principalType = "User"` rows are hydrated; **group-principal assignments are returned as `SkippedGroupAssignments`, and while any exist the sync deactivates nobody** (review Major): a user assigned through a group is invisible until #121 expands them, so "absent from the user list" cannot mean "departed". The admin sees `skippedGroupAssignmentCount > 0` / `deactivatedCount: 0`; the groups are named in the Warning log and the audit row (`departureDetectionSuspended: true`). Service-principal assignees are dropped (not people). #121 re-enables departure detection by expanding the groups via `ListGroupMemberIdsAsync(transitive: true)`.
2. **Hydration path**: `appRoleAssignedTo` returns no user fields, so ids are hydrated with `GET /users?$filter=id in (...)&$select=id,displayName,mail,userPrincipalName,employeeId` in chunks of **15** (Graph's documented default cap on `in`). Verified against the live docs before choosing: `directoryObjects/getByIds` needs `Directory.Read.All` and has no `$select` (so no `employeeId`); N+1 `GET /users/{id}` was rejected on cost. Result: least-privilege roles are exactly `Application.Read.All` + `User.Read.All` (+ `GroupMember.ReadBasic.All` for #41 — only `id` is selected).
3. **Service principal**: resolved once from `AzureAd:ClientId` via `GET /servicePrincipals(appId='…')?$select=id` and cached for the process (failures are *not* cached); `Entra:ServicePrincipalObjectId` overrides for forks that assign developers to a different enterprise app. Missing SP → `InvalidOperationException` with a message naming both settings (a misconfiguration, so 500 + log, not a client-facing 4xx).
4. **Paging** uses `WithUrl(@odata.nextLink)` loops rather than `PageIterator`: the iterator drives a callback and can't `yield` into `IAsyncEnumerable` (group members) without buffering; the loop streams and is shorter. `$top=999` on every collection call. `ListAssignedUsersAsync` is buffered (`Task<EntraAssignedUsers>`) because the caller needs the group list before it can act on the users — a few thousand small records at most.
5. **Group members** use plain `/members` (+ `/transitiveMembers`) with `$select=id` and filter to users by `@odata.type` client-side — deliberately *not* the `/members/microsoft.graph.user` OData cast, which is an "advanced query" requiring `ConsistencyLevel: eventual` + `$count` and reads an eventually-consistent index (wrong for reconciliation).
6. **Transient faults**: the SDK's own pipeline (`GraphClientFactory` defaults → Kiota `RetryHandler`: 3 retries, back-off, honours `Retry-After` on 429/503/504). No Polly, no new package.
7. **Deprovision scope**: a departed user is only `IsActive = false` + audit; the full Entra-departure pipeline (APIM deletion, hard stop, request cancellation — plan #21 trigger B) is **#65**, called out in the XML docs of the service and in the plan. Rows are never deleted. A user who reappears is **not** auto-reactivated (plan #21: admin re-activates), and an admin-deactivated user still in Entra stays inactive.
8. **Counting**: every matched user counts as *updated* (`LastSyncedDate` is always stamped), so a no-change second run is `added 0 / updated N / deactivated 0`. Already-inactive absent users are not re-counted or touched.
9. **Safety guard**: an empty assigned-user list while active users exist locally → `409 ConflictException`, nothing saved — unless the emptiness is explained by a group assignment (then decision 1 applies: 200, nobody deactivated). That signal is almost always a wrong SP id or a missing Graph role, and with #65 it would otherwise delete every APIM subscription.
10. **Disabled → 400 via `ArgumentException`**: the handler's four-type map is untouched, as agreed. If a "feature disabled" → 503 mapping is preferred later it's a one-line follow-up.
11. **Testing the real Graph client without a tenant**: `GraphServiceClient(HttpClient, AnonymousAuthenticationProvider, baseUrl)` over a scripted `StubHttpMessageHandler` (URL-decoded route matching, records every request). This pins the wire shape (`$select`, `$top`, chunk sizes, next-link URLs) and the SP resolution/caching behaviour without touching `ApiTestFactory`. Endpoint tests swap the directory with `factory.WithWebHostBuilder(...)` — no shared-file edits.
12. Docs across the progressive chain now agree there is no Graph client secret: `architecture/overview.mdx` (one sentence), `getting-started/why-foundrygate.mdx` ("uses the app's own identity — no passwords or secrets to manage"), `getting-started/fork-guide.mdx` (no `GRAPH_CLIENT_SECRET`; Graph roles go on `id-foundrygate-api-{env}` after the infra deploy; new troubleshooting entries), `components/SystemDiagram.astro` (Key Vault node holds the APIM key-encryption key, no app secrets), and `reference/infrastructure.md` (from #111) gains the Graph app-role table under Role assignments. The stale `Azure__KeyVaultUri` row #111 left for "whichever PR lands second" is dropped from `configuration.mdx`.

## Verification (after the review fix pass, merged with `origin/main` incl. #111)

- `dotnet build FoundryGate.sln -c Release` — **0 warnings, 0 errors** (FoundryGate.Web compiles against the extended `UserSyncResult`)
- `dotnet test src/FoundryGate.Tests.Predeployment -c Release --no-build` — **170 passed, 0 failed** (126 on main + 44 new)
- `dotnet format FoundryGate.sln --verify-no-changes` — exit 0
- `docs-site`: `npm ci && npm run build` — exit 0
- `plans/11-entra-sync.md` Verification: the #40 items are checked (build, idempotent, new users appear, >1 page); #41 items left for the groups wave.

New tests: `EntraUserSyncServiceTests` (15 — incl. group-assignment suspends departure detection; group-only directory is not a conflict), `GraphEntraDirectoryClientTests` (9 incl. theory — group assignee reported, SP assignee dropped), `DisabledEntraDirectoryClientTests` (3), `EntraServiceCollectionExtensionsTests` (3), `EntraOptionsValidationTests` (8 incl. theory), `UsersSyncEndpointTests` (6).

## Deferred / filed

- **#120** — live validation against a real tenant: the owner checklist (grant `Application.Read.All`, `User.Read.All`, `GroupMember.ReadBasic.All` to `id-foundrygate-api-{env}` as Graph app-role assignments; PowerShell snippet; local delegated notes) and a manual test pass, now including the group-assignment suspension case. The same role table is in `reference/infrastructure.md` for #109's owner steps.
- **#121** — expand group-based app role assignments to their user members, which re-enables departure detection for group-assigned tenants.
- **#122** — should `accountEnabled = false` count as departed (disabled-then-deleted lag)?
- **#123** — `SystemConfiguration.EntraTenantId` is unused under identity-based Graph auth.

## Files

Api: `Configuration/EntraOptions.cs` (new), `Configuration/AppSettings.cs` (+1 property), `Controllers/UsersController.cs` (new), `Services/Entra/{IEntraDirectoryClient,GraphEntraDirectoryClient,DisabledEntraDirectoryClient,EntraUser,IEntraUserSyncService,EntraUserSyncService,EntraServiceCollectionExtensions}.cs` (new), `Services/ApiServiceCollectionExtensions.cs` (+1 line), `appsettings.json`, `FoundryGate.Api.csproj`; `Directory.Packages.props` (`Microsoft.Graph` 6.5.0).
Tests: `Support/{FakeEntraDirectoryClient,StubHttpMessageHandler}.cs`, `Api/Services/Entra/*Tests.cs`, `Api/Configuration/EntraOptionsValidationTests.cs`, `Api/Endpoints/UsersSyncEndpointTests.cs`.
Docs: `reference/api.md`, `reference/configuration.mdx`, `architecture/overview.mdx`, `plans/11-entra-sync.md`.


kolatts:
Heads-up to avoid a merge conflict: PR #111 also edits `docs-site/src/content/docs/reference/configuration.mdx` (the API env-var table rows above `Azure__KeyVaultUri`, and the Function App settings section) to match the infra reference. I stopped at the row directly above your `Graph__*` → `Entra__*` replacement so the hunks don't touch; the stale `Azure__KeyVaultUri` row is tracked for whichever PR lands second. No action needed unless you extend your edit upward.

kolatts:
**REQUEST CHANGES**

This is the cleanest Graph integration I've reviewed in a while — the `IEntraDirectoryClient` seam, identity-based auth, next-link streaming, the stubbed-HTTP wire-shape tests and the atomic audit+save are all exactly what CONVENTIONS.md asks for. One correctness gap blocks it: group-based app-role assignments are skipped with a log line, and then the sync deactivates everyone they covered. The 409 guard was built for precisely that class of misfire; it just doesn't cover the more common one.

Verified locally: build 0/0, 168/168 tests, `dotnet format` clean. Graph docs confirm `Application.Read.All` for `servicePrincipals(appId=)` + `appRoleAssignedTo`, `User.Read.All` as the application least-privilege for `/users`, and the 15-expression `in` cap.

---

**Issues**

- **Major** `src/FoundryGate.Api/Services/Entra/GraphEntraDirectoryClient.cs` line 113 (with `EntraUserSyncService.cs` line 83) — Assignments with `principalType = Group` are dropped and counted into a `LogWarning`, so in any tenant that assigns FoundryGate through a security group (the normal enterprise pattern, and how `FoundryGate.Admin` is usually granted) the first `POST /users/sync` flips every previously-provisioned user to `IsActive = false` with a 200 and a clean audit row. A warning log is not a safety guard — until #121 lands, treat a non-zero skipped count the way the empty-directory case is treated (409, nothing saved), or at least refuse the deactivation branch; there is no `/users/{id}/activate` yet to undo a mass flip, and with #65 this branch deletes APIM subscriptions.

- **Minor** `docs-site/src/components/SystemDiagram.astro` line 219 — The Key Vault node still lists "Graph client secret"; the whole point of #110 is that there isn't one, and CONVENTIONS.md's "Architecture diagrams" rule says the public diagram must show the current system.

- **Minor** `docs-site/src/content/docs/getting-started/why-foundrygate.mdx` line 196 — Still says "Microsoft Graph API (app-only credentials)" — the exact phrase this PR removed from `architecture/overview.mdx`. Docs invariant 2 wants every level of the progressive chain updated, and this is the simplest one.

- **Minor** `docs-site/src/content/docs/getting-started/fork-guide.mdx` lines 47–50 and 133 — A fork following this page mints a `GRAPH_CLIENT_SECRET` repo secret that nothing reads, and grants `User.Read.All` / `Group.Read.All` / `GroupMember.Read.All` to the app registration — missing `Application.Read.All` and aimed at the wrong principal (the roles go on the API's managed identity, per #120). The result is a 500 on the first sync.

- **Nit** `src/FoundryGate.Api/Services/Entra/DisabledEntraDirectoryClient.cs` line 23 — "(or set Entra:ServicePrincipalObjectId)" reads as an alternative to granting the roles; it only skips the SP lookup, `appRoleAssignedTo` still needs `Application.Read.All`.

- **Nit** `docs-site/src/content/docs/reference/configuration.mdx` line 45 (and #120's role table) — Graph documents `GroupMember.ReadBasic.All` as the least-privileged application permission for both `/members` and `/transitiveMembers`, and the client only selects `id`; `GroupMember.Read.All` is one notch over. Worth settling before #120 hardens the owner checklist.

---

**Not blocking, for the record.** Shared-file overlap with #127/#131 is adjacent-line only (`AppSettings.cs` property, the one `Add*Services()` line, `appsettings.json` sections); `Directory.Packages.props` doesn't collide. One thing to watch: #131 rewrites the `reference/configuration.mdx` rows right above the `Graph__*` rows this PR replaces and keeps the `Graph__ClientSecret` row in its version — whichever merges second needs to drop it. The `Users.Local.Where(u => u.UserId == 0)` audit side-channel and the sequential 15-id hydration (~100 round trips for 1,500 users inside one request) both scored under my threshold — noting them so they don't surprise anyone later.

kolatts:
Thanks — fix pass pushed (`fix(api): suspend departure detection when group-principal app-role assignments exist; review fixes for #124`), merged with `origin/main` (#111 in). Point by point:

**Major — group-principal assignments skipped, then everyone they covered deactivated.** Agreed, and you're right that a warning log is not a guard. Fixed at both layers:
- `IEntraDirectoryClient.ListAssignedUsersAsync` now returns `EntraAssignedUsers(Users, SkippedGroupAssignments)`: user assignees hydrated as before, group assignees surfaced by id + `principalDisplayName` (added to the `$select`), service-principal assignees dropped (they aren't people).
- `EntraUserSyncService` **suspends departure detection whenever `SkippedGroupAssignments` is non-empty**: adds/updates still happen, the deactivation branch is skipped entirely, the groups are named in a `LogWarning` and in the `users.synced` audit row (`departureDetectionSuspended: true`, `skippedGroupAssignments: [{groupObjectId, displayName}]`), and `UserSyncResult` gains `SkippedGroupAssignmentCount` so the admin sees why `deactivatedCount` is 0. Chose "refuse the deactivation branch" over a 409 so a group-assigned tenant can still import/refresh users. The 409 empty-directory guard no longer fires when the emptiness is explained by a group assignment (everyone assigned via the group is the expected zero-user case).
- Domain contract change checked: `FoundryGate.Web` only consumes `ApiCallResult<UserSyncResult>` generically, compiles unchanged.
- Pinned by `A_group_principal_assignment_suspends_departure_detection_but_not_adds_and_updates` (active user absent from the list stays active, `LastSyncedDate` untouched, new user still added, counts `(1, 1, 0, 2)`, audit details asserted) and `A_group_principal_assignment_with_no_user_assignees_is_not_a_conflict`; the stubbed-HTTP client test now asserts the Group assignee is reported and the ServicePrincipal one is not.
- Documented in `reference/api.md`, the interface XML docs, `fork-guide` troubleshooting, and `plans/11`. #121 updated: expansion re-enables departure detection.

**Minor — `SystemDiagram.astro` "Graph client secret".** Key Vault node now reads "`fg-apim-key-encryption` key · no app secrets (Graph uses the API identity)"; the figure's alt text said "Key Vault holds secrets" and was corrected too.

**Minor — `why-foundrygate.mdx` "app-only credentials".** Now: "reads your Entra directory through the Microsoft Graph API using the app's own identity — no passwords or secrets to manage."

**Minor — `fork-guide.mdx` `GRAPH_CLIENT_SECRET` / role list.** The secret row is gone. Step 2 no longer tells a fork to grant Graph permissions to the app registration; it says the roles (`Application.Read.All`, `User.Read.All`, `GroupMember.ReadBasic.All`) go on the API's user-assigned identity `id-foundrygate-api-{env}` after the infra deploy, linking the role table I added to `reference/infrastructure.md` (Role assignments) and the PowerShell snippet in #120. The "Graph sync fails with 403" troubleshooting entry (admin consent on the app) is replaced by three accurate ones: 400 (`Entra__Enabled` off), 500 with `Authorization_RequestDenied` (missing role on the identity — no restart needed, resolution isn't cached on failure), and `skippedGroupAssignmentCount > 0`.

**Nit — `DisabledEntraDirectoryClient` message.** Dropped "(or set Entra:ServicePrincipalObjectId)"; it now lists the three roles only, and a test asserts the setting is not presented as an alternative.

**Nit — `GroupMember.ReadBasic.All`.** Switched everywhere: code XML docs, `configuration.mdx`, `infrastructure.md`, `fork-guide`, the PR body and #120's role table + PowerShell snippet (with a note on why `Read.All` is one notch over since only `id` is selected).

**Merge / adjacent PRs.** Merged `origin/main`; the `configuration.mdx` hunks didn't collide and I dropped the stale `Azure__KeyVaultUri` row #111 left for the second lander. #131 hasn't merged yet — if it lands first, its `Graph__ClientSecret` row will need dropping on its side or on a follow-up merge here.

**For the record items.** Noted; the `Users.Local.Where(UserId == 0)` side-channel is gone as a by-product (added oids are now collected explicitly), the sequential hydration stays.

Gates after the fix pass: build 0 warnings; **170/170** Predeployment tests (126 + 44); `dotnet format --verify-no-changes` clean; docs `npm run build` clean.

Issue #119, opened 2026-09-01
Monthly reset Function cannot reference Api's IQuotaResolutionService — decide HTTP call vs moving resolution to Data


kolatts:
Discovered while implementing #32 / #33.

`plans/10-background-services.md` (#38) says the Functions host registers `IQuotaResolutionService` and calls `ResolveAsync` per active user for the new period. But per CONVENTIONS.md services live in `FoundryGate.Api/Services/<Area>/` (no separate Services project) and `FoundryGate.Functions` references **Data + Domain only** — it cannot reach `FoundryGate.Api.Services.Quota.QuotaResolutionService`, `GatewayTierMapper`, or `Gateway:Tiers` options.

Options for the #38 agent (pick one, record it in `fable-refactor-log.md`):

1. **Functions calls the Api** — `POST /api/v1/quota/reset` (already idempotent, audited once per run, preserves `TokensUsed`) with the Function's managed identity as an Entra client-credentials caller holding the `FoundryGate.Admin` app role. Keeps one implementation of the resolution rules; the Api must accept an app-only token (no `oid`-as-user — `ICurrentUserAccessor` currently 403s a principal without a `User` row, so the reset would need a system-actor path or `IAuditWriter.AddSystem`).
2. **Move resolution to Data** — relocate `QuotaResolutionService` + `GatewayTierMapper` + `BillingPeriod` consumers into `FoundryGate.Data` (options binding would have to move too), leaving Api with the controller/allocation service. Touches CONVENTIONS.md ("services live in Api").
3. Functions references Api as a project (pulls ASP.NET Core into the isolated worker) — probably not.

`BillingPeriod` was placed in `FoundryGate.Domain.Quota` (not `Api/Services/Quota` as the #32 brief suggested) precisely so the Functions host can share the period math regardless of which option wins.

kolatts:
Claimed by an implementation agent on `claude/functions-38-39-84-119`. Orchestrator ruling: option 2-and-a-half — a new `src/FoundryGate.Core` class library (references Data + Domain, zero ASP.NET) holds the services more than one host needs; CONVENTIONS §Solution structure is amended in the same PR.

Issue #118, opened 2026-09-01
Implement ApimGatewayTierSync: move the APIM subscription to the user's resolved tier product


kolatts:
Follow-up to #32 / #33 (quota resolution → tier-product mapping, per the #7 direction update).

`QuotaResolutionService` now resolves every allocation to an APIM tier product id (`GatewayTiers.Standard` / `Power` / `Unlimited`, caps from `Gateway:Tiers`) and, whenever the resolved tier for a user with a non-empty `User.ApimSubscriptionId` differs from the tier recorded on their previous allocation (or no previous allocation is known), calls the seam `FoundryGate.Api.Services.Quota.IGatewayTierSync.SyncAsync(User user, string tierProductId, CancellationToken ct)`. The registered implementation is `NullGatewayTierSync` (logs at Debug) — nothing reaches APIM yet.

## Ask

Implement `ApimGatewayTierSync : IGatewayTierSync` in the users/lifecycle wave (alongside the APIM key service from #36/#37):

- Move the developer's APIM subscription to the tier product (`PUT .../subscriptions/{sid}` with `scope = /products/{tierProductId}`); the subscription key must survive the move. Verify live that a re-scope does not rotate the primary key — if it does, the implementation has to create-on-new-product + delete-old and hand the new key to the user, and the tier change becomes a key event (audit `key.rotated`).
- Must be **idempotent**: `QuotaResolutionService` calls it *before* the caller's `SaveChangesAsync` (so a failed APIM move fails the request rather than leaving the DB claiming a tier the gateway does not enforce), and it may be called for a subscription already on the target product (the "previous tier" is inferred from the latest `QuotaAllocation.TierProductId`, which is unknown for a user's first allocation).
- Consider recording the product a subscription is currently on (e.g. a `User.TierProductId` column, owned by the keys/lifecycle wave) so the sync decision no longer has to infer it from allocation history.
- Replace the `NullGatewayTierSync` registration in `Services/Quota/QuotaServiceCollectionExtensions.cs` (`AddQuotaServices`).
- Tests: a hand-rolled fake of the APIM client, plus the existing `QuotaResolutionServiceTests` "sync invoked only on tier change" cases keep passing.

kolatts:
Claimed by an implementation agent on branch claude/users-lifecycle-28-29-65-66 (users lifecycle wave: IUserLifecycleService, GET /users/me, admin user endpoints, ApimGatewayTierSync).

PR #113, opened 2026-09-01
docs(site): current-architecture diagrams — system, deploy pipeline, gateway-centric quota/key

kolatts:
## Summary

Replaces the pre-gateway `ArchDiagram` with four token-driven, accessible Astro diagram components and brings the surrounding prose on `architecture/overview`, `architecture/feasibility` and `getting-started/why-foundrygate` in line with the gateway-centric architecture (CLAUDE.md ground truths, CONVENTIONS.md §Architecture diagrams / §CI/CD, plans 24/25, `infra/`).

Closes #89

## What each diagram shows

| Component | Renders on | Shows |
|---|---|---|
| `SystemDiagram.astro` (new, replaces `ArchDiagram.astro`) | `architecture/overview`, `architecture/feasibility` | **Data plane**: developer CLIs (Claude Code → `x-api-key`, Codex → `api-key`) → APIM v2 front doors `/anthropic/*` + `/openai/v1` → `fg-require-product` → tier products Standard 5M/20K · Power 20M/40K · Unlimited —/100K with per-product `llm-token-limit` (`403` monthly, `429 + Retry-After` TPM) → managed-identity → Anthropic pool (p1 eastus2 → p2 swedencentral, breaker 3×429/5xx per 30 s, retry ×2) + OpenAI backend → Foundry accounts with the `infra/main.bicep` default deployments → diagnostics → Log Analytics/App Insights. **Control plane**: Entra → Blazor UI (SWA) → Api (Container Apps) → Azure SQL (EF Core); Functions (monthly reset, usage reconciliation) + storage; Key Vault. **Bridge**: the Api's APIM Management API calls (create under tier product, `regeneratePrimaryKey`, move tier, delete) and the KQL rollup of `ApiManagementGatewayLlmLog` into the reconciliation Function. Solid azure = real-time enforcement path; dashed = out-of-band — the legend makes the CLAUDE.md distinction explicit. |
| `DeployPipeline.astro` (new) | `architecture/overview` (new "Deploy pipeline" section) | PR track: build (warnings=errors) → predeployment tests → format → Claude review → Bicep what-if comment. Merge track: infra → database (whitelist runner IP → 60 s firewall wait → TCP 1433 probe → dacpac deploy with data-loss blocking → seed reference → seed test non-prod) → Api/Functions/UI in parallel → postdeployment Playwright → summary. OIDC + GitHub Environment gates called out. |
[...]

Issue #108, opened 2026-09-01
Bind the Gateway__* configuration section set by infra (APIM/Foundry/workspace addressing) in Api and Functions


kolatts:
Follow-up to #43/#44. `infra/modules/control-plane.bicep` sets these environment variables / app settings on the Container App AND the Function App so the control plane never needs ARM resource ids typed into `SystemConfiguration` by hand:

| Key | Value |
|---|---|
| `Gateway__SubscriptionId` | subscription id |
| `Gateway__ResourceGroup` | `rg-foundrygate-{env}` |
| `Gateway__ApimName` | `apim-foundrygate-{env}-{suffix}` |
| `Gateway__ApimGatewayUrl` | `https://<apim>.azure-api.net` |
| `Gateway__LogAnalyticsWorkspaceId` | workspace ARM resource id (reconciliation KQL, #84) |
| `Gateway__KeyEncryptionKeyUri` | versionless Key Vault key URI for APIM key wrapping (#95); empty if `createKeyEncryptionKey=false` |
| `Gateway__FoundryAccountNames__{i}` | one per Foundry account, pool order (index 0 = primary) |

Also set: `AZURE_CLIENT_ID`, `Azure__KeyVaultUrl`, `ConnectionStrings__FoundryGate`, `AzureAd__{Instance,TenantId,ClientId,Audience}`, `Cors__AllowedOrigins__0`, `OpenTelemetry__{Enabled,ConnectionString}`; Functions additionally `AZURE_FUNCTIONS_ENVIRONMENT` + `DOTNET_ENVIRONMENT` (= `qa`/`prod`).

## Ask

- Add a `GatewayOptions` class to `FoundryGate.Api/Configuration/AppSettings.cs` (and the Functions equivalent) bound from `Gateway`, validated with `ValidateRecursively()` — the section is optional locally, so carry `Enabled`/nullable semantics consistent with CONVENTIONS.md.
- Have the APIM key service (#36/#37), Foundry deployment service (#61) and reconciliation function (#84) read these instead of (or as the seed for) `SystemConfiguration["ApimResourceId"]`/`["FoundryResourceId"]`; update `docs-site/src/content/docs/reference/configuration.mdx` accordingly.
- If the names should change, change them in `modules/control-plane.bicep` and `docs-site/src/content/docs/reference/infrastructure.md` in the same PR — the infra side is the source of truth for the keys.

kolatts:
[...]

PR #127, opened 2026-09-01
feat(quota): five-level resolution → tier-product mapping, allocation endpoints, manual reset

### LINKED ISSUE #32: Implement five-level quota resolution logic and write to QuotaAllocation

### LINKED ISSUE #33: Implement QuotaAllocation read endpoints and manual reset
Part of #7.

`GET /quota/allocations` (admin, paged) returns all current-period allocations with user info. `GET /quota/allocations/me` returns the authenticated user's own allocation. `GET /quota/allocations/{userId}` (admin) returns a specific user's allocation. `POST /quota/reset` (admin) is idempotent — runs the full monthly reset for all active users against the current calendar month; safe to call multiple times in a month without creating duplicates or zeroing `TokensUsed` on an already-running period.

kolatts:
## Summary

Implements the quota epic's two sub-issues under the #7 direction update (enforcement is APIM `llm-token-limit` on tier **products**; no `cache-store-value` push; suspension is offboarding-only).

- **`Services/Quota/QuotaResolutionService`** — five levels, first match wins: `User.IsUnlimited` → `User.MonthlyTokenQuota` → any `Group.IsUnlimited` → max `Group.MonthlyTokenQuota` → `SystemConfiguration[DefaultMonthlyTokenQuota]`. User-level settings always beat group-level ones (pinned by test: a user override beats an unlimited group). Upserts `QuotaAllocation` **without saving**: new rows `TokensUsed = 0`; existing rows keep `TokensUsed` / `IsHardStopped` / `ResetDate`. The system default is parsed defensively — missing or non-numeric throws `ConfigurationValidationException` with a fix-it message, never a silent 0. `ResolveManyAsync` bulk-loads for the reset.
- **Tier mapping** — `GatewayTierMapper` over `Gateway:Tiers` (`Configuration/GatewayTierOptions.cs`; validated fail-fast: non-empty, ids ⊆ `GatewayTiers.All`, no duplicates, exactly one unlimited, ≥1 finite). Unlimited → `unlimited`; else the smallest tier whose cap ≥ quota (exactly at the cap fits); above every finite cap → largest finite tier flagged **`IsGatewayCapped`** (the gateway 403s at the tier cap, not the numeric quota — user-visible, documented). Defaults live in `appsettings.json` and `GatewayTierOptionsTests` parity-checks them against `infra/main.bicep`'s `quotaTiers` (ids **and** caps, in order).
- **`IGatewayTierSync` seam** (`SyncAsync(User, string tierProductId, ct)`) — invoked *before* the caller saves when a user with an `ApimSubscriptionId` resolves to a tier different from their most recent allocation's (or none is known). `NullGatewayTierSync` (Debug log) registered; the APIM move is **#118**.
- **Schema** — `QuotaAllocation` gains `ResolvedLevelType` (`FoundryGate.Domain.Quota.QuotaLevelType`, int), `TierProductId` (`nvarchar(64)`), `IsGatewayCapped`, plus a `(PeriodYear, PeriodMonth)` index for the current-period list; `QuotaAllocations.sql` matches (SchemaParityTests). `TestDataSeeder` fills the new columns.
- **Contracts** — `QuotaAllocationResponse` gains `UserDisplayName`, `UserEmail`, `ResolvedLevelType`, `TierProductId`, `IsGatewayCapped`; `QuotaResetResult` gains `PeriodYear`/`PeriodMonth`. Web compiles unchanged (it only deserializes these).
- **`Controllers/QuotaController`** (`/api/v1/quota`): `GET allocations` (admin, paged, current period, ordered by display name), `GET allocations/me` (resolves + creates on first call of the month; 403 until `GET /users/me` provisioned the caller), `GET allocations/{userId}` (admin, read-only → 404 when no row yet), `POST reset` (admin, idempotent: re-resolves every **active** user for the current UTC month, preserves `TokensUsed`, clears `IsHardStopped`, stamps `ResetDate`, **one** `quota.reset` audit row per run attributed to the admin, committed atomically).
- **`FoundryGate.Domain.Quota.BillingPeriod`** — UTC calendar month from `TimeProvider`; placed in Domain (not `Api/Services/Quota` as briefed) so the Functions host can share it.

## Design decisions for the reviewer

1. **`Gateway:Tiers` defaults are in `appsettings.json`, not C# initializers.** The configuration binder *appends* configured list items to a pre-populated list, so C# defaults + a fork's override would silently yield duplicate tiers. Empty-by-default + non-empty validation + the bicep parity test is the honest shape. Commented on #108 so `GatewayOptions` folds `Tiers` in under the same section key.
2. **Tier sync runs before the DB save.** A failed gateway move fails the request; the reverse order (DB saved, gateway not moved) would leave the DB claiming a tier the gateway doesn't enforce. Implementations must be idempotent (#118 body says so).
3. **"Previous tier" is inferred from allocation history** (same-period row, else most recent earlier row, else unknown → sync). Nothing records which product a subscription is actually on; #118 suggests a `User.TierProductId` owned by the keys/lifecycle wave.
4. **Admin `GET allocations/{userId}` is read-only (404), `/me` auto-creates.** Only `/me` was briefed to auto-create; an admin GET mutating state felt wrong. Easy to flip if the #51 user-detail page wants it.
5. **`/me` auto-create is not audited** — the row is derived state, not an admin action; `/me` races on the unique index are handled by detaching and re-reading the winner.
6. **`ResolveAsync(int userId, BillingPeriod period, ct)`** rather than `(int, int, int, ct)` — one value type instead of two loose ints; `BillingPeriod.Current(timeProvider)` is the only sanctioned way to get "this month".
7. Discovered: **Functions cannot reference Api's `IQuotaResolutionService`** as plans/10 assumes (Functions references Data+Domain only) — filed **#119** with the options.

## Verification

- `dotnet build FoundryGate.sln -c Release` — **0 warnings, 0 errors** (also after merging `origin/main` @ #113)
- `dotnet test src/FoundryGate.Tests.Predeployment -c Release --no-build` — **210 passed** (126 on main + 84 new: `QuotaResolutionServiceTests` 26, `QuotaAllocationServiceTests` 13, `QuotaEndpointTests` 17, `GatewayTierMapperTests` 11, `GatewayTierOptionsTests` 9, `BillingPeriodTests` 8 — theory rows counted individually)
- `docs-site`: `npm ci && npm run build` — 12 pages built
- `plans/07-quota.md` Verification: checked / superseded-with-reason / deferred to #118, #119

## Docs

Prose-only additions kept separate from the diagram components #113 rewrote: `architecture/overview` (tier mapping + reset semantics paragraphs), `reference/api` (`/quota` semantics and response fields), `reference/configuration` (`Gateway:Tiers`). Landing page unchanged — nothing on it became untrue.

Closes #32
Closes #33


kolatts:
**REQUEST CHANGES**

The code is the best-tested thing in this repo so far — precedence, upsert, audit atomicity and the tier boundaries are all pinned and I couldn't break them. What blocks it is the docs claim: "landing page unchanged, nothing became untrue" isn't true. The mapper rounds every numeric quota *up* to a tier cap, so the 500K/1M/2M developers on the landing page are all enforced at 5M — "when the budget runs out, requests stop" no longer describes what the gateway does.

Verified locally on the branch: build clean (the Functions `WorkerExtensions` MSB3030 in my worktree is a long-path artefact — Functions is untouched and CI is green), 210/210 predeployment tests, `dotnet format` clean, bicep parity test passes on the Linux runner, and I probed two of the design claims (below).

---

**Issues**

- **Major** `docs-site/src/pages/index.astro` L987/L995 + `getting-started/why-foundrygate.mdx` L41/L61 + `reference/api.md` — The docs state the over-cap direction (`isGatewayCapped`) but not the common one: a quota *below* a cap is under-enforced (1M → Standard → the gateway stops them at 5M; a `0` quota lands on Standard too). CLAUDE.md's docs invariants want this true at every level — the honest framing at the simple levels is probably "budget tier", and the reference should say plainly that the gateway enforces the tier cap in both directions and `allocatedTokens` is the dashboard/reconciliation number. Open question for the product: does a free-form numeric quota still earn its keep, or should admins pick a tier?

- **Minor** `src/FoundryGate.Api/Services/Quota/QuotaAllocationService.cs` L70 — `ResetAsync` filters `IsActive`; `GetMyAllocationAsync` doesn't, so a deactivated user with a still-valid token gets a fresh allocation (and, once #118 lands, a tier sync onto a product). Same 403 as the no-row path, or reference the issue that owns the `IsActive` gate (#65?).

- **Minor** `src/FoundryGate.Api/Services/Quota/QuotaAllocationService.cs` L128 — "Idempotent" holds sequentially only: two admins (or a reset racing a `/me`) both `Add` the same rows and the loser is a bare 500 off the unique index. `/me` got the detach/re-read guard; reset didn't — catching to `ConflictException` (409) would be enough.

- **Minor** `src/FoundryGate.Api/Configuration/GatewayTierOptions.cs` L36 — Verified by probe: `ValidateRecursively` does not walk `List<GatewayTier>` items, so `[Range]`/`[StringLength]` on `GatewayTier` never run and a tier with `MonthlyTokenQuota = -5000000` starts fine as a finite tier no quota ever selects (every user silently lands on Power). `Validate` already iterates the items — check the range there.

- **Minor** `src/FoundryGate.Api/Services/Quota/QuotaResolutionService.cs` L90 — The prior-tier query materialises *every* historical allocation for every user without a current-period row, which is all active users × all months on the first reset of each month. Inferring the tier from history is fine to defer to #118's `User.TierProductId`, but the fetch should be bounded (previous period only, or a windowed `TOP 1` per user).

- **Minor** `src/FoundryGate.Api/Configuration/AppSettings.cs` L46 — Coordination, not a bug: PR #131 adds `public GatewayOptions Gateway` bound from the same `Gateway` section. Whichever merges second gets a duplicate-member compile error and has to fold `Tiers` into `GatewayOptions` (exactly what your #108 comment said — just flagging that it's now real, not future). The `Gateway__*` env vars from #111 compose fine with `GatewayTierOptions` today; the binder ignores unknown keys.

- **Nit** `src/FoundryGate.Domain/Quota/BillingPeriod.cs` L29/L35, `Services/Quota/QuotaResolution.cs` L21–33 — `IsBefore`, `Start` and all five `QuotaResolution` shortcuts have no production callers (`IsBefore` can't even be used at its stated site — it isn't EF-translatable, which is why the two queries inline it).

---

Things I checked and won't argue with: user-level beats group-level (incl. finite override over unlimited group), zero groups / null-quota groups fall through correctly, `TokensUsed`/`IsHardStopped`/`ResetDate` survive re-resolve, audit row lands in the same `SaveChanges`, tier sync before save is the right failure order, `BillingPeriod` as a `readonly record struct` in Domain is fine (TimeProvider is BCL), the appsettings-not-C#-defaults rationale is correct (probed: the binder appends to a pre-populated list), `GenerateSmartDefaults = true` in the CLI means the new NOT NULL columns deploy over existing rows, and the `QuotaAllocations.sql` mirror passes schema parity. #118/#119 are the right places for the deferred pieces.

kolatts:
## Author response — fix pass pushed as 4b56a73

Thanks — the Major was right, and the orchestrator ruled on the product question: **a developer's monthly budget IS a gateway tier.** Recorded as **D-013** in `fable-refactor-log.md` (options considered: free-form + round-up, more tiers, suspension-based soft enforcement; why "quota must equal a tier cap" won).

### Major — docs honesty / rounding-up semantics → **fixed (semantics changed, not just docs)**
- `GatewayTierMapper.Map` is now an **exact match**: a finite quota equal to a cap → that tier; unlimited → unlimited tier. A value matching no cap is never rejected on read — it is enforced at the **next tier up** (or the largest finite tier) and flagged `IsGatewayCapped`, whose doc now reads "did not match a configured tier; enforced at the next tier up". `1M → Standard, capped` is pinned as a legacy case, alongside `0`, `5M+1`, `20M+1`.
- New write-path guard `GatewayTierMapper.EnsureValidQuota(long? quota, string paramName)` → `ArgumentException` (400) whose message lists the allowed values (`IsValidQuota` / `Describe()` alongside). CONVENTIONS gained a "Quota values are tiers" bullet under API service/controller conventions so #29 / #30 / #35 call it before persisting.
- `GET /api/v1/quota/tiers` (any authenticated user) → `QuotaTierResponse(ProductId, DisplayName, MonthlyTokenQuota?, IsUnlimited)`, finite tiers ascending then unlimited; `GatewayTier` gained `DisplayName` (appsettings: Standard / Power / Unlimited).
- Seeds: `DefaultMonthlyTokenQuota` → `5000000` (= Standard). `TestDataSeeder` users/group draw from 5M / 20M / unlimited only; `TestDataSeederTests` pins that and that no seeded allocation is capped.
- Docs at every level: landing page demo table is 5M / 20M / Unlimited plus one plain sentence ("Budgets come in a few set sizes…"); "requests stop the second the budget is spent" stays because it is true again. `why-foundrygate`: "alex gets 5 million tokens", the card says "budget tier — Standard, Power or Unlimited", and the enforcement paragraph says a budget *is* a tier. `overview`: "A budget is a tier" paragraph. `reference/api`: the exact rule (must equal a tier cap; 400 otherwise; legacy → next tier up + flag) and the tiers endpoint. `reference/configuration`: default 5M, DisplayName column, and "adding a tier = bicep `quotaTiers` + `Gateway:Tiers` + `GatewayTiers.All`".

### Minor — `/me` ignores `IsActive` → **fixed**
`GetMyAllocationAsync` throws `UnauthorizedAccessException` (403, "deactivated … POST /users/{id}/activate") before any lookup; creates no row and triggers no tier sync. Service test + endpoint test.

### Minor — reset not race-safe → **fixed (adopt, not 409)**
On `DbUpdateException` with pending Adds, `ResetAsync` detaches its Added rows whose (user, period) now exists, copies the resolution outputs onto the winners (their `TokensUsed` intact), stamps them, and saves again — the failed save left the audit row pending, so it is still one atomic unit with exactly one audit row. Test uses a hand-rolled `RacingResolutionService` that inserts the winner through a **second connection** to the shared-cache SQLite DB between resolve and save. Went with adopt-and-continue per the orchestrator rather than 409, since two admins clicking reset should both succeed.

### Minor — `ValidateRecursively` doesn't walk `List<GatewayTier>` → **fixed**
`GatewayTierOptions.Validate` now checks each item (blank `ProductId`, cap outside `[0, MaxMonthlyTokenQuota]`) with `Tiers[i].Member` paths. Pinned at both levels: `Validator` on the options and `AppSettings.ValidateRecursively()` with a `-5000000` cap — the exact probe from the review.

### Minor — prior-tier query materialises all history → **fixed**
`ResolveManyAsync` now issues one `GroupBy(UserId).Select(g => g.OrderByDescending(...).First())` — EF translates it to a ROW_NUMBER window, one row per user, and the SQLite harness proves the translation (bob has July=Power and August=Standard rows; August wins). Single-user `ResolveAsync` was already `FirstOrDefaultAsync`.

### Minor — `AppSettings.Gateway` collision with #131/#133 → **acknowledged, deferred to the rebase**
Per the merge order (#124 → #131 → #133 → #127) I merge last and will fold `Tiers` into the unified `GatewayOptions` (keeping `GatewayTier`, the item validation, and the `Gateway:Tiers` key) when the orchestrator signals main is ready. Nothing else in this PR depends on the class name.

### Nit — dead `BillingPeriod.IsBefore`/`Start` and `QuotaResolution` shortcuts → **removed**
Tests read `result.Allocation.*` directly. `QuotaResolution` is now just `(Allocation, IsNew, PreviousTierProductId, TierSyncRequested)`.

### Also
- Audit details for `quota.reset` dropped `createdCount` (not knowable before the save under the race handling); kept `usersResetCount / periodYear / periodMonth / tierSyncCount`.

**Gates:** `dotnet build -c Release` 0 warnings; **227/227** predeployment tests (+17); `dotnet format --verify-no-changes` clean; `docs-site` build 12 pages. CI on 4b56a73 running.

kolatts:
## Rebase onto main (#131 + #133 merged) — a51eb99

Merged `origin/main` and folded the tier configuration into the unified `GatewayOptions`, per the coordination note in the review:

- **`GatewayOptions.Tiers`** (`List<GatewayTier>`, `[Required]`, never optional) now lives on main's class alongside `SubscriptionId / ResourceGroup / ApimName / KeyEncryptionKeyUri / FoundryAccountNames`; `GatewayTierOptions` is deleted. `Validate()` runs both halves — the Foundry/APIM partial-addressing rules unchanged, plus the tier rules (non-empty, ids ⊆ `GatewayTiers.All`, no duplicates, item-level cap range, exactly one unlimited, ≥1 finite). The `Gateway:Tiers` config path and the `appsettings.json` defaults are unchanged.
- `GatewayTierMapper`, `AddQuotaServices`, and the test helper take `GatewayOptions`; `GatewayTierOptionsTests` → `GatewayOptionsTiersTests` (+ a "tier rules don't disturb fully valid addressing" case).
- `AppSettingsValidationTests` covers the merged class from both sides: all of main's Foundry/APIM cases (the "absent section is valid" case is now "absent *addressing* is valid" — tiers always ship in `appsettings.json`, so an entirely absent `Gateway` section is a startup failure with the fix in the message), plus missing-tiers, negative-cap-via-`ValidateRecursively`, and combined addressing+tier errors reported together.
- `ApiTestFactory`'s `UseSetting` Gateway addressing is untouched; tiers reach the test host through the real `appsettings.json`, as before.
- No `FeatureNotConfiguredException` path needed here — tiers are required, not a feature toggle.
- Docs: `reference/configuration` keeps main's `Gateway__*` table and notes `Tiers` is the one non-optional member of the section; CONVENTIONS keeps both new bullets.

**Gates on the merged tree:** `dotnet build -c Release` 0 warnings; **464/464** predeployment tests; `dotnet format --verify-no-changes` clean; docs-site build 12 pages. CI on a51eb99 running.

kolatts:
**Addendum — merged main again after #146 (fb004a8).** Clean merge, no conflicts. `QuotaController` has no request-body record, so the init-property-record convention has nothing to apply here; the `/api/v1/quota/*` routes were already lowercase. Gates on fb004a8: `dotnet build -c Release` 0 warnings; **475/475** predeployment tests; `dotnet format --verify-no-changes` clean; docs-site build 12 pages.

Issue #132, opened 2026-09-01
Live-validate APIM key lifecycle (provision/rotate/reveal/revoke) and Key Vault key wrapping against real APIM + Key Vault


kolatts:
> Parent: #220 — Live validation of the dev environment

Follow-up to the #36/#37/#95 keys wave. Everything above the two Azure seams — `IApimManagementClient` (ARM) and `IKeyProtector` (Key Vault `CryptographyClient`) — is covered by the predeployment suite with fakes; the ARM and Key Vault calls themselves have never run against real resources. Needs a deployed control plane (#105) with the API identity holding **API Management Service Contributor** on the APIM instance and **Key Vault Crypto User** on the vault (`control-plane-rbac.bicep`), and `Gateway__SubscriptionId/ResourceGroup/ApimName/KeyEncryptionKeyUri` + `KeyProtection__Provider=KeyVault` set on the Container App (#108).

## Manual checklist

Run as an admin who has already called `GET /users/me`; target a developer user `{id}` who has also called `GET /users/me`.

### Key Vault protector
- [ ] API starts with `KeyProtection__Provider=KeyVault` and the versionless `Gateway__KeyEncryptionKeyUri`; starting with the URI blank fails with the `ConfigurationValidationException` naming `Gateway:KeyEncryptionKeyUri`.
- [ ] Starting a `qa`/`prod` host with `KeyProtection__Provider=DataProtection` refuses to boot.
- [ ] After a provision, `SELECT ApimSubscriptionKey FROM Users WHERE UserId = {id}` is `kv1:https://<vault>/keys/fg-apim-key-encryption/<version>:<base64>` — 384 base64-decoded bytes for the RSA-3072 key, total length < 1000.
- [ ] Rotate the Key Vault key (new version). `POST /keys/me/reveal` on the OLD row still returns the right key (unwrap pins the recorded version); a subsequent `POST /keys/me/rotate` writes an envelope carrying the NEW version id.
- [ ] Disable the old key version in Key Vault → reveal of a row still on it returns 500 (logged `RequestFailedException`), never a garbage key. Re-enable.

### APIM management client
- [ ] `POST /keys/{id}/provision` → 200 with `plaintextKey`; APIM portal shows subscription `foundrygate-{id}`, display name `FoundryGate {email}`, scope `.../products/standard`, state `active`. Calling the gateway with that key succeeds; the developer's `remaining-tokens` headers reflect the `standard` product policy.
[...]

PR #111, opened 2026-09-01
feat(infra): control plane modules (SQL, Container Apps, SWA, Functions, Key Vault, RBAC) behind deployControlPlane

kolatts:
## Summary

Adds the FoundryGate **control plane** to the existing subscription-scope `infra/main.bicep` — SQL, Container Apps (API), Static Web App (UI), Functions Flex Consumption, Key Vault, container registry, two user-assigned identities and the role assignments between them — behind a `deployControlPlane` toggle that defaults to `false`, so the gateway-only deployment the team validated (`test.bicepparam`) is byte-for-byte unchanged in behaviour. New `dev.bicepparam` / `prod.bicepparam` turn it on. The gateway modules are untouched apart from one added `workspaceName` output on `monitoring.bicep`.

Closes #43
Closes #44

## What's in the box

| Module | Notes |
|---|---|
| `modules/control-plane.bicep` | RG-scope orchestrator; owns the naming convention; wires leaf modules in dependency order (identities → KV/ACR/storage/SQL/SWA → RBAC → Container App + Function App) |
| `modules/sql.bicep` | **Entra-only** (`azureADOnlyAuthentication: true`, admin = Entra group; no SQL login/password anywhere — CONVENTIONS.md supersedes the issue bodies' "SQL admin password"). Serverless `GP_S_Gen5` dev / provisioned `GP_Gen5_2` + Geo backups prod. Only `AllowAllWindowsAzureIps` declared; runner/dev IP rules belong to CLI `ip setup` (#96) and are intentionally undeclared so re-runs never wipe them |
| `modules/container-app.bicep` | Consumption env with logs via **diagnostic setting** (`destination: azure-monitor`) instead of the workspace shared key; API app with UAMI, identity-based ACR pull, `/health` + `/health/ready` probes, min 1 replica, 8080, placeholder image for the bootstrap run |
| `modules/function-app.bicep` + `storage-account.bicep` | Flex Consumption (`FC1`, .NET 10 isolated), identity-based `AzureWebJobsStorage__*` and deployment container, **storage shared-key access disabled** |
| `modules/key-vault.bicep` | RBAC mode, soft delete, purge protection param (prod on), optional `fg-apim-key-encryption` RSA-3072 key for #95 (URI exposed as output + `Gateway__KeyEncryptionKeyUri`). **No secrets created** — see design decisions |
[...]

Issue #84, opened 2026-09-01
Emit per-developer token metrics and reconcile usage from ApiManagementGatewayLlmLog


kolatts:
Metrics + reconciliation feed (part of #81). Two layers with different guarantees:

1. **Dashboards (near-real-time)**: `llm-emit-token-metric` policy with built-in Subscription ID / Product ID dimensions → App Insights custom metrics. Caveat: APIM caps each dimension at **100 unique values and silently discards beyond** — fine for small orgs and aggregate views, not billing.
2. **Billing-grade (authoritative)**: enable the "Logs related to generative AI gateway" diagnostic setting → **`ApiManagementGatewayLlmLog`** table (per-request prompt/completion/total tokens, model). Join to `ApiManagementGatewayLogs` on `CorrelationId` for `ApimSubscriptionId`. The usage-sync Function (#39) reads this KQL rollup instead of App Insights REST, updates `QuotaAllocation.TokensUsed`, and detects drift against gateway-enforced quotas (enforcement itself is now policy-side — see #82).
3. **Cost attribution**: Claude bills as a single aggregate Marketplace CCU meter — Azure Cost Management cannot break Claude spend down per deployment or user. Compute per-developer cost as tokens × rate card (rates stored in `SystemConfiguration`); OpenAI models additionally support the `deployment` billing tag in Cost Management for cross-checking.

Acceptance:
- [ ] KQL query returning per-subscription prompt/completion tokens for an arbitrary period, checked into the repo
- [ ] Usage-sync Function consumes it and populates `QuotaAllocation.TokensUsed`
- [ ] Dashboard shows per-developer tokens and estimated cost (rate card × tokens)
- [ ] Streaming-interruption undercount caveat documented

kolatts:
Claimed by an implementation agent on `claude/functions-38-39-84-119` (items 1, 2 and 4 — checked-in KQL, usage-sync Function, streaming caveat. Item 3, cost attribution, is being split out as a follow-up issue).

PR #131, opened 2026-09-01
feat(api): /foundry deployment endpoints over ARM with create-once safety

kolatts:
## Summary

Admins can list, create and delete Azure AI Foundry model deployments from the API; developers can list the models they can point a CLI at. ARM sits behind a thin seam so nothing in the test suite touches Azure, and the service enforces the create-once rules the E-006/E-007 incidents taught us.

Closes #61

> **Review fix pass (`668ad96`):** post-commit audit/save on `CancellationToken.None`; Anthropic **deletes** refused too (infra owns Claude deployments end to end); new `FeatureNotConfiguredException` → **503** for an absent `Gateway` section or a configured account Azure does not have (#133's `UnconfiguredApimManagementClient` adopts it on rebase); ARM account-not-found vs deployment-not-found distinguished, `/models` skips a missing region with a Warning; 30 s `IMemoryCache` on `/models`; docs at all three levels; CONVENTIONS bullets (commit point, 503 exception, `UseSetting`). Merged with main after #111 and #124.

**Ownership split:** Claude (Anthropic-format) deployments are managed by the infrastructure deploy end to end; the API lists everything and manages OpenAI-format deployments.

| Method | Path | Auth | Result |
|---|---|---|---|
| `GET` | `/api/v1/foundry/deployments` | Admin | every deployment in every configured account, primary account first; always live |
| `GET` | `/api/v1/foundry/deployments/{accountName}/{deploymentName}` | Admin | one deployment (poll target; `404`) |
| `POST` | `/api/v1/foundry/deployments` | Admin | `201` + `Location`; `409` existing name; `400` Anthropic / unconfigured account / validation; `503` Gateway section absent or account missing in Azure |
| `DELETE` | `/api/v1/foundry/deployments/{accountName}/{deploymentName}` | Admin | `204`; `404`; `400` Anthropic (infra-owned); never recreates |
| `GET` | `/api/v1/foundry/models` | Any | distinct deployment names with model/version/format/state; 30 s cache; a missing region is skipped, not fatal |

## What's in the box

[...]

PR #133, opened 2026-09-01
feat(api): APIM subscription keys — provisioning, rotation, reveal, revocation, encryption at rest

### LINKED ISSUE #36: Implement APIM key provisioning using Azure SDK and Managed Identity
Part of #9.

Create an `IApimKeyService` backed by `Azure.ResourceManager.ApiManagement`. In `POST /keys/{userId}/provision` (admin) and auto-provision from first login: call APIM Management to create a subscription named `foundrygate-{userId}` scoped to the configured product. Retrieve the primary key, encrypt it using Azure Key Vault key wrapping (`CryptographyClient`), store the ciphertext in `User.ApimSubscriptionKey`, and store the APIM subscription resource ID in `User.ApimSubscriptionId`. The key is never stored in plaintext.

### LINKED ISSUE #37: Implement key rotation (self-service and admin) and revocation
Part of #9.

`POST /keys/me/rotate` and `POST /keys/{userId}/rotate` (admin): call APIM `regeneratePrimaryKey`, decrypt-then-re-encrypt the new key via Key Vault, and update `User.ApimSubscriptionKey`. `DELETE /keys/{userId}` and the deactivate flow: call APIM to delete the subscription, set `ApimSubscriptionId` and `ApimSubscriptionKey` to null, mark `IsActive = false`, and hard-stop the current `QuotaAllocation`. `GET /keys/me` returns the masked key (last 4 chars visible); a separate "reveal" call decrypts and returns the full key once — not stored in session.

### LINKED ISSUE #95: Encrypt ApimSubscriptionKey at rest before real APIM keys land in the User table
## Context

`User.ApimSubscriptionKey` (`src/FoundryGate.Data/Entities/User.cs`, added in PR #92 / #22) is currently a plain `nvarchar` column. Spec §11 (Security Considerations) calls for APIM subscription keys to be encrypted at rest; the column's own doc comment already flags this:

> Encrypted at rest (encryption is applied by the service layer, not this column).

No encryption exists yet — the column just stores whatever string is written to it. That's fine today because nothing writes real subscription keys yet (key provisioning is out of scope for #22/#23, and lands with the quota/key service work in #32 and related issues).

## Ask

Before any code path actually provisions/writes a real APIM subscription key into this column:

- Decide the encryption approach (e.g. Azure SQL Always Encrypted / column encryption, application-level encryption via Key Vault-backed key, or equivalent) consistent with CONVENTIONS.md's Key Vault/`TokenCredential` patterns.
- Apply it to `User.ApimSubscriptionKey` specifically (and audit whether `ApimSubscriptionId` needs the same treatment — it's an ARM resource id, likely not secret, but worth confirming).
- Make sure whatever reads the key back (the `/keys/me` endpoint per spec §4.5) decrypts correctly and never logs the plaintext.

Referenced: foundrygate-spec.md §11 (Security Considerations). Follow-up endorsed in the review on PR #92: https://github.com/kolatts/foundry-gate/pull/92#issuecomment-5498081091

### LINKED ISSUE #116: reference/api.md: DELETE /keys/{userId} semantics conflict with spec §5.3 and plan 09
Noticed while aligning the KeyLifecycle diagram (#89):

- `docs-site/src/content/docs/reference/api.md` documents `DELETE /keys/{userId}` as **"Revoke key (user stays active)"**.
- `foundrygate-spec.md` §5.3 and `plans/09-apim-keys.md` say the same call deletes the APIM subscription **and** sets `User.IsActive = false` plus `QuotaAllocation.IsHardStopped = true` (i.e. it is the deactivation path), and the issue #7 direction comment reserves subscription *suspension* for deactivation/offboarding.

One of these is wrong. Decide whether an admin can revoke a key while leaving the user active (then spec/plan need a separate deactivation endpoint or a flag), or whether revocation implies deactivation (then fix the API reference row). The new `KeyLifecycle` diagram currently follows the spec/plan reading (Deleted = offboarding / Entra departure / admin revoke).

### LINKED ISSUE #117: Key rotation: regeneratePrimaryKey leaves the APIM secondary key valid — regenerate both or never expose it
Raised in the #113 review.

`POST /keys/me/rotate` (plan 09, spec §5.2) calls APIM `regeneratePrimaryKey`. That invalidates the primary key the developer holds, but every APIM subscription also has a **secondary key** that stays valid and is never rotated. FoundryGate never shows the secondary to anyone, so from the developer's point of view rotation is complete — but it is a live, unrotated credential per subscription that an APIM Contributor could read.

Decide in the keys wave (#9 / #65):
- rotate both (`regeneratePrimaryKey` + `regenerateSecondaryKey` in the same operation), or
- treat the secondary as never-issued and regenerate it on provision + every rotation so it is always unknown to everyone, or
- document it as an accepted, RBAC-gated risk.

Whatever is chosen, `reference/api.md` and the overview's key-lifecycle text should say which keys a rotation touches.

kolatts:
## Summary

Wires FoundryGate to the APIM management plane for developer subscription keys and encrypts the stored key at rest, per plans/09 and the lifecycle contract in plans/21.

- **`Services/Security`** — `IKeyProtector` with two implementations selected by `KeyProtection:Provider`: `KeyVaultKeyProtector` (RSA-OAEP-256 wrap of the key bytes under `Gateway:KeyEncryptionKeyUri` via `CryptographyClient`, stored as the versioned envelope `kv1:{versionedKeyId}:{base64}`) and `DataProtectionKeyProtector` (`dp1:{token}`, **local only**). `KeyProtectorFactory` fail-fasts at host start through a small `AddResolveOnStartup<T>()` helper: `KeyVault` without a key URI → `ConfigurationValidationException`; `DataProtection` in `qa`/`prod` → refused.
- **`Services/Keys`** — `IApimManagementClient` (thin seam over `Azure.ResourceManager.ApiManagement`, 7 operations; `ArmApimManagementClient` + `UnconfiguredApimManagementClient` for local dev without APIM, itself refused outside `local`) and `IApimKeyService`/`ApimKeyService`: provision (under the tier product, orphan reuse with both keys regenerated), rotate (**both** keys, #117), key-only revoke (#116), move-to-product (tier change, for #118), masked read from a stored hint, audited reveal. Every mutation adds its audit row through `IAuditService` and saves once.
- **`Controllers/KeysController`** — `GET /keys/me`, `POST /keys/me/reveal`, `POST /keys/me/rotate`; admin `POST /keys/{userId}/provision?tier=`, `POST /keys/{userId}/rotate`, `DELETE /keys/{userId}` (204, idempotent, user stays active).
- **Data** — `User.ApimSubscriptionKey` widened to `nvarchar(1000)`; new `ApimSubscriptionKeyHint nvarchar(4)` and `ApimKeyIssuedDate datetimeoffset null`; `Users.sql` mirrored (SchemaParity green). `Domain/Keys/ApimSubscriptionNames` (`foundrygate-{UserId}` ↔ `UserId`) shared with the future Functions reconciliation job (#84). `AuditActions.KeyTierChanged` added.
- **Config** — `GatewayOptions` (`SubscriptionId`, `ResourceGroup`, `ApimName`, `KeyEncryptionKeyUri`; all-or-none validation) and `KeyProtectionOptions` on `AppSettings`, bound from the `Gateway` / `KeyProtection` sections — names match the `Gateway__*` env vars `control-plane.bicep` sets (#108, PR #111). `appsettings.json` defaults to `KeyVault`; `appsettings.local.json` opts into `DataProtection`.
- **Docs** — `reference/api.md` Keys table rewritten to the implemented semantics; `reference/configuration.mdx` env vars updated to `Gateway__*`/`KeyProtection__Provider`; `architecture/overview.mdx` Rotate paragraph (prose only); spec §4.5/§5.1–5.3; plans/09 (verification ticked) and plans/21 (no suspend surface, `RevokeAsync` step).

Closes #36
Closes #37
Closes #95
Closes #116
Closes #117

## Design decisions (please scrutinize)

1. **Key-only revocation, no suspend anywhere** (#116 ruling): `DELETE /keys/{userId}` deletes the subscription and clears the four key fields; `IsActive` is untouched. `SuspendAsync`/`ReenableAsync` and a `SetStateAsync` on the management client were deliberately **not** built — nothing in plan 21 calls them any more and a dead suspend path invites the wrong mental model. `RevokeAsync(User)` is the building block `POST /users/{id}/deactivate` (#65) calls first; it is a no-op (`false`) for a user with no key so the Entra-departure path can call it unconditionally.
2. **Rotation regenerates both APIM keys** (#117): the secondary is never issued, stored, or returned, and rotating it alongside the primary means no credential on the subscription outlives a rotation. Orphan reuse regenerates both too, so a reused subscription never keeps keys FoundryGate did not mint.
3. **Masked view from a stored hint, not a decrypt.** `ApimSubscriptionKeyHint` (last 4) makes `GET /keys/me` — and the `/users/me` profile that embeds it — a plain row read instead of a Key Vault unwrap per page load (Key Vault RSA-3072 ops are rate-limited per vault). Four characters is exactly what the masked display discloses by design. `GetMasked(User)` is therefore synchronous.
4. **`ApimKeyIssuedDate`** added so `ApiKeyRevealResponse.IssuedDate` on reveal reports when the key was minted/rotated (plan 09's "updates RotatedAt"), not when it was revealed.
5. **Envelope records the Key Vault key *version*.** Wrap targets the versionless URI (Key Vault picks current), the result's `KeyId` is stored; unwrap targets the stored version. A Key Vault key rotation therefore needs no re-encryption sweep, and rows still on an old version are findable with `LIKE 'kv1:%/{version}:%'`. Column budget checked with a real RSA-4096 wrap + long vault name in tests (< 1000).
6. **Direct RSA wrap of the key bytes** (no DEK indirection): APIM keys are 32 chars; RSA-3072 OAEP-256 wraps up to 318 bytes.
7. **`ApimSubscriptionId` stays plaintext** (#95 asked to confirm): it is an ARM resource id — an address, not a credential.
8. **Options live under `Gateway`, not a separate `Apim` section**, to match what infra already sets (`Gateway__SubscriptionId/ResourceGroup/ApimName/KeyEncryptionKeyUri`). #108's remaining fields (`ApimGatewayUrl`, `LogAnalyticsWorkspaceId`, `FoundryAccountNames`) are for the Foundry/reconciliation waves to add to the same `GatewayOptions` class. PR #111 stays the source of truth for the names.
9. **Fail-fast via eager resolution.** Both singletons are factory-registered (they need `AppSettings` + the `AppEnvironment.Types` singleton, which `AddFoundryGateApiServices()` cannot see); `AddResolveOnStartup<T>()` resolves them in an `IHostedService.StartAsync` so a misconfiguration aborts startup rather than surfacing as a 500 on the first key call — the same mechanism `OptionsBuilder.ValidateOnStart()` uses.
10. **Actor resolved before any APIM side effect.** An admin without a `User` row hitting `POST /keys/{id}/provision` gets the 403 *before* a subscription is created (otherwise the refused request would leave an orphan). Covered by `Admin_without_a_User_row_returns_403_on_provision_…`.
11. **Subscription name is `foundrygate-{UserId}`** per plan 21/#66/the orchestrator brief. `User.UserUnique`'s doc comment suggests the Guid for external naming; I followed the plan (an int identity is not secret and the Functions job needs the inverse mapping to be trivial) — flagging in case the reviewer prefers the Guid.
12. **Not-found contract of the management client**: `Get` → null, `Delete` → false, everything else → `ApimSubscriptionNotFoundException`, which the key service maps to a **409** with "revoke and re-provision" guidance (a subscription deleted in the portal is a state conflict, not a missing resource).
13. **`ProvisionForUserAsync` refuses a deactivated user (409)** — provisioning alone would leave an inactive user holding a working key; `POST /users/{id}/activate` (plan 21 Trigger C) is the path. The building block `ProvisionAsync(User, …)` does not check `IsActive` because the re-activation pipeline provisions before flipping the flag.

## Public interface shapes

```csharp
// Services/Security
public interface IKeyProtector
{
    Task<string> ProtectAsync(string plaintext, CancellationToken cancellationToken);
    Task<string> UnprotectAsync(string ciphertext, CancellationToken cancellationToken);
}
public sealed record KeyEnvelope(string Scheme, string? KeyId, string Payload); // "kv1" | "dp1"
public static IKeyProtector KeyProtectorFactory.Create(KeyProtectionOptions, GatewayOptions, AppEnvironment.Types, TokenCredential, IDataProtectionProvider);

// Services/Keys
public interface IApimManagementClient
{
    Task<ApimSubscriptionWithKeys> CreateOrUpdateSubscriptionAsync(string subscriptionName, string displayName, string productId, CancellationToken ct);
    Task<ApimSubscription?> GetSubscriptionAsync(string subscriptionName, CancellationToken ct);           // null = absent (orphan probe)
    Task<ApimSubscriptionKeys> ListSecretsAsync(string subscriptionName, CancellationToken ct);
    Task RegeneratePrimaryKeyAsync(string subscriptionName, CancellationToken ct);
    Task RegenerateSecondaryKeyAsync(string subscriptionName, CancellationToken ct);
    Task UpdateScopeAsync(string subscriptionName, string productId, CancellationToken ct);
    Task<bool> DeleteSubscriptionAsync(string subscriptionName, CancellationToken ct);                   // false = already gone
}
public sealed record ApimSubscription(string Name, string ResourceId, string DisplayName, string Scope, string? ProductId, string State);
public sealed record ApimSubscriptionKeys(string PrimaryKey, string SecondaryKey);
public sealed record ApimSubscriptionWithKeys(ApimSubscription Subscription, ApimSubscriptionKeys Keys);
public sealed class ApimSubscriptionNotFoundException(string subscriptionName) : Exception;

public interface IApimKeyService
{
    // building blocks (plan 21 orchestrator): tracked User in, APIM call, mutate, audit, SaveChanges
    Task<ApiKeyRevealResponse> ProvisionAsync(User user, string tierProductId, CancellationToken ct);   // 409 if key exists, 400 bad tier, orphan reuse
    Task<ApiKeyRevealResponse> RotateAsync(User user, CancellationToken ct);                            // both keys; 404 no key; 409 subscription vanished
    Task<bool> RevokeAsync(User user, CancellationToken ct);                                            // key-only; false = nothing to revoke
    Task MoveToProductAsync(User user, string tierProductId, CancellationToken ct);                     // tier change (#118)
    ApiKeyResponse GetMasked(User user);
    Task<ApiKeyRevealResponse> RevealAsync(User user, CancellationToken ct);
    // endpoint entry points
    Task<ApiKeyResponse> GetMineAsync(CancellationToken ct);
    Task<ApiKeyRevealResponse> RevealMineAsync(CancellationToken ct);
    Task<ApiKeyRevealResponse> RotateMineAsync(CancellationToken ct);
    Task<ApiKeyRevealResponse> ProvisionForUserAsync(int userId, string tierProductId, CancellationToken ct);
    Task<ApiKeyRevealResponse> RotateForUserAsync(int userId, CancellationToken ct);
    Task<bool> RevokeForUserAsync(int userId, CancellationToken ct);
}

// Configuration (AppSettings)
public GatewayOptions Gateway { get; set; }            // SubscriptionId, ResourceGroup, ApimName, KeyEncryptionKeyUri; IsApimConfigured
public KeyProtectionOptions KeyProtection { get; set; } // Provider: KeyVault (default) | DataProtection

// Domain
public static class ApimSubscriptionNames { const string Prefix = "foundrygate-"; string ForUser(int userId); bool TryGetUserId(string? name, out int userId); }
```

## Verification

- `dotnet build FoundryGate.sln -c Release`: succeeded, **zero warnings**.
- `dotnet test src/FoundryGate.Tests.Predeployment -c Release --no-build`: **229 passed, 0 failed** (126 on main + 103 new).
  - Protectors: round trip under both providers, `kv1`/`dp1` envelope grammar, version pinning across a simulated Key Vault rotation (real RSA-3072 fake), one client per key id, RSA-4096 length budget, foreign-envelope and tamper refusal, factory rules (`DataProtection` refused in `qa`/`prod`; `KeyVault` needs a URI).
  - `ApimKeyServiceTests` (real SQLite + real audit path + fake APIM): provision/conflict/bad tier/orphan reuse/ARM failure leaves row untouched; rotate both keys; revoke clears fields + keeps `IsActive` + idempotent + tolerates already-gone; masked ≤ 4 chars (no 5-char window survives); reveal audited without APIM; move-to-product before/after; `/me` operations and deactivated-caller refusal; **plaintext never reaches the logger (formatted messages *and* structured state) or any audit `Details`**.
  - `KeysEndpointTests` (`ApiTestFactory` + fake APIM + ephemeral Data Protection): 401 anonymous; 403 no `User` row (`/me` and admin-without-row); 403 non-admin on admin routes; 200 masked; 404 reveal/rotate with no key; 200 provision (default and `?tier=power`), 400 bad tier, 409 second provision / inactive user, 404 unknown user; reveal = minted key; rotate replaces primary **and** secondary; 204 revoke keeps user active, idempotent, re-provisionable.
  - Existing suites untouched and green (SchemaParity picks up the three column changes; DomainArchitecture allowlist unchanged).
- `dotnet format FoundryGate.sln --verify-no-changes`: clean.
- `cd docs-site && npm ci && npm run build`: Complete.
- `plans/09-apim-keys.md` Verification: all code items ticked; the live item points at #132.
- No live Azure was touched. Nothing under `.github/` or `infra/`.

## Follow-ups filed

- #132 — live-validate the key lifecycle + Key Vault wrapping against real APIM/Key Vault (exact manual checklist).

## Shared-file footprint (for the parallel quota/Entra/Foundry PRs)

`Services/ApiServiceCollectionExtensions.cs` (+2 lines), `Configuration/AppSettings.cs` (+2 properties, +3 classes appended), `appsettings*.json` (own sections), `Directory.Packages.props` + `FoundryGate.Api.csproj` (2 packages), `Domain/Constants/AuditActions.cs` (+1 constant), `Tests/Api/ApiTestFactory.cs` (config key + fake APIM/ephemeral Data Protection registration + `Apim` property). No files owned by the quota, Entra or Foundry waves were created or edited.

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -36,6 +36,10 @@
     <PackageVersion Include="Azure.Core" Version="1.62.0" />
     <PackageVersion Include="Azure.Monitor.OpenTelemetry.AspNetCore" Version="1.6.0" />
     <PackageVersion Include="Azure.Security.KeyVault.Secrets" Version="4.11.0" />
+
+    <!-- Api: APIM subscription lifecycle + Key Vault key wrapping of stored APIM keys (#36, #37, #95) -->
+    <PackageVersion Include="Azure.ResourceManager.ApiManagement" Version="1.3.1" />
+    <PackageVersion Include="Azure.Security.KeyVault.Keys" Version="4.10.0" />
     <PackageVersion Include="Microsoft.AspNetCore.OpenApi" Version="10.0.11" />
     <PackageVersion Include="Microsoft.Extensions.Diagnostics.HealthChecks.EntityFrameworkCore" Version="10.0.11" />
     <PackageVersion Include="Microsoft.Identity.Web" Version="4.14.2" />
--- a/docs-site/src/components/KeyLifecycle.astro
+++ b/docs-site/src/components/KeyLifecycle.astro
@@ -1,7 +1,7 @@
 ---
 // APIM subscription (developer key) state machine — gateway-centric model, lifecycle per
 // plans/21-provision-deprovision.md (#65/#66): a key IS an APIM subscription created under
-// a tier product. Rotation (regeneratePrimaryKey) and quota changes (re-scope to another
+// a tier product. Rotation (regeneratePrimaryKey + regenerateSecondaryKey — both keys) and quota changes (re-scope to another
 // tier product) keep it Active; 403/429 are gateway responses and change nothing. Admin
 // deactivation, Entra departure and admin key revocation all DELETE the subscription;
 // re-activation re-runs provisioning (reusing an orphan subscription if one exists).
@@ -12,7 +12,7 @@
   <p id="fgd-kl-title" class="fgd-sr-only">APIM key lifecycle state machine</p>
   <p id="fgd-kl-desc" class="fgd-sr-only">
     From No key, first login or admin provision creates an APIM subscription under the developer's tier product and
-    the key becomes Active. While Active: rotating the key calls regeneratePrimaryKey and stays Active; changing the
+    the key becomes Active. While Active: rotating the key regenerates both the primary and the secondary key and stays Active; changing the
     quota re-scopes the subscription to the new tier's product and stays Active (the gateway's monthly counter
     restarts for that subscription; live validation tracked in #88); a 403 for an exhausted monthly quota or a 429
     with Retry-After for the per-minute cap are gateway responses that change nothing. Admin deactivation, Entra
@@ -69,7 +69,7 @@
       <!-- Active ⟲ rotate (right loop) -->
       <path d="M 430,148 C 482,140 482,182 430,176" class="fgd-s-edge fgd-s-edge--green" marker-end="url(#fgd-kl-head-green)" />
       <text x="488" y="156" class="fgd-s-label fgd-s-label--green">rotate key</text>
-      <text x="488" y="170" class="fgd-s-label fgd-s-mono">regeneratePrimaryKey</text>
+      <text x="488" y="170" class="fgd-s-label fgd-s-mono">regenerate both keys</text>
 
       <!-- Active ⟲ change quota (left loop) -->
       <path d="M 220,148 C 168,140 168,182 220,176" class="fgd-s-edge fgd-s-edge--green" marker-end="url(#fgd-kl-head-green)" />
--- a/docs-site/src/components/SystemDiagram.astro
+++ b/docs-site/src/components/SystemDiagram.astro
@@ -18,7 +18,7 @@ const ariaLabel =
   'Azure Static Web Apps calls FoundryGate.Api on Azure Container Apps, which stores state in Azure SQL ' +
   'through EF Core; FoundryGate.Functions run the monthly reset and usage reconciliation, reading ' +
   'ApiManagementGatewayLlmLog from Log Analytics; Key Vault holds secrets. The API manages APIM ' +
-  'subscriptions through the Management API: create under a tier product, regenerate key, move tier, delete. ' +
+  'subscriptions through the Management API: create under a tier product, regenerate both keys, move tier, delete. ' +
   'Legend: solid azure arrows are the real-time enforcement path on the request; dashed grey arrows are telemetry ' +
   'and reconciliation, out of band; dashed amber arrows are control-plane calls for lifecycle and tiers.';
 ---
@@ -153,7 +153,7 @@ const ariaLabel =
     <div class="fgd-arrow fgd-arrow--app fgd-arrow--up">
       <span class="fgd-arrow__label">
         <strong>APIM Management API</strong> — create subscription under a tier product ·
-        <code>regeneratePrimaryKey</code> · move tier · delete
+        regenerate both keys · move tier · delete
       </span>
       <span class="fgd-arrow__line"></span>
       <span class="fgd-arrow__head"></span>
--- a/docs-site/src/content/docs/architecture/overview.mdx
+++ b/docs-site/src/content/docs/architecture/overview.mdx
@@ -200,9 +200,13 @@ people-lifecycle events change its state — there is no suspended state:
 <KeyLifecycle />
 
 **Rotate** — `POST /keys/me/rotate` (or an admin's `POST /keys/{userId}/rotate`) calls
-`regeneratePrimaryKey` on the subscription. The primary key the developer holds is
-replaced instantly, the new one is shown once, and the subscription stays `active` with
-its counters intact.
+`regeneratePrimaryKey` *and* `regenerateSecondaryKey` on the subscription: the primary key
+the developer holds is replaced instantly and shown once, and the secondary — which
+FoundryGate never issues or stores — is replaced too, so no credential outlives a rotation.
+The subscription stays `active` with its counters intact. Stored keys are wrapped with a
+Key Vault RSA key before they reach SQL; `GET /keys/me` shows only the last four characters,
+and `POST /keys/me/reveal` decrypts on demand (audited) rather than caching anything in the
+browser.
 
 **Change quota** — the control plane re-scopes the subscription to the new tier's
 product. The new product's `llm-token-limit` applies from the next request, and because
--- a/docs-site/src/content/docs/reference/api.md
+++ b/docs-site/src/content/docs/reference/api.md
@@ -53,11 +53,14 @@ Base path: `/api/v1`. All endpoints require a valid Entra ID bearer token. Admin
 
 | Method | Path | Auth | Description |
 |---|---|---|---|
-| `GET` | `/keys/me` | Any | Own key info (masked, last 4 visible) |
-| `POST` | `/keys/me/rotate` | Any | Rotate own key — returns new key value once |
-| `POST` | `/keys/{userId}/rotate` | Admin | Rotate any user's key |
-| `POST` | `/keys/{userId}/provision` | Admin | Provision a new key for a user with no active key |
-| `DELETE` | `/keys/{userId}` | Admin | Revoke key (user stays active) |
+| `GET` | `/keys/me` | Any | Own key info (masked, last 4 visible; `isProvisioned: false` when none). Served from a stored hint — no decryption |
+| `POST` | `/keys/me/reveal` | Any | Decrypt and return own full key once. Audited (`key.revealed`), never cached. `404` when no key |
+| `POST` | `/keys/me/rotate` | Any | Rotate own key — regenerates **both** APIM keys (primary and never-issued secondary), returns the new primary once. `404` no key; `409` if the APIM subscription vanished |
+| `POST` | `/keys/{userId}/rotate` | Admin | Rotate any user's key (same semantics) |
+| `POST` | `/keys/{userId}/provision` | Admin | Provision a key for a user with none, under `?tier=standard\|power\|unlimited` (default `standard`). Returns plaintext once. `409` key exists or user deactivated; `400` unknown tier; reuses an orphaned APIM subscription with fresh keys |
+| `DELETE` | `/keys/{userId}` | Admin | Revoke key only: APIM subscription deleted, stored key cleared, `key.revoked` audited. **User stays active** and can be re-provisioned; `204` even when no key existed. Deactivation is `POST /users/{id}/deactivate` |
+
+Callers of every `/keys/me` route must already have a FoundryGate user row (`GET /users/me` provisions one) — otherwise `403`. The plaintext key is stored encrypted (Key Vault RSA key wrapping; see [Configuration](/reference/configuration/)) and appears in exactly one response per mint or reveal.
 
 ## Foundry
 
--- a/docs-site/src/content/docs/reference/api.md
+++ b/docs-site/src/content/docs/reference/api.md
@@ -60,7 +60,7 @@ Base path: `/api/v1`. All endpoints require a valid Entra ID bearer token. Admin
 | `POST` | `/keys/{userId}/provision` | Admin | Provision a key for a user with none, under `?tier=standard\|power\|unlimited` (default `standard`). Returns plaintext once. `409` key exists or user deactivated; `400` unknown tier; reuses an orphaned APIM subscription with fresh keys |
 | `DELETE` | `/keys/{userId}` | Admin | Revoke key only: APIM subscription deleted, stored key cleared, `key.revoked` audited. **User stays active** and can be re-provisioned; `204` even when no key existed. Deactivation is `POST /users/{id}/deactivate` |
 
-Callers of every `/keys/me` route must already have a FoundryGate user row (`GET /users/me` provisions one) — otherwise `403`. The plaintext key is stored encrypted (Key Vault RSA key wrapping; see [Configuration](/reference/configuration/)) and appears in exactly one response per mint or reveal.
+Callers of every `/keys/me` route must already have a FoundryGate user row (`GET /users/me` provisions one) — otherwise `403`. The plaintext key is stored encrypted (Key Vault RSA key wrapping; see [Configuration](/reference/configuration/)) and appears in exactly one response per mint or reveal. Reveal is not yet rate-limited (tracked in #136). Provisioning is race-safe: two concurrent `provision` calls for one user cannot both mint — the second gets `409`.
 
 ## Foundry
 
--- a/docs-site/src/content/docs/reference/configuration.mdx
+++ b/docs-site/src/content/docs/reference/configuration.mdx
@@ -12,9 +12,9 @@ These are stored in the `SystemConfiguration` database table and editable by adm
 | Key | Default | Description |
 |---|---|---|
 | `DefaultMonthlyTokenQuota` | `"1000000"` | Per-user monthly token budget when no user or group override applies. |
-| `ApimResourceId` | `""` | ARM resource ID of the APIM instance (`/subscriptions/.../Microsoft.ApiManagement/service/...`). Used by `IApimKeyService` to target Management plane calls. |
+| `ApimResourceId` | `""` | ARM resource ID of the APIM instance (`/subscriptions/.../Microsoft.ApiManagement/service/...`). Informational — the key service addresses APIM through the `Gateway__*` environment variables below, not this value. |
 | `ApimGatewayUrl` | `""` | APIM gateway base URL shown to developers on `/me` for CLI configuration (e.g., `https://my-apim.azure-api.net`). |
-| `ApimProductId` | `"foundrygate"` | APIM product name that all Foundry API routes belong to. New APIM subscriptions are scoped to this product. |
+| `ApimProductId` | `"foundrygate"` | Legacy single-product name. Superseded: a developer's subscription is created under their **quota-tier product** (`standard`, `power`, `unlimited`), not this one. Kept for the seeded reference data. |
 | `FoundryResourceId` | `""` | ARM resource ID of the Azure AI Foundry account. Used by `IFoundryDeploymentService` for model deployment management. |
 | `EntraTenantId` | `""` | Azure AD tenant ID used by the Microsoft Graph SDK for bulk user sync. |
 | `EntraGroupSyncEnabled` | `"false"` | Set to `"true"` to enable automatic Entra group member sync. When `false`, group sync must be triggered manually from the admin UI. |
@@ -36,12 +36,15 @@ These are set as Container App environment variables (sourced from Key Vault ref
 | `AzureAd__TenantId` | Config / env var | Entra tenant ID for bearer token validation |
 | `AzureAd__ClientId` | Config / env var | Entra App Registration client ID |
 | `AzureAd__Audience` | Config / env var | Token audience (usually `api://{clientId}`) |
-| `Azure__ApimResourceGroup` | Config / env var | Resource group containing the APIM instance |
-| `Azure__ApimServiceName` | Config / env var | APIM service name (short name, not full resource ID) |
+| `Azure__KeyVaultUrl` | Config / env var | Key Vault URI for resolving `@KeyVault(SecretName)` references in appsettings. Optional locally |
+| `Gateway__SubscriptionId` | Set by infra | Azure subscription id holding the APIM instance. With the next two, addresses the APIM management plane for subscription-key operations; all three or none |
+| `Gateway__ResourceGroup` | Set by infra | Resource group of the APIM instance |
+| `Gateway__ApimName` | Set by infra | APIM service name (short name, not the full resource ID) |
+| `Gateway__KeyEncryptionKeyUri` | Set by infra | Versionless URI of the Key Vault RSA key (`fg-apim-key-encryption`) that wraps stored APIM subscription keys (RSA-OAEP-256). Required when `KeyProtection__Provider` is `KeyVault` |
+| `KeyProtection__Provider` | Config / env var | `KeyVault` (default; cloud) or `DataProtection` (machine-local key ring — accepted in the `local` environment only; the API refuses to start with it anywhere else) |
 | `Azure__FoundrySubscriptionId` | Config / env var | Azure subscription ID containing Foundry |
 | `Azure__FoundryResourceGroup` | Config / env var | Resource group containing the Foundry account |
 | `Azure__FoundryAccountName` | Config / env var | Foundry account name (short name) |
-| `Azure__KeyVaultUri` | Config / env var | Key Vault URI for key wrapping (APIM key encryption) |
 | `Graph__ClientId` | Config / env var | App registration client ID for Graph SDK (may differ from AzureAd__ClientId) |
 | `Graph__ClientSecret` | Key Vault secret | Client secret for Graph app-only credentials |
 
--- a/foundrygate-spec.md
+++ b/foundrygate-spec.md
@@ -227,10 +227,11 @@ PUT    /requests/{id}/reject         Admin only
 
 ```
 GET    /keys/me                      Any: get own APIM key (masked except last 4)
-POST   /keys/me/rotate               Any: rotate own APIM subscription key
+POST   /keys/me/reveal               Any: decrypt and return own full key once (audited)
+POST   /keys/me/rotate               Any: rotate own APIM subscription key (both APIM keys regenerated)
 POST   /keys/{userId}/rotate         Admin: rotate key for any user
-POST   /keys/{userId}/provision      Admin: provision a new APIM key for a user
-DELETE /keys/{userId}                Admin: revoke APIM key (hard deactivate)
+POST   /keys/{userId}/provision      Admin: provision a new APIM key for a user (?tier=standard|power|unlimited)
+DELETE /keys/{userId}                Admin: revoke APIM key only — user stays active (deactivation is PUT /users/{id}/deactivate)
 ```
 
 ### 4.6 Admin / Configuration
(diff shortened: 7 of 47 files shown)


kolatts:
**REQUEST CHANGES**

This is careful work — the seam design, the fail-fast factory, the actor-before-side-effects ordering, and the "plaintext never reaches the logger" test are all the right instincts, and 229 green on a fresh build here. One correctness bug in the Key Vault version-pinning story blocks it, and `AppSettings.Gateway` has to be unified with #127/#131 before any of the three can land.

Verified locally: `dotnet build` zero warnings, 229/229 predeployment tests, `dotnet format --verify-no-changes` clean. Note the branch predates the #111 merge (`git log HEAD..origin/main` shows the control-plane infra landing), so a rebase is due anyway — good news, since the `Gateway__*` env vars and the `fg-apim-key-encryption` RSA-3072 key this PR binds to now exist on `main`. APIM `PATCH /subscriptions/{sid}` confirmed against the REST reference: `primaryKey`/`secondaryKey` are optional and untouched when omitted, so `MoveToProductAsync` really is key-preserving. Dropping Suspend/Reenable entirely is consistent with the #116 ruling and with infra (nothing reads subscription `state` except the `active` on create). Key-only 204-idempotent revoke matches the ruling. `foundrygate-{UserId}` is the right call — plan 21/#66 name it, an identity int isn't secret (it's already in every admin URL), #84 needs the inverse mapping to be trivial, and an orphan adopted after an identity reseed gets both keys regenerated so nothing leaks — but see the doc-comment nit.

---

**Issues**

- **Major** `src/FoundryGate.Api/Services/Security/KeyVaultKeyProtector.cs` lines 39–42, 61–63 — "Wrap targets the versionless URI so Key Vault picks the current version" is only true once per process: `CryptographyClient.InitializeAsync` fetches the key on first use, caches a local RSA provider for the client's lifetime (verified in the SDK source — `if (_provider != null) return;`), and your `_clients` dictionary caches that client forever, so after a Key Vault key rotation every new wrap still goes to the *old* version (local wrap checks only `NotBefore`/`ExpiresOn`, not `Enabled`). Consequence: the standard "rotate, then disable the old version" hygiene step makes every key provisioned or rotated since the rotation unrevealable — and `POST /keys/me/rotate` cannot fix it until an app restart. The #132 checklist item "a subsequent rotate writes an envelope carrying the NEW version id" will fail live. Consider resolving the current version per wrap (`KeyClient.GetKeyAsync(name)` → versioned client, one cheap GET) or a short TTL on the versionless entry, and have the fake model a stale cached client so the test catches it.
- **Minor** `src/FoundryGate.Api/Configuration/AppSettings.cs` line 57 — three open PRs define `AppSettings.Gateway` three different ways (#127 `GatewayTierOptions { Tiers }`, #131 `GatewayOptions { SubscriptionId, ResourceGroup, FoundryAccountNames }`, this one), and your all-three-or-none APIM rule would reject #131's test config (SubscriptionId + ResourceGroup without ApimName) at startup. Proposed unified shape below; whoever rebases last needs it.
- **Minor** `src/FoundryGate.Api/Services/Security/KeyVaultKeyProtector.cs` line 50–56 — `UnprotectAsync` builds a `CryptographyClient` for whatever absolute URI the DB row carries; pin the envelope's host (and key name) to `keyEncryptionKeyUri` so a tampered row can't point the API identity at a different vault it happens to have access to. Two lines; the SDK's challenge-resource check covers token theft, not cross-vault.
- **Minor** `src/FoundryGate.Api/Controllers/KeysController.cs` line 28 — `POST /keys/me/reveal` is audited but unthrottled; a leaked bearer token can pull the plaintext indefinitely and the only trace is `key.revealed` rows. Not for this PR — file an issue for a per-user rate limit (ASP.NET rate limiter partitioned on oid) and reference it.
- **Minor** `docs-site/src/components/KeyLifecycle.astro` line 72 (and line 4/15, `SystemDiagram.astro` line 156) — the diagram label still says `regeneratePrimaryKey` while the PR made rotation regenerate both keys; CONVENTIONS "Architecture diagrams" says every wave updates the components, and #117 asked the lifecycle text to say which keys a rotation touches.
- **Minor** `src/FoundryGate.Api/Services/Keys/ApimKeyService.cs` lines 58, 157, 288 — every building block resolves the actor via `ICurrentUserAccessor`/`IAuditService`, so `RevokeAsync` cannot be "called unconditionally" from plan 21 Trigger B (Entra departure from a sync/Functions job with no HTTP caller). Fine for this PR, but #65 will need an `IAuditWriter.AddSystem`-based path — worth a line in the #65 issue now.
- **Minor** `src/FoundryGate.Api/Services/Keys/ApimKeyService.cs` lines 52–85 — two concurrent `POST /keys/{id}/provision` both pass `HasKey` and both see `GetSubscriptionAsync == null`, so both PUT; the REST contract generates fresh keys on a PUT that omits them, so the first caller's returned key can die and last-DB-write wins. Rare and admin-only; a conditional `ExecuteUpdate` on `ApimSubscriptionId = ''` (or a rowversion) closes it.
- **Nit** `src/FoundryGate.Data/Entities/User.cs` line 16–20 — `UserUnique`'s comment still says it exists to name the APIM subscription; the PR (rightly) used `UserId`. Update the comment so the entity doesn't argue with `ApimSubscriptionNames`.
- **Nit** `src/FoundryGate.Api/Services/Keys/ApimKeyService.cs` line 247 — `user.ApimKeyIssuedDate ?? timeProvider.GetUtcNow()` reports "issued just now" for a key that wasn't; `HasKey` ⇒ `ApimKeyIssuedDate != null` is an invariant you write, so throw or assert rather than fabricate.
- **Nit** `src/FoundryGate.Api/Services/Keys/ApimKeyService.cs` lines 122–124 — if `ListSecretsAsync` or the save fails after both regenerates, the row keeps a dead ciphertext with no signal; a warning log naming the remedy ("rotate again") would save someone an hour.
- **Nit** `src/FoundryGate.Api/Services/Keys/ApimKeyService.cs` lines 68–79 — the orphan branch re-scopes and re-keys but never asserts `State == active` (the create path sets it); a hand-made or suspended orphan is adopted as-is and its fresh key 401s at the gateway. One check on `existing.State`.
- **Nit** `src/FoundryGate.Api/Configuration/AppSettings.cs` line 137 — `[Required]` on a non-nullable enum is a no-op.

---

**Proposed unified `GatewayOptions`** (one class, bound from `Gateway`, replaces #127's root `GatewayTierOptions` and both `GatewayOptions` variants; #127's per-tier record stays as the `Tiers` item type):

```csharp
public class GatewayOptions : IValidatableObject
{
    // ARM scope shared by APIM and Foundry (Gateway__SubscriptionId / __ResourceGroup, #108)
    public string? SubscriptionId { get; set; }
    public string? ResourceGroup { get; set; }

    // APIM (this PR)
    public string? ApimName { get; set; }
    public string? ApimGatewayUrl { get; set; }          // #108, shown on /me
    public string? KeyEncryptionKeyUri { get; set; }     // #95

    // Foundry (#131)
    public List<string> FoundryAccountNames { get; set; } = [];

    // Reconciliation (#84)
    public string? LogAnalyticsWorkspaceId { get; set; }

    // Quota tiers (#127) — always required, shipped defaults in appsettings.json
    [Required] public List<GatewayTier> Tiers { get; set; } = [];

    public bool IsApimConfigured => Has(SubscriptionId) && Has(ResourceGroup) && Has(ApimName);
    public bool IsFoundryConfigured => Has(SubscriptionId) && Has(ResourceGroup) && FoundryAccountNames.Count > 0;

    public IEnumerable<ValidationResult> Validate(ValidationContext ctx)
    {
        // ApimName or FoundryAccountNames present ⇒ SubscriptionId + ResourceGroup required
        // (the pair is shared scope; neither feature owns it — drop this PR's "all three or none").
        // KeyEncryptionKeyUri, when present, absolute https.
        // FoundryAccountNames: no blanks (#131).
        // Tiers: non-empty, ids ⊆ GatewayTiers.All, unique, ≤1 unlimited (#127's rules, unchanged).
    }
}
```

`IsApimConfigured` (Keys registration) and `IsFoundryConfigured` (Foundry registration) each fail-fast outside `local` exactly as they do today; the test factory sets `Gateway:ApimName` alongside #131's Foundry keys. Suggested landing order: #127 first (it owns `Tiers` + validation tests), then this PR rebases and folds its four properties in, then #131 adds `FoundryAccountNames`.

Once the version-pinning fix and the rebase land, I'd be happy to re-review quickly — the rest of this is ready.

kolatts:
Thanks — the version-pinning catch was a real one. Two commits on top: `fix(keys): review pass …` (every item below) and a merge of `origin/main` now that #124 and #131 have landed, with the `GatewayOptions` fold. Point-by-point:

**Minor — unified `GatewayOptions` (three PRs, three shapes) + the all-three-or-none rule.** Done on the merge. My `Gateway` class is gone from `AppSettings.cs`; `Configuration/GatewayOptions.cs` (#131's) now carries `ApimName`, `KeyEncryptionKeyUri` and `IsApimConfigured => Sub && RG && ApimName` alongside `FoundryAccountNames`/`IsFoundryConfigured`, with one validation rule: **a feature member present (`ApimName` or `FoundryAccountNames`) ⇒ `SubscriptionId` + `ResourceGroup` required**; the shared pair on its own is valid (test `…_shared_scope_without_any_feature_member_is_valid_neither_feature_is_configured` pins #131's config shape). All-three-or-none is dropped. `KeyEncryptionKeyUri` keeps its https check. Room is left for #127's `Tiers` (remarks say so). `UnconfiguredApimManagementClient` now throws main's `FeatureNotConfiguredException` (→ 503), the same contract as `/foundry/*`. `ApiTestFactory` sets `Gateway:ApimName` next to the Foundry keys via `UseSetting`. `KeyProtection` stays its own small section (`Provider` only) — it is not gateway addressing.

**Major — `CryptographyClient` pins the first version it sees (`KeyVaultKeyProtector`).** Fixed, and you were right about the SDK: `InitializeAsync` fetches once and keeps a local provider for the client's lifetime, so a cached versionless client is a cached version. The protector now never builds a client on the versionless URI. Wrap resolves the key's **current version explicitly** through `KeyClient.GetKeyAsync(name)` on the vault, caches the versioned id for `CurrentVersionTtl` (5 min), and wraps with a `CryptographyClient` bound to that versioned id (one client per version, cached — versions are immutable). It also refuses to wrap if the current version is `Enabled == false`. Unwrap stays pinned to the envelope's version and never consults `KeyClient`. Tests (`KeyVaultKeyProtectorTests`): the fake `CryptographyClient` factory **throws if asked for a versionless client** (models the stale-client trap), a `FakeKeyClient` answers "current version is X", and `After_a_Key_Vault_key_rotation_new_wraps_use_the_new_version_and_old_envelopes_still_unwrap` flips X, shows the ≤ TTL lag, advances the clock, and asserts the new wrap carries v2 while v1 and v2 envelopes both unwrap with zero `KeyClient` calls. `Current_version_is_resolved_through_KeyClient_once_per_TTL_window` pins the cache. #132's checklist item now has a code path that can pass.

**Minor — `UnprotectAsync` trusts whatever URI the row carries.** Fixed: the envelope's key id must be `https`, on the configured vault host (case-insensitive), the configured key **name**, and versioned; anything else is `InvalidOperationException` before a client is even constructed. Six-case theory `Unprotect_refuses_envelopes_that_do_not_name_a_version_of_the_configured_key` (other vault, other key, versionless, http, non-URI, dp1) plus a host-casing acceptance test.

**Minor — reveal unthrottled.** Filed #136 (`backend`, v0.2): per-user limiter partitioned on `oid` for `/keys/me/reveal` (+ `/rotate`), 429 + `Retry-After`, anomaly signal, hermetic test via `ApiTestFactory`. Referenced from `KeysController.RevealMineAsync` and `reference/api.md`.

**Minor — diagram labels still say `regeneratePrimaryKey`.** Fixed in `KeyLifecycle.astro` (header comment, `aria` description, the SVG loop label → "regenerate both keys") and `SystemDiagram.astro` (alt text and the APIM card). Docs build green.

**Minor — building blocks always need an HTTP caller (plan 21 Trigger B).** Added `RevokeAsSystemAsync(User, reason, ct)`: same body as `RevokeAsync`, audit row via `IAuditWriter.AddSystem` (`ActorUserId = null`, `reason` in details), never touches `ICurrentUserAccessor`. Test builds the service with a null `HttpContext`, shows `RevokeAsync` 403s there and the system variant works, is idempotent, and rejects a blank reason. Noted on #65 (with the transaction-joining behaviour below) so the orchestrator wave knows what exists.

**Minor — concurrent provisions both PUT, first caller's key dies.** Fixed, and I went one step further than a post-hoc conditional update because that would still let the second PUT regenerate the first caller's keys. `ProvisionAsync` now **claims the row before calling APIM**: `ExecuteUpdateAsync` on `Users WHERE UserId = @id AND ApimSubscriptionId = ''` setting it to the deterministic ARM id (`IApimManagementClient.GetSubscriptionResourceId`, pure), inside a transaction the service owns — unless `dbContext.Database.CurrentTransaction` already exists, in which case it joins the orchestrator's (plan 21). Loser sees 0 rows → 409, no APIM call. APIM failure → the transaction disposes → claim rolled back (test asserts the DB column is `''` again and that a retry succeeds). On SQL Server the claimed row is X-locked until commit, so a true simultaneous provisioner blocks then sees 0 rows; the lock is one PK row for the duration of one ARM PUT, admin-only — acceptable. Why not `[Timestamp] RowVersion`: EF's SQLite provider doesn't maintain rowversion without a trigger workaround in the test harness, and optimistic concurrency would only detect the collision *after* the second PUT had already killed the first key; the claim prevents the second PUT. Tests: `…_Conflict_when_a_concurrent_request_already_claimed_the_row_and_never_calls_APIM`, `…_rolls_the_claim_back`, `…_joins_a_transaction_the_caller_already_opened_instead_of_starting_its_own`.

**Nit — `UserUnique` comment.** Rewritten: it's the `{Entity}Unique` non-sequential external id per CONVENTIONS, explicitly *not* the APIM subscription name, with the why.

**Nit — `RevealAsync` fabricates `IssuedDate`.** Now throws `InvalidOperationException` on `HasKey && ApimKeyIssuedDate == null` (an invariant this service writes); test `Reveal_refuses_a_row_that_has_a_key_but_no_issued_date_rather_than_inventing_one`.

**Nit — dead ciphertext after a failed rotation.** After both regenerates succeed, `ListSecrets` + store + audit + save run in a guarded block. On failure: previous key/hint/issued-date restored so the row is self-consistent, the unsaved `key.rotated` row is detached, an **Error** log names the remedy ("rotate again, or revoke and re-provision"), and a `key.rotation-failed` audit row (new `AuditActions.KeyRotationFailed`, details include the exception type and remedy) is written best-effort; the original exception is rethrown. Test `Rotate_failure_after_APIM_regenerated_…` uses the fake's `ThrowOnListSecrets` and then proves the remedy (a second rotate) yields a revealable key.

**Nit — orphan `State` never checked.** An orphan whose `State != active` is now deleted and created afresh rather than adopted (there's no suspend/resume surface by design, and a regenerated key on a suspended subscription still 401s). Warning logged; `Provision_replaces_an_orphan_that_is_not_active_instead_of_adopting_it`.

**Nit — `[Required]` on a non-nullable enum.** Removed, with a comment saying why the default is the safe one.

**Verification (post-merge head):** `dotnet build -c Release` zero warnings · `dotnet test` **360/360** (main's 244 + this PR's 116; +18 tests since the first push) · `dotnet format --verify-no-changes` clean · `docs-site` `npm run build` complete · CI green. Interface deltas since the first push: `IApimKeyService.RevokeAsSystemAsync(User, string reason, ct)`; `IApimManagementClient.GetSubscriptionResourceId(string)`; `KeyProtectorFactory.Create(…, TimeProvider)`; `KeyVaultKeyProtector(Uri, KeyClient, Func<Uri, CryptographyClient>, TimeProvider)`; `AuditActions.KeyRotationFailed`; `ApimKeyService` takes `IAuditWriter`; `GatewayOptions` gains `ApimName`, `KeyEncryptionKeyUri`, `IsApimConfigured`.

Ready for re-review.

Issue #142, opened 2026-09-01
Live-validate CLI ip setup / ip cleanup / db grant-identities against the dev Azure SQL server


kolatts:
> Parent: #220 — Live validation of the dev environment

Companion to #105 (control-plane live validation). The CLI half of the database deploy — `ip setup` / `ip cleanup` (#96) and `db grant-identities` (#106) — is unit-tested behind `IAzureSqlServerClient` / `ISqlBatchExecutor` fakes and the generated T-SQL was executed against docker SQL Server 2022 (role grants + idempotent re-run; the `FROM EXTERNAL PROVIDER` / `TYPE = E` branches cannot run outside Azure SQL). Nothing has touched a live Azure SQL server yet. Tick these on the first real `dev` run of `_deploy-database.yml` (or by hand with `az login` against `sql-foundrygate-dev-e7k2`).

## `ip setup`

- [ ] From a developer machine: `foundrygate ip setup --env dev` detects the public IPv4, resolves `sql-foundrygate-dev-e7k2` by listing `rg-foundrygate-dev` (no `--server`), prompts, and creates `fg-dev-{machine}-{user}` with start = end = that IP. Visible in the portal under the server's Networking blade.
- [ ] Run it again unchanged → "already allows … — nothing to do", **no ARM write** (check the activity log has no second `Microsoft.Sql/servers/firewallRules/write`).
- [ ] From the GitHub Actions runner (`GITHUB_ACTIONS=true`): the rule is `gha-{GITHUB_RUN_ID}-{yyyyMMddHHmm}` and the OIDC principal's `AzureCliCredential` (azure/login@v2 session) is what ARM sees — confirm a stray `AZURE_CLIENT_ID` in the job env does **not** make the CLI try Workload/ManagedIdentity (see `CliTokenCredential`).
- [ ] `AZURE_SUBSCRIPTION_ID` in the job env selects the subscription; without it the default-subscription path also resolves the RG (single-subscription principal).
- [ ] The OIDC principal has enough RBAC for the firewall write (SQL Server Contributor / Contributor on `rg-foundrygate-dev`) — it is separate from the SQL Entra admin group membership in #109.
- [ ] 60 s propagation wait + TCP 1433 probe pass on the first attempt (if not, note how long propagation actually took).
- [ ] `--env production` addressed `rg-foundrygate-prod` / `sql-foundrygate-prod-*` (naming alias check) — prod only once it exists.

## `ip cleanup`

[...]

Issue #105, opened 2026-09-01
Live-validate the control-plane Bicep deploy (SQL/Container Apps/SWA/Functions/Key Vault/RBAC)


kolatts:
> Parent: #220 — Live validation of the dev environment

Follow-up to #43/#44 (PR pending). The control-plane modules under `infra/modules/` compile, lint clean and pass `az deployment sub what-if` (59 creates, Succeeded), but several items in `plans/13-bicep-infra.md` Verification can only be proven by an actual deploy — which costs money and (for the gateway half) touches create-once Anthropic deployments, so it was deliberately NOT run by the implementing agent.

**Status 2026-09-05:** a first live `dev` deploy is being attempted, so this is no longer indefinitely deferred. The deploy itself is `needs-human` (the `dev` Environment's OIDC variables and the one-time `create-model-deployments=true` dispatch are owner steps — #109); every checklist item below is agent-verifiable once the environment is up.

## To verify on the first real dev deploy (`infra/parameters/dev.bicepparam`, `createModelDeployments=false` after day 0)

- [ ] Container App `ca-foundrygate-api-dev` provisions with the placeholder image, then with the real `crfoundrygatedeve7k2.azurecr.io/foundrygate-api:<sha>` (AcrPull via the user-assigned identity; no registry password)
- [ ] Container App env vars resolve: `AZURE_CLIENT_ID`, `Azure__KeyVaultUrl`, `ConnectionStrings__FoundryGate` (Entra auth), `AzureAd__*`, `Cors__AllowedOrigins__0` (SWA hostname), `OpenTelemetry__*`, `Gateway__*`
- [ ] `/health` and `/health/ready` probes pass once the API image runs (readiness needs the DB contained user — see the companion issue)
- [ ] Flex Consumption Function App `func-foundrygate-dev-e7k2` provisions with `allowSharedKeyAccess=false` storage: identity-based `AzureWebJobsStorage__*` + identity-authenticated deployment container — confirm the host starts and `Azure/functions-action` publishes
- [ ] `.NET 10` isolated (`functionAppConfig.runtime.version: '10.0'`) accepted by Flex Consumption in eastus2 (fallback: `9.0` param)
[...]

Issue #78, opened 2026-06-04
Add FoundryGate.Cli dotnet tool: foundrygate db compare / deploy / seed / local setup


kolatts:
Part of #76.

Create `src/FoundryGate.Cli` as a packable dotnet tool (`PackAsTool: true`, `ToolCommandName: foundrygate`). Uses `Spectre.Console.Cli` for commands, `Microsoft.SqlServer.DacFx` for schema comparison and dacpac deployment, `LibGit2Sharp` for reverting trivial ordering-only diffs. Four commands: `foundrygate db compare` (DacFx SchemaComparison local DB → .sqlproj, publishes delta, reverts ordering noise — Windows-only with graceful error on other OSes); `foundrygate db deploy` (DacServices dacpac deploy, supports SQL auth and DefaultAzureCredential, `--drop` flag); `foundrygate db seed` (upserts SystemConfiguration rows); `foundrygate local setup` (Docker SQL container → EF migrate → compare → seed).

Issue #75, opened 2026-06-04
Write deploy-all.yml: full-stack ordered deploy via workflow_call


kolatts:
Part of #67.

`.github/workflows/deploy-all.yml`. `workflow_dispatch` only. Input: `environment` (dev|prod). Uses `workflow_call` to invoke each pipeline in dependency order: `infra-deploy.yml` first, then `api-deploy.yml`, `functions-deploy.yml`, and `ui-deploy.yml` in parallel (all `needs: [infra]`), then `docs-deploy.yml` independently (no Azure dependency). Designed for rebuilding a fresh environment after an `infra-destroy` run. Each called workflow receives the `environment` input and the caller inherits its GitHub Environment gates — so a prod full-stack deploy still requires the production reviewer approval.

kolatts:
Claimed: implementing in branch claude/cicd-deploy-workflows-67 (session https://claude.ai/code/session_01RtuGhG69SaSuVfx92RwwNk). One PR will close #68-#75 together.

PR #143, opened 2026-09-01
feat(cli): real ip setup/cleanup + db grant-identities, wired into _deploy-database.yml

### LINKED ISSUE #96: Implement foundrygate ip setup: whitelist runner/dev IP on Azure SQL Server firewall
Part of #76 / #67.

## Context

#78 added \`FoundryGate.Cli\`'s \`db deploy\`/\`db seed-reference\`/\`db seed-test\` commands and a **stub** \`ip setup --env <env>\` command (\`src/FoundryGate.Cli/Commands/Ip/Setup/SetupCommand.cs\`). The stub exists only so the reusable \`.github/workflows/_deploy-database.yml\` (#79) has the right shape to call — it prints guidance and exits non-zero rather than doing anything, and the workflow step that calls it is marked \`continue-on-error: true\` with a comment pointing at this issue.

The reason it's a stub: there is no real Azure SQL Server resource yet. That lands with the Bicep infra modules (#43, #44) and the environment/CI-identity wiring in the CI/CD epic (#67, #68, #69).

## Ask

Once the Azure SQL Server resource exists (post #43/#44 infra deploy), implement \`ip setup\` for real, modeled on imagile-app's \`Imagile.App.Cli\Commands\Ip\Setup\SetupCommand.cs\`:

1. Detect the caller's public IP address.
2. Build an \`ArmClient\` (Azure.Identity credential chain — Workload/Managed Identity in CI, Azure CLI locally, matching CONVENTIONS.md's `AppTokenCredential`-style pattern).
3. Resolve the target environment's SQL Server resource id (needs the Bicep-defined naming convention from #43/#44 — there's no `FoundryGateAzureResources`-style resource-name helper yet either; add one alongside this).
4. Create or update a `SqlFirewallRuleResource` scoped to that IP (rule name convention TBD — imagile-app names it after the caller, e.g. CI runner vs. a developer's machine).
5. Keep the `--env` and `--yes` options already on the stub command; wire them through instead of the stub's guidance message.
6. Update `.github/workflows/_deploy-database.yml`'s "Whitelist runner IP address" step to drop `continue-on-error: true` once this actually works, and remove the comment pointing here.

## Why a stub now instead of blocking #78/#79

#77/#78/#79 (the dacpac tables, CLI, and reusable deploy workflow) are all independently valuable and testable against local docker SQL without any Azure resources existing. Gating them on Bicep infra landing first would serialize unrelated work for no benefit — the workflow just needs the right *shape* now (a firewall step that runs, doesn't hard-fail the pipeline, and is easy to swap from stub to real later).

### LINKED ISSUE #106: db deploy: create contained users for the API and Functions managed identities after the dacpac
Follow-up to #43/#44. Azure SQL is Entra-only (`modules/sql.bicep`: `azureADOnlyAuthentication: true`, admin = Entra group, no SQL login). The API and Functions connect with `Authentication=Active Directory Default` using their user-assigned identities, which Bicep creates — but a managed identity cannot be granted database access through ARM. Someone has to run, inside `sqldb-foundrygate-{env}`:

```sql
CREATE USER [id-foundrygate-api-{env}] FROM EXTERNAL PROVIDER;
ALTER ROLE db_datareader ADD MEMBER [id-foundrygate-api-{env}];
ALTER ROLE db_datawriter ADD MEMBER [id-foundrygate-api-{env}];
-- same for [id-foundrygate-func-{env}]
```

## Ask

Add a post-dacpac step to `.github/workflows/_deploy-database.yml` (or a `db grant-identity` CLI command in `FoundryGate.Cli`, mirroring how `db seed-reference` runs after `db deploy`) that idempotently creates these users and role memberships. The identity names are outputs of `infra/main.bicep` (`apiIdentityName`, `functionsIdentityName`) and follow the convention `id-foundrygate-api-{env}` / `id-foundrygate-func-{env}` (see `modules/control-plane.bicep`).

Notes:
- `CREATE USER ... FROM EXTERNAL PROVIDER` requires the executing principal (the CI OIDC app, via the SQL Entra admin group) to have `Directory Readers` or the newer "no Directory Readers needed for service principals" path — verify against the live server; if it fails, assign Directory Readers to the SQL server's system-assigned identity (`modules/sql.bicep` would then need `identity: SystemAssigned`).
- Consider `db_ddladmin` only if the API ever needs to create objects at runtime (it should not — the dacpac owns schema).

kolatts:
## Summary

Replaces the `ip setup` stub from #78 with the real thing, adds `ip cleanup` and `db grant-identities`, and wires all three into `_deploy-database.yml`. Everything Azure- or SQL-shaped sits behind two small interfaces (`IAzureSqlServerClient`, `ISqlBatchExecutor`) so the command logic is unit-tested with fakes; no live Azure was touched (checklist filed as #142).

### CLI surface

| Command | Options |
|---|---|
| `foundrygate ip setup` | `--env <env>` (required) `[--yes\|-y] [--name <rule>] [--ip <ipv4>] [--server <name>] [--resource-group <rg>] [--subscription <id>]` |
| `foundrygate ip cleanup` | `--env <env>` (required) `[--older-than <hours>=2] [--dry-run] [--server] [--resource-group] [--subscription]` |
| `foundrygate db grant-identities` | `--env <env>` (required) `[--api-identity <name>] [--functions-identity <name>] [--api-identity-client-id <guid>] [--functions-identity-client-id <guid>] [--connection-string <cs>] [--dry-run] [--server] [--resource-group] [--subscription]` |

- **`ip setup`**: public IPv4 via api.ipify.org → ifconfig.me (IPv6 answers rejected; `--ip` overrides); server resolved by listing `rg-foundrygate-{env}` for the single `sql-foundrygate-{env}-*` match — zero or several matches is an error naming the candidates, `--server`/`--resource-group` bypass; idempotent (a rule already allowing the IP → no ARM write). Rule names: `gha-{run id}-{yyyyMMddHHmm}` on GitHub Actions, `fg-dev-{machine}-{user}` otherwise, sanitised to `[A-Za-z0-9_.-]`, ≤128. Confirmation prompt unless `--yes`; refuses to hang on a non-interactive stdin.
- **`ip cleanup`**: only `gha-*` rules are candidates. This run's own `gha-{GITHUB_RUN_ID}-*` go unconditionally; others once their embedded UTC minute is ≥ `--older-than` old, or when they carry no timestamp. `fg-dev-*`, `AllowAllWindowsAzureIps`, hand-made rules are never touched.
- **`db grant-identities`**: one batch per identity — `IF NOT EXISTS (sys.database_principals) … CREATE USER … FROM EXTERNAL PROVIDER` (or `WITH SID = <client id>, TYPE = E` when a client id is given — no Directory Readers lookup), then `IS_ROLEMEMBER`-guarded `ALTER ROLE db_datareader / db_datawriter ADD MEMBER`. **No `db_ddladmin`**: the dacpac owns every schema object and EF never migrates at runtime. Names default to `id-foundrygate-api-{env}` / `id-foundrygate-func-{env}`; the connection string defaults to the resolved server + `sqldb-foundrygate-{env}` with `Authentication=Active Directory Default`.
- `Helpers/FoundryGateAzureResources` mirrors the infra naming convention and maps the GitHub Environment `production` → Bicep `prod` (the workflow passes `inputs.environment` as `--env`). `CliTokenCredential` reuses `Imagile.Framework.Configuration.Azure.AppTokenCredential` (same as the Api) and ignores `AZURE_CLIENT_ID` when `GITHUB_ACTIONS=true`, where only the azure/login az session exists.

### Workflow (`_deploy-database.yml`)

- "Whitelist runner IP address" now runs the real `ip setup --env … --server … --resource-group … --yes`; `continue-on-error: true` and the #96 stub comment are gone.
- 60 s propagation wait and TCP 1433 probe unchanged.
- New "Grant database access to the API and Functions identities" step after seeding, over the same `CONNECTION_STRING`.
- New final "Remove runner firewall rules" step: `if: always() && steps.login.outcome == 'success' && steps.cli.outcome == 'success'` → `ip cleanup --older-than <input>`.
- `workflow_call` contract is backward compatible: five new inputs, all defaulted (`api-identity-name`, `functions-identity-name`, `api-identity-client-id`, `functions-identity-client-id`, `firewall-rule-max-age-hours`). `AZURE_SUBSCRIPTION_ID` exported at job level for the CLI's `ArmClient`. Inputs reach `run:` scripts through `env:`; header comments rewritten to match.

### Docs

- `reference/infrastructure.md`: "not expressible in Bicep" list is now a table marking which steps the CLI automates; firewall model section gains the rule naming/lifetime table and the auth/RBAC notes; outputs table points the identity outputs at the new workflow inputs. Docs build passes.
- `plans/23-database-tooling.md`: status/verification notes updated; the stub line replaced.

## Verification

- `dotnet build FoundryGate.sln -c Release` — 0 warnings, 0 errors.
- `dotnet test src/FoundryGate.Tests.Predeployment -c Release --no-build` — **229/229** (main: 130; +99 in `Cli/*`: naming convention incl. aliases, rule naming/sanitising/timestamps, `ip setup` create/update/no-write idempotency/prompt/overrides/zero-and-ambiguous resolution, `ip cleanup` classification/dry-run/explicit server, generated T-SQL exact text + escaping + role list, runner ordering/dry-run/failure stop, public-IP fallback chain, credential selection).
- `actionlint` 1.7.7 on `_deploy-database.yml` — clean.
- `npm run build` in `docs-site/` — 13 pages, clean.
- CLI smoke: `--help` for all three commands; `db grant-identities --env production --dry-run …` prints the `prod` batches; bad `--ip` / bad `--env` exit 1 with a clear message.
- **Generated T-SQL executed against docker SQL Server 2022** (the repo's `docker-compose.yml`): pre-created stand-in users (incl. one named `odd]name'x`), ran the exact batches twice → both users `db_datareader`=1, `db_datawriter`=1, `db_ddladmin`=0, second run a clean no-op; `WITH SID` variant compile-checked with `SET NOEXEC ON` (its dynamic statement renders as `CREATE USER [id-foundrygate-api-dev] WITH SID = 0x1111…, TYPE = E;`). This probe caught a real bug — `EXEC(<expr>)` rejects function calls (`Incorrect syntax near 'QUOTENAME'`) — fixed by assembling into `@sql` + `sp_executesql`. Container torn down afterwards.

## Deferred / follow-ups

- #142 — live validation checklist against `sql-foundrygate-dev-e7k2` (firewall write RBAC, propagation timing, `FROM EXTERNAL PROVIDER` by a service principal vs. the `WITH SID` path / Directory Readers, readiness probe going green). Companion to #105.

## Decisions for the reviewer

1. **Rule naming** deviates slightly from the #96 comment's `ci-<runner id>` / `dev-<upn>` suggestion: `gha-{run id}-{UTC minute}` and `fg-dev-{machine}-{user}`. The timestamp is in the CI name on purpose — ARM has no creation time on a firewall rule, and `ip cleanup --older-than` needs one.
2. **`ip cleanup` treats an un-timestamped `gha-*` rule as stale** (nothing else would ever remove one). Developer/hand-made rules are never candidates.
3. **`WITH SID` alternative** for contained users is exposed (`--*-identity-client-id`, matching workflow inputs) because #106 flags the Directory Readers risk when a service principal issues `CREATE USER … FROM EXTERNAL PROVIDER`. Default path is still `FROM EXTERNAL PROVIDER`; the caller workflow (#68–#75) can pass `apiIdentityClientId`/`functionsIdentityClientId` outputs if live validation says so.
4. **No `db_ddladmin`**, per #106's own note and CONVENTIONS.md's schema pipeline.
5. **Public types in the Cli** for the tested logic (runners, naming, SQL builder, interfaces) instead of `InternalsVisibleTo`: the test project already binds `WebApplicationFactory<Program>` to the Api's `Program`, and making the Cli's top-level-statement `Program` visible would make that reference ambiguous. Command classes stay `internal`.
6. **`production` → `prod` alias** in the naming helper, because the workflow's `inputs.environment` is the GitHub Environment name while every Azure resource uses the Bicep `environmentName`.

Closes #96
Closes #106

### DIFF
--- a/.github/workflows/_deploy-database.yml
+++ b/.github/workflows/_deploy-database.yml
@@ -23,10 +23,18 @@ name: Deploy Database (Reusable)
 #   - Membership of the SQL server's Entra admin group (#109) — `db deploy`, the seeders and
 #     `db grant-identities` all connect with Authentication=Active Directory Default; there is no
 #     password fallback by design.
-#   - For `CREATE USER ... FROM EXTERNAL PROVIDER` issued BY a service principal, Azure SQL must be able
-#     to look the identity up in Entra: either the server has a managed identity with Directory Readers,
-#     or the caller passes the identities' client ids (api-identity-client-id / functions-identity-client-id,
-#     both `az deployment sub show` outputs) so the users are created WITH SID and no lookup happens.
+#
+# CALLER CONTRACT — the identity client ids are effectively REQUIRED:
+#   `db grant-identities` creates each contained user WITH SID = <client id>, TYPE = E, which asks Entra
+#   for nothing. The alternative, `CREATE USER ... FROM EXTERNAL PROVIDER`, makes Azure SQL resolve the
+#   identity's NAME in Entra, and that requires the LOGICAL SERVER's own managed identity to hold the
+#   Directory Readers directory role — `infra/modules/sql.bicep` declares no server identity at all, so
+#   that path fails against FoundryGate's infrastructure as deployed (#106/#142).
+#   Callers must therefore pass `api-identity-client-id` / `functions-identity-client-id` from the
+#   deployment outputs `apiIdentityClientId` / `functionsIdentityClientId` (`az deployment sub show`,
+#   exported by the deploy orchestration in #144). The grant step fails fast with that message when they
+#   are empty rather than silently falling back. A fork that HAS granted its server Directory Readers can
+#   set `allow-external-provider: true` to take the name-resolution path deliberately.
 #
 # OIDC permissions: a reusable workflow cannot grant itself `id-token: write` — GitHub only issues
 # the ACTIONS_ID_TOKEN_REQUEST_URL/TOKEN that azure/login@v2 needs when the CALLING workflow (or
@@ -85,7 +93,7 @@ on:
         type: string
         default: ""
       api-identity-client-id:
-        description: "API identity client id (deployment output apiIdentityClientId). When set the user is created WITH SID, which needs no Directory Readers permission on the SQL server; empty = CREATE USER ... FROM EXTERNAL PROVIDER."
+        description: "API identity client id (deployment output apiIdentityClientId). The contained user is created WITH SID from it, which needs no Directory Readers permission on the SQL server. Required in practice: the grant step fails when it is empty unless allow-external-provider is true."
         required: false
         type: string
         default: ""
@@ -94,6 +102,11 @@ on:
         required: false
         type: string
         default: ""
+      allow-external-provider:
+        description: "Escape hatch for a fork whose SQL logical server HAS a managed identity holding the Entra Directory Readers role: creates the contained users with CREATE USER ... FROM EXTERNAL PROVIDER instead of requiring the client ids. FoundryGate's own infra/modules/sql.bicep declares no server identity, so the default is false."
+        required: false
+        type: boolean
+        default: false
       firewall-rule-max-age-hours:
         description: "gha-* firewall rules older than this are removed by the final `ip cleanup` step (this run's own rules are always removed)."
         required: false
@@ -249,15 +262,30 @@ jobs:
 
       # Idempotent T-SQL (IF NOT EXISTS ... CREATE USER; IS_ROLEMEMBER-guarded ALTER ROLE ... ADD MEMBER
       # for db_datareader + db_datawriter) over the same Entra connection the deploy used. The dacpac
-      # excludes Users (ExcludeObjectTypes) so `--drop-objects` never removes what this step creates.
+      # excludes both Users and RoleMembership (DacpacDeployOptions.ExcludedObjectTypes, unit-tested), so
+      # `--drop-objects` removes neither the users this step creates nor their role memberships.
+      #
+      # The client ids are the deployment outputs apiIdentityClientId / functionsIdentityClientId — see the
+      # CALLER CONTRACT in this file's header. Without them the CLI would have to use CREATE USER ... FROM
+      # EXTERNAL PROVIDER, which needs Directory Readers on a server identity FoundryGate's Bicep does not
+      # declare, so we stop here with an actionable message instead of failing deep inside the CLI.
       - name: Grant database access to the API and Functions identities
         env:
           API_IDENTITY_NAME: ${{ inputs.api-identity-name }}
           FUNCTIONS_IDENTITY_NAME: ${{ inputs.functions-identity-name }}
           API_IDENTITY_CLIENT_ID: ${{ inputs.api-identity-client-id }}
           FUNCTIONS_IDENTITY_CLIENT_ID: ${{ inputs.functions-identity-client-id }}
+          ALLOW_EXTERNAL_PROVIDER: ${{ inputs.allow-external-provider }}
         run: |
           ARGS=()
+          if [ "${ALLOW_EXTERNAL_PROVIDER}" = "true" ]; then
+            echo "::warning::allow-external-provider is set — contained users will be created with CREATE USER ... FROM EXTERNAL PROVIDER, which requires this SQL logical server's managed identity to hold the Entra Directory Readers role."
+            ARGS+=(--allow-external-provider)
+          elif [ -z "${API_IDENTITY_CLIENT_ID}" ] || [ -z "${FUNCTIONS_IDENTITY_CLIENT_ID}" ]; then
+            echo "::error::api-identity-client-id and functions-identity-client-id are required. Pass the deployment outputs apiIdentityClientId / functionsIdentityClientId to _deploy-database.yml so the contained users are created WITH SID; without them the only option is CREATE USER ... FROM EXTERNAL PROVIDER, which needs the Entra Directory Readers role on a SQL server identity infra/modules/sql.bicep does not declare (#106/#142). If your fork's server identity does have Directory Readers, set allow-external-provider: true."
+            exit 1
+          fi
+
           [ -n "${API_IDENTITY_NAME}" ] && ARGS+=(--api-identity "${API_IDENTITY_NAME}")
           [ -n "${FUNCTIONS_IDENTITY_NAME}" ] && ARGS+=(--functions-identity "${FUNCTIONS_IDENTITY_NAME}")
           [ -n "${API_IDENTITY_CLIENT_ID}" ] && ARGS+=(--api-identity-client-id "${API_IDENTITY_CLIENT_ID}")
--- a/.github/workflows/_deploy-database.yml
+++ b/.github/workflows/_deploy-database.yml
@@ -1,19 +1,32 @@
 name: Deploy Database (Reusable)
 
 # Reusable workflow_call: deploys the FoundryGate.Database dacpac to a target environment's Azure
-# SQL Server, then syncs reference data. Modeled on imagile-app's .github/workflows/_deploy-database.yml
-# (runner IP whitelist -> firewall propagation wait -> TCP probe -> dacpac deploy -> seed), minus
-# the metabase/tenant shard split (a FoundryGate fork IS the tenant, CONVENTIONS.md).
+# SQL Server, then syncs reference data and grants the app identities access. Modeled on imagile-app's
+# .github/workflows/_deploy-database.yml (runner IP whitelist -> firewall propagation wait -> TCP probe
+# -> dacpac deploy -> seed), minus the metabase/tenant shard split (a FoundryGate fork IS the tenant,
+# CONVENTIONS.md), plus two steps imagile-app does by hand:
+#   - `db grant-identities` after seeding: the contained database users for the API and Functions
+#     user-assigned managed identities (Entra-only server, #106 — Bicep cannot express this).
+#   - `ip cleanup` at the very end (`if: always()`): removes this run's `gha-*` firewall rule and any
+#     stale ones a cancelled/crashed earlier run left behind (#96). Developer `fg-dev-*` rules and the
+#     Bicep-declared `AllowAllWindowsAzureIps` are never touched.
 #
-# Expects two artifacts to already exist in the calling workflow run (produced by a build-dacpac
-# job and a CLI publish job — neither exists yet as of #77/#78/#79; the full-stack orchestration
-# lands with #67's CI/CD epic, e.g. deploy-all.yml):
+# Expects two artifacts to already exist in the calling workflow run (produced by the caller's build
+# jobs — the deploy orchestration in #68-#75):
 #   - "dacpac": contains FoundryGate.Database.dacpac (dotnet build src/FoundryGate.Database -o artifacts)
 #   - "foundrygate-cli": contains a framework-dependent publish of FoundryGate.Cli
 #     (dotnet publish src/FoundryGate.Cli --no-self-contained -o cli)
 #
-# Standalone-valid today: this file's shape (inputs, jobs, steps) does not depend on anything else
-# in #67 to be syntactically/semantically complete, only to actually run successfully end-to-end.
+# Azure permissions the OIDC principal needs (beyond what azure/login@v2 itself needs):
+#   - SQL Server Contributor (or Contributor) on the resource group — `ip setup`/`ip cleanup` are plain
+#     ARM writes to Microsoft.Sql/servers/firewallRules.
+#   - Membership of the SQL server's Entra admin group (#109) — `db deploy`, the seeders and
+#     `db grant-identities` all connect with Authentication=Active Directory Default; there is no
+#     password fallback by design.
+#   - For `CREATE USER ... FROM EXTERNAL PROVIDER` issued BY a service principal, Azure SQL must be able
+#     to look the identity up in Entra: either the server has a managed identity with Directory Readers,
+#     or the caller passes the identities' client ids (api-identity-client-id / functions-identity-client-id,
+#     both `az deployment sub show` outputs) so the users are created WITH SID and no lookup happens.
 #
 # OIDC permissions: a reusable workflow cannot grant itself `id-token: write` — GitHub only issues
 # the ACTIONS_ID_TOKEN_REQUEST_URL/TOKEN that azure/login@v2 needs when the CALLING workflow (or
@@ -36,7 +49,7 @@ on:
   workflow_call:
     inputs:
       environment:
-        description: "GitHub Environment name (dev, production, ...) — provides the deployment gate and scopes environment secrets."
+        description: "GitHub Environment name (dev, production, ...) — provides the deployment gate and scopes environment secrets. Also passed to the CLI as --env; it maps 'production' to the Bicep environmentName 'prod' when deriving resource/identity names."
         required: true
         type: string
       sql-server-name:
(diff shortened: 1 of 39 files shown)


kolatts:
**APPROVE**

Real implementation with real seams, and the docker-SQL probe that caught `EXEC(<expr>)` choking on `QUOTENAME` is exactly the legwork that makes a "no live Azure was touched" PR trustworthy. The generated T-SQL is injection-safe end to end, the SID conversion is right, `ip cleanup` won't eat anyone's rules, and the `workflow_call` contract still fits #144's callers unchanged. One thing I'd settle before the first dev run, plus a docs invariant that got missed — neither blocks the merge.

---

**Issues**

- **Major** `.github/workflows/_deploy-database.yml` L26-29 · `Commands/Db/GrantIdentities/ContainedUserSql.cs` L64 — the default `FROM EXTERNAL PROVIDER` branch won't work against the infra as it stands, so this is less "verify live" than "we already know the answer": Microsoft's Directory Readers page says *"to enable Microsoft Entra object creation in SQL Database … on behalf of a Microsoft Entra application, the Directory Readers role … is required. If the role or permissions aren't assigned to the logical server identity, creating Microsoft Entra users in Azure SQL will fail"* — and `infra/modules/sql.bicep` gives the server no `identity:` block at all. The `WITH SID` hatch is the right answer and you built it, but #144's `export-outputs.sh` doesn't export `apiIdentityClientId`/`functionsIdentityClientId` and neither caller passes them, so it's plumbed to this workflow's edge and stops there — worth turning #142's open question into a wired decision plus an issue against #144. (Good news for #109: the same page says the server identity is *not* needed just to set up the Entra group admin, so the connect path is fine.)

- **Minor** `docs-site/src/components/DeployPipeline.astro` L32-39 · `docs-site/src/content/docs/architecture/overview.mdx` L236-240 — the diagram's Database stage still lists five sub-steps ending at "seed reference → seed test" and the overview prose narrates the same five, so the public site now under-describes a pipeline this PR gave two more steps to; CONVENTIONS.md names the deploy pipeline flow explicitly, and CLAUDE.md's progressive-docs invariant wants the moderate level moving alongside `reference/*`, not just the deep one.

- **Minor** `src/FoundryGate.Cli/Commands/Db/Deploy/DeployCommand.cs` L104, claim at `_deploy-database.yml` L252 — `ExcludeObjectTypes = [ObjectType.Users]` excludes the user but not `ObjectType.RoleMembership`, which DacFx treats as its own object type, so `--drop-objects` can still strip `db_datareader`/`db_datawriter` and leave the running API unprivileged from the dacpac step until `grant-identities` catches up several steps later — the comment's "never removes what this step creates" is half true, and #142's checklist should say so.

- **Minor** `src/FoundryGate.Cli/Helpers/FoundryGateAzureResources.cs` L38, L79 — `^[a-z0-9]{1,10}$` rejects any hyphenated or longer environment name, and both runners normalize before `AzureSqlServerResolver` (which ignores `--env` entirely once `--server` and `--resource-group` are supplied, which is exactly how the workflow calls it), so a fork on a `pre-prod` Environment fails on a value nothing would have read; normalizing lazily, where a name is actually derived, drops the coupling.

- **Nit** `src/FoundryGate.Cli/Commands/Ip/IPublicIpProvider.cs` L57-67 — `TryParseIpv4` accepts `0.0.0.0`, which is the one IPv4 value you don't want here: start == end == `0.0.0.0` is Azure's allow-all-Azure sentinel (your own `sql.bicep` L76-84 calls it "the magic 0.0.0.0 rule"), so a typo'd `--ip` or a captive-portal body quietly creates a second allow-Azure rule wearing a `gha-*` name — and `HttpPublicIpProviderTests.cs` L79 currently pins that as correct.

- **Nit** `Commands/Ip/Setup/SetupCommand.cs` L83 · `Ip/Cleanup/CleanupCommand.cs` L78 · `Db/GrantIdentities/GrantIdentitiesCommand.cs` L112 — the exception filters miss `Azure.Identity.AuthenticationFailedException`, which derives straight from `Exception` rather than `RequestFailedException` (checked), so the single most likely failure in practice — no or expired `az` session behind `AppTokenCredential` — comes out as a raw stack trace instead of the tidy `ip setup failed: …` line the filters exist to produce.

- **Nit** `Commands/Db/GrantIdentities/ContainedUserSql.cs` L71-73 — the role name is the one value that reaches DDL by raw interpolation instead of `QUOTENAME`, in the file whose entire argument is that it never does that; a private const list today, an injection point the day someone makes roles configurable.

- **Nit** `Commands/Db/GrantIdentities/GrantIdentitiesCommand.cs` L105-107, L161-164 — `NoOpSqlBatchExecutor` is unreachable, because the runner already `continue`s past `_executor` on `dryRun`; two representations of one decision, and the dead one is the safety net.

- **Nit** `src/FoundryGate.Cli/Helpers/ArmAzureSqlServerClient.cs` L103-125 — every public method re-resolves subscription → resource group → server from ARM, so one `ip setup` is ~8 round-trips for 3 operations and `ip cleanup` pays two more per rule inside the delete loop; both handles are immutable for the client's lifetime, and `GetSqlServerResource(CreateResourceIdentifier(...))` constructs without a GET at all.

---

Verified locally against the branch: build clean (the `FoundryGate.Functions` WorkerExtensions MSB3030 reproduces on `origin/main` too — pre-existing, not yours), 341/341 Predeployment tests green, `dotnet format --verify-no-changes` clean, no `continue-on-error` left, and all five new workflow inputs defaulted so `api-deploy.yml` / `deploy-all.yml` call it unchanged. The `WITH SID` byte order checks out against Microsoft's service-principal tutorial — `CAST(CAST(sid AS varbinary(16)) AS uniqueidentifier)` round-trips to the app id, exactly the inverse of what you emit — and `AppTokenCredential(null)` really does resolve to `ChainedTokenCredential(AzureCliCredential, VisualStudioCredential)`, so the `GITHUB_ACTIONS` guard lands on the `azure/login` session the way the comment claims. I went looking for a `set -e` trap in the `[ -n "$X" ] && ARGS+=(...)` lines and there isn't one — the short-circuit is exempt mid-script — and I couldn't find a path where `ip cleanup` reaches `AllowAllWindowsAzureIps`, `fg-dev-*` or a hand-made rule.

kolatts:
Thanks — all nine addressed. Two commits: [`4fb5a77`](https://github.com/kolatts/foundry-gate/commit/4fb5a77) (code + workflow) and [`441f3fd`](https://github.com/kolatts/foundry-gate/commit/441f3fd) (docs), plus a merge of `origin/main`.

---

### Major — `FROM EXTERNAL PROVIDER` needs Directory Readers the infra doesn't declare

You're right that this was "we already know the answer", not "verify live", so it's now a wired decision rather than a deferred question. `WITH SID` is no longer the alternative — it's the path:

- **CLI**: `db grant-identities` **refuses** any grant without a client id. The message names the two options, the deployment outputs to pass, and why (`GrantIdentitiesRunner`, unit-tested). `FROM EXTERNAL PROVIDER` survives only behind the new `--allow-external-provider` flag, for an operator whose server identity really does hold Directory Readers.
- **Workflow**: the grant step passes `api-identity-client-id` / `functions-identity-client-id` and **fails with an `::error::`** naming `apiIdentityClientId` / `functionsIdentityClientId` when either is empty — no silent fallback. The escape hatch is exposed as an `allow-external-provider` input (defaults false, emits a `::warning::` when set).
- The contract is stated in the workflow header under a **`CALLER CONTRACT`** heading, so #144's callers have something to point at.
- The two inputs stay `required: false` at the YAML level on purpose: #144's callers are being updated concurrently, so failing loudly at *runtime* beats breaking the `workflow_call` signature under them.
- **#142** rewritten: the "pick one" fork is gone, replaced by the decision plus three checks (SID path works; omitting the ids fails with the message; the opt-in path if a fork takes it). `reference/infrastructure.md` gains a **"Why the client ids, not `FROM EXTERNAL PROVIDER`"** section. No separate issue against #144 — it's already in flight with the export.

### Minor — docs invariant, Database stage

`DeployPipeline.astro`'s Database stage now lists `grant identities (contained SQL users)` and `remove runner firewall rule`, and the sr-only narration matches. Deliberately scoped to that stage's `sub` array and the one sentence in the description — nothing near the trigger model #144's fix pass is editing. `architecture/overview.mdx`'s database paragraph names both steps and why Bicep can't express them, and `CONVENTIONS.md` §CI/CD's one-line chain is updated too.

### Minor — `db deploy` exclusions

`ObjectType.RoleMembership` added alongside `ObjectType.Users`. The options moved out of the action body into `DacpacDeployOptions.Create` so it *is* unit-testable, and `DacpacDeployOptionsTests` now pins both exclusions (plus the data-loss default and the `--drop-objects` mapping). The workflow comment now says "excludes both Users and RoleMembership … removes neither the users this step creates nor their role memberships", and #142's checklist item says the same.

### Minor — `NormalizeEnvironment` regex

Both halves of your suggestion, since they're complementary:

- `NormalizeEnvironment` is now trim + alias + lowercase, nothing else. A value that never reaches a name builder is nobody's problem.
- The shape rule moved to `ValidatedEnvironment`, called only by `ResourceGroupName` / `SqlServerNamePrefix` / `SqlDatabaseName` / `ApiIdentityName` / `FunctionsIdentityName`, and widened to `^[a-z0-9]([a-z0-9-]{0,22}[a-z0-9])?$` — inner hyphens, 1–24 chars. A fork on `pre-prod` or `integration-eu` now works, tested end to end through the derived names; `-dev`, `dev_2`, `dev env` and an over-long name still fail, but at the builder rather than at parse.

### Nits

- **`0.0.0.0`** — rejected in `TryParseIpv4` (`!parsed.Equals(IPAddress.Any)`), the pinning test flipped to `false`, and a new test proves a bad endpoint body of `0.0.0.0` falls through to the next endpoint rather than becoming a `gha-*` allow-all-Azure rule. Comment cites `sql.bicep`'s magic-rule note.
- **`AuthenticationFailedException`** — new `CliErrors.IsExpected` / `Describe`, used by all three filters. The credential case gets "No usable Azure credential was found — run `az login` locally, or make sure the job ran azure/login@v2 before this step in CI." (`CredentialUnavailableException` derives from it, so it's covered.)
- **`QUOTENAME` on roles** — role names now bind to `@role` and reach DDL via `@quotedRole = QUOTENAME(@role)`, exactly like the principal. New test asserts no `[db_datareader]` literal survives anywhere in the batch.
- **`NoOpSqlBatchExecutor`** — gone. The executor is `ISqlBatchExecutor?` and the runner requires it only for a non-dry run, so there's one representation of the decision.
- **ARM round-trips** — `ArmAzureSqlServerClient` caches the subscription and the resource group per invocation, and `GetServerResourceAsync` now builds the handle from `SqlServerResource.CreateResourceIdentifier(...)` — zero round-trips, per your note. `ip cleanup`'s delete loop pays nothing per rule now.

---

**Verification** (post-merge of `origin/main`): `dotnet build FoundryGate.sln -c Release` 0 warnings / 0 errors · `dotnet test src/FoundryGate.Tests.Predeployment -c Release --no-build` **594/594** (357 before the merge brought in main's new suites; +12 net from this pass) · `dotnet format --verify-no-changes` clean · `actionlint` 1.7.7 clean on `_deploy-database.yml` · `npm ci && npm run build` in `docs-site/` 13 pages · CI `build-test`, `docs-build`, `Build Astro site` all green on the PR.

**Still open:** #142's live run is unchanged in spirit — it just has a shorter, more decided checklist now. The one real dependency is #144 exporting `apiIdentityClientId` / `functionsIdentityClientId` and passing them through; until it does, the grant step fails fast, which is the intended behaviour rather than a regression.

Issue #145, opened 2026-09-01
Retire RequestDtoBindingController actions as production controllers bind the request records


kolatts:
Follow-up from #128. `src/FoundryGate.Tests.Predeployment/Support/RequestDtoBindingController.cs` is a test-only controller (registered as an application part by `ApiTestFactory`) that binds the seven request records which have no production controller yet, so `RequestDtoBindingTests` can prove each one is a 400 — not a 500 — through the real MVC pipeline.

As each wave lands its production controller, its own endpoint tests should assert the invalid-body 400 for the record it binds, and the matching action (and `InvalidBodies` row) here should be deleted so the same record is not covered twice:

- users wave (#29): `UpdateUserQuotaRequest` → `users/quota`
- groups wave: `CreateGroupRequest`, `UpdateGroupRequest`, `AddGroupMemberRequest` → `groups`, `groups/update`, `groups/members`
- requests wave: `SubmitQuotaIncreaseRequest`, `ReviewQuotaIncreaseRequest` → `requests`, `requests/review`
- config wave: `UpdateSystemConfigRequest` → `config`

When the last action goes, delete the controller, the test class and the `AddApplicationPart` line in `ApiTestFactory.ConfigureTestServices`. The `DomainArchitectureTests` guard stays — it is what keeps the bug from coming back regardless of test coverage.

kolatts:
Claiming this on branch `claude/hardening-entra-groups-config` — every remaining record now has a production endpoint, so the controller goes.

PR #146, opened 2026-09-01
fix(api): request DTOs validate as 400 instead of 500, and generated URLs are lowercase

### LINKED ISSUE #128: Request DTOs as positional records with [property:] validation attributes make MVC throw 500 on every POST/PUT body
Found while implementing #61 — the first `POST` body under `/api/v1`. Every request DTO in `FoundryGate.Domain` is a **positional record** with validation attributes targeted at the generated properties (`[property: Required, ...]`). ASP.NET Core MVC rejects that shape at validation time — every `POST`/`PUT` with such a body is a **500**, before the action runs:

```
System.InvalidOperationException: Record type 'FoundryGate.Domain.Foundry.Contracts.CreateFoundryDeploymentRequest' has validation metadata defined on property 'Capacity' that will be ignored. 'Capacity' is a parameter in the record primary constructor and validation metadata must be associated with the constructor parameter.
   at Microsoft.AspNetCore.Mvc.ModelBinding.ModelMetadata.ThrowIfRecordTypeHasValidationOnProperties()
```

Nothing caught it so far because `RequestDtoValidationTests` uses `System.ComponentModel.DataAnnotations.Validator.TryValidateObject`, which validates **properties** — and so does Blazor's `DataAnnotationsValidator` (the reason `ValidationConstants` is "shared between API DataAnnotations attributes and Blazor form validation"). Moving the attributes to the constructor parameters (dropping `property:`) satisfies MVC but silently disables both `Validator` and the Blazor forms.

The placement all three agree on is a **non-positional record with init properties** — done for `CreateFoundryDeploymentRequest` in the #61 PR (see its `<remarks>`):

```csharp
public record CreateFoundryDeploymentRequest
{
    [Required, StringLength(64, MinimumLength = 2), RegularExpression(...)]
    public string DeploymentName { get; init; } = string.Empty;
    [Range(1, ValidationConstants.FoundryDeploymentMaxCapacity)]
    public int Capacity { get; init; }
}
```

## Affected (grep `\[property:` in `src/FoundryGate.Domain`)

- `Users/Contracts/UserContracts.cs` — `UpdateUserQuotaRequest`
- `Groups/Contracts/GroupContracts.cs` — `CreateGroupRequest`, `UpdateGroupRequest`, `AddGroupMemberRequest`
- `Requests/Contracts/QuotaIncreaseRequestContracts.cs` — `SubmitQuotaIncreaseRequest`, `ReviewQuotaIncreaseRequest`
- `Config/Contracts/ConfigContracts.cs` — `UpdateSystemConfigRequest`

## Ask

- Convert each to an init-property record (keep `with` semantics; `= string.Empty` defaults so a missing JSON field is a field-level `[Required]` error rather than a `$`-level deserialization error).
- Update `RequestDtoValidationTests` and the Blazor callers (`FoundryGate.Web` constructs some of these positionally).
- Add a convention test in `FoundryGate.Tests.Predeployment/Domain` that fails for any Domain record with both a primary constructor and a validation attribute on a property, so this can't recur.
- Note the rule in CONVENTIONS.md §API service/controller conventions.

Worth doing before the quota (#32/#33), keys (#36/#37) and users/groups endpoint waves land their first `POST`s.

### LINKED ISSUE #129: Generated URLs (201 Location headers) keep the [controller] token's casing — set RouteOptions.LowercaseUrls host-wide
Found while implementing #61. `ApiControllerBase` routes `api/v1/[controller]`; routing matches case-insensitively, but **URL generation** (`CreatedAtRoute`/`CreatedAtAction`/`Url.Link`) renders the `[controller]` token with the class-name casing, so `POST /api/v1/foundry/deployments` returns

```
Location: http://<host>/api/v1/Foundry/deployments/{accountName}/{deploymentName}
```

while `reference/api.md` documents lowercase paths. Harmless functionally (the `FoundryEndpointTests` Location assertion is case-insensitive for this reason) but sloppy on the wire, and every future `201` will have the same shape.

## Ask

One line in `Program.cs` (host-wide, so not done from a feature branch):

```csharp
builder.Services.Configure<RouteOptions>(options => options.LowercaseUrls = true);
```

then tighten the `Location` assertion in `FoundryEndpointTests.Admin_create_returns_201_with_Location_and_writes_the_audit_row` back to `StringComparison.Ordinal`.

kolatts:
## Summary

Two host-wide fixes found while landing #61's first `POST` body, done before the users/groups/requests/config waves land theirs.

**#128 — request DTOs were 500s.** Every positional `*Request` record with `[property: Required/StringLength/Range/...]` made MVC throw at bind time (`Record type ... has validation metadata defined on property ... that will be ignored`) — a 500 on every POST/PUT of that type, before the action ran. Attributes on the constructor parameters would satisfy MVC but hide them from `DataAnnotations.Validator` and Blazor's `DataAnnotationsValidator`. The one placement all three agree on is the **init-property record** `CreateFoundryDeploymentRequest` already used:

```csharp
public record CreateGroupRequest
{
    [Required]
    [StringLength(ValidationConstants.GroupNameMaxLength, MinimumLength = 1)]
    public string Name { get; init; } = string.Empty;
    ...
}
```

- Converted `UpdateUserQuotaRequest`, `CreateGroupRequest`, `UpdateGroupRequest`, `AddGroupMemberRequest`, `SubmitQuotaIncreaseRequest`, `ReviewQuotaIncreaseRequest`, `UpdateSystemConfigRequest` — every attribute, default and XML doc preserved (`<param>` → property `<summary>`). Response records stay positional (MVC never binds them). `FoundryGate.Web` compiles unchanged (it only passes these records through; no page constructs one yet).
- `RequestDtoValidationTests` → object-initializer construction, plus an "empty body binds to an instance whose only error is `[Required] Name`" case.
- **`DomainArchitectureTests`** gains a reflection guard over every Domain record: a validation attribute on a positional parameter, *or* on the property a positional parameter generates, fails the build naming the record/member.
- **`RequestDtoBindingTests`** proves the bug is gone through the real pipeline: the test assembly is registered as an application part of the `ApiTestFactory` host and a test-only `Support/RequestDtoBindingController` binds each of the seven records; a Theory posts an invalid body per record and asserts **400 ValidationProblemDetails naming the member, never 500**, and a Fact round-trips a valid body (record equality).
- `CONVENTIONS.md` → "API service/controller conventions": request DTOs are init-property records with attributes on the properties; response DTOs are positional records.

**#129 — `Location: .../api/v1/Foundry/...`.** `RouteOptions.LowercaseUrls = true` (`LowercaseQueryStrings = false`) in `Program.cs`; the `FoundryEndpointTests` create test now asserts the `Location` **ordinally** — using a mixed-case deployment name, so it also pins that `LowercaseUrls` lowercases route values and that the lowercased `Location` still resolves (200 on follow). CONVENTIONS bullet records the consequence: anything a route value identifies must resolve case-insensitively (Foundry account and deployment names do — service, ARM client and fake are all `OrdinalIgnoreCase`).

## Verification

- `dotnet build FoundryGate.sln -c Release` → **0 Warning(s), 0 Error(s)**
- `dotnet test src/FoundryGate.Tests.Predeployment -c Release --no-build` → **Passed: 249, Failed: 0** (+11 new: 8 Theory rows + 1 Fact in `RequestDtoBindingTests`, 1 default-instance test, 1 architecture guard)
- `dotnet format FoundryGate.sln --verify-no-changes` → clean
- **Mutation check:** temporarily reverting `ReviewQuotaIncreaseRequest` to the positional `[property: …]` shape failed exactly the two guards — `DomainArchitectureTests` ("`ReviewQuotaIncreaseRequest.ReviewNotes`: validation attribute on a positional record's property — MVC returns 500…") and `RequestDtoBindingTests` (`Expected: BadRequest, Actual: InternalServerError`) — and nothing else. Restored before commit.

## Decisions for the reviewer

- **Test-only controller lives in the test project** (`Support/RequestDtoBindingController`, wired via `services.AddControllers().AddApplicationPart(typeof(ApiTestFactory).Assembly)` at the end of `ConfigureTestServices`). No production surface is touched; the second `AddControllers()` is idempotent and leaves Program.cs's global `AuthorizeFilter` in place (the binding tests use an authenticated client). Each production wave retires its action here when its own endpoint tests bind the record — tracked in #145.
- **The architecture guard scans every Domain record, not just `*Request`** — there is no legitimate reason for any positional record to carry validation attributes, and the failure message says how to fix it. It also flags attributes on constructor *parameters* (the MVC-only placement that silently disables Blazor forms).
- **`LowercaseUrls` lowercases route values too**, not just the `[controller]` token. Today that's safe (all lookups behind route values are case-insensitive) and the create test pins it; the CONVENTIONS bullet makes it a rule for future routes.
- **Concurrency with #127/#133**: only the request records changed, no reordering of surrounding files; the CONVENTIONS bullets are appended at the *end* of the section (PR #127's bullet lands mid-section) and the `ApiTestFactory` line is at the end of `ConfigureTestServices` (PR #133 adds mid-block), so both merges stay trivial.

Closes #128
Closes #129

### DIFF
--- a/CONVENTIONS.md
+++ b/CONVENTIONS.md
@@ -173,6 +173,9 @@ Landed with #42 (the first `/api/v1` controller); every endpoint wave builds on 
   `DomainArchitectureTests` fails the build for either broken placement;
   `Api/Endpoints/RequestDtoBindingTests` posts each request record through the real pipeline
   (via a test-only controller the factory registers) and asserts 400, never 500.
+- **Generated URLs are lowercase** (`RouteOptions.LowercaseUrls`, #129): `CreatedAtRoute`'s
+  `Location` reads `/api/v1/foundry/...`, route values included — so anything a route value
+  identifies must resolve case-insensitively (Foundry account and deployment names do).
 
 ## Schema pipeline (no EF migrations)
 
--- a/CONVENTIONS.md
+++ b/CONVENTIONS.md
@@ -161,6 +161,18 @@ Landed with #42 (the first `/api/v1` controller); every endpoint wave builds on 
   `Program.cs`'s first line, which is also how `UseEnvironment` gets through. External
   systems (ARM, APIM, Graph) are fakes owned by the factory (`FoundryClient` is the
   precedent) — never a live client.
+- **Request DTOs are init-property records; response DTOs are positional records.** A
+  `*Request` body is `public record X { [Required, StringLength(n)] public string Name
+  { get; init; } = string.Empty; … }` — attributes on the properties, `= string.Empty` on
+  every non-nullable string so a missing JSON field is a field-level `[Required]` 400 rather
+  than a `$`-level deserialization error. Never a positional record with `[property: …]`
+  (MVC throws at bind time → 500 on every body of that type, #128) and never attributes on
+  the constructor parameters (MVC is happy, but `Validator` and Blazor's
+  `DataAnnotationsValidator` read properties, so nothing validates). Responses are never
+  bound by MVC, so they stay positional (`public record Y(int Id, string Name);`).
+  `DomainArchitectureTests` fails the build for either broken placement;
+  `Api/Endpoints/RequestDtoBindingTests` posts each request record through the real pipeline
+  (via a test-only controller the factory registers) and asserts 400, never 500.
 
 ## Schema pipeline (no EF migrations)
 
--- a/src/FoundryGate.Api/Program.cs
+++ b/src/FoundryGate.Api/Program.cs
@@ -83,6 +83,18 @@ builder.Services.AddFoundryGateHealthChecks();
 // authenticated by default; opt out per-action with [AllowAnonymous].
 builder.Services.AddControllers(options => options.Filters.Add(new AuthorizeFilter()));
 
+// URL generation (CreatedAtRoute → Location headers) renders the whole path lowercase, so a 201
+// points at /api/v1/foundry/... exactly as reference/api.md documents it rather than the
+// [controller] token's class-name casing (/api/v1/Foundry/...) (#129). Matching was always
+// case-insensitive, and so are the lookups behind every route value today (Foundry account and
+// deployment names resolve case-insensitively in the service and in ARM). Query strings keep their
+// casing: their values can be case-sensitive identifiers.
+builder.Services.Configure<RouteOptions>(options =>
+{
+    options.LowercaseUrls = true;
+    options.LowercaseQueryStrings = false;
+});
+
 // Global error handling (CONVENTIONS.md: "one IExceptionHandler + ProblemDetails, not
 // per-controller try/catch"). AddProblemDetails also shapes the framework's own
 // non-exception error responses (e.g. the 404 catch-all wired up via UseStatusCodePages below).
--- a/src/FoundryGate.Domain/Config/Contracts/ConfigContracts.cs
+++ b/src/FoundryGate.Domain/Config/Contracts/ConfigContracts.cs
@@ -17,10 +17,14 @@ public record SystemConfigEntryResponse(
     DateTimeOffset UpdatedDate,
     int? UpdatedByUserId);
 
-/// <summary>PUT /config/{key} body.</summary>
-public record UpdateSystemConfigRequest(
-    [property: Required, StringLength(ValidationConstants.ConfigValueMaxLength)]
-    string Value);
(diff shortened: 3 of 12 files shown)


kolatts:
**Orchestrator review — Approve.** Verified the diff: all seven request records converted to init-property records with every attribute/default preserved (`[Required]`/`[StringLength]`/`[Range]`/`[RegularExpression]`), responses left positional, `RouteOptions.LowercaseUrls=true`/`LowercaseQueryStrings=false`, pipeline regression test via a test-only application part (retirement tracked in #145), reflection guard in DomainArchitectureTests, CONVENTIONS bullets. Mutation check in the PR body is convincing. No findings ≥70. Merging.

Issue #147, opened 2026-09-02
Guard the one-pending-quota-request-per-period rule with a filtered unique index


kolatts:
`POST /requests` enforces "one pending request per user per period" with a read-then-write check in
`QuotaRequestService.SubmitCoreAsync`:

```csharp
if (await dbContext.QuotaIncreaseRequests.AnyAsync(r => r.UserId == ... && r.StatusType == Pending, ct))
{
    throw new ConflictException(...);
}
```

Two concurrent submissions from the same developer (double-clicked button, a retrying client) can
both pass the check and both insert, leaving two pending requests for one user and period — the
reviewer queue then shows the same person twice, and approving one leaves the other stranded until
someone rejects it by hand. Nothing is corrupted; it is a duplicate-row nuisance.

The database has no constraint behind the check: `QuotaIncreaseRequests` indexes `UserId`,
`RequestedByUserId`, `ReviewedByUserId` and `QuotaIncreaseRequestUnique`, none of them unique across
`(UserId, PeriodYear, PeriodMonth)`.

**Fix:** a filtered unique index —
`CREATE UNIQUE NONCLUSTERED INDEX [IX_QuotaIncreaseRequests_PendingPerUserPeriod] ON [dbo].[QuotaIncreaseRequests]([UserId], [PeriodYear], [PeriodMonth]) WHERE [StatusType] = 0`
— declared on the entity with `HasIndex(...).HasFilter(...)`, mirrored into
`FoundryGate.Database/dbo/Tables/QuotaIncreaseRequests.sql`, with the service catching
`DbUpdateException` on the insert and re-throwing the same `ConflictException` (the
`QuotaAllocationService` "lost the race, adopt the winner" precedent). Check first that
`SchemaParityTests`' regex index parser understands a filtered index, and that the SQLite test
harness creates the partial index the same way — both may need a small extension.

Found while implementing #34/#35 (branch `claude/quota-requests-34-35`).

kolatts:
Claiming this on branch `claude/backend-sweep-a` (backend sweep A: #182, #193, #147, #159, #170, #108).

PR #148, opened 2026-09-02
feat(api): quota increase requests — submit, list, approve and reject

### LINKED ISSUE #34: Implement quota increase request submission and listing (developer + admin)
Part of #8.

`POST /requests` lets any authenticated user (or admin acting on behalf of a user) submit a `QuotaIncreaseRequest` with `requestedQuota` (null = requesting unlimited) and a required `justification` string. Validate that no pending request already exists for the same user/period. `GET /requests` returns: all requests for admins (paged, filterable by status); only the caller's own requests for non-admins. `GET /requests/{id}` is accessible to the request owner or any admin.

### LINKED ISSUE #35: Implement admin approve and reject endpoints with quota re-resolution
Part of #8.

`PUT /requests/{id}/approve` (admin only) sets `Status = Approved`, records `ReviewedBy` and `ReviewedAt`, updates `User.MonthlyTokenQuota` or `User.IsUnlimited` to the approved value, and immediately calls the quota resolution service to update the current-period `QuotaAllocation`. `PUT /requests/{id}/reject` sets `Status = Rejected` with optional `ReviewNotes`. Both actions write an audit log entry with the reviewer and the before/after quota values.

kolatts:
Implements the quota increase workflow end to end: a developer asks to move up a budget tier, an
admin approves or rejects, and an approval takes effect immediately.

Closes #34
Closes #35

## What landed

**`Services/Requests/IQuotaRequestService`** (per-area folder, one `AddRequestsServices()` line in
`ApiServiceCollectionExtensions`):

```csharp
Task<QuotaIncreaseRequestResponse> SubmitAsync(SubmitQuotaIncreaseRequest request, CancellationToken ct);
Task<QuotaIncreaseRequestResponse> SubmitForUserAsync(int userId, SubmitQuotaIncreaseRequest request, CancellationToken ct);
Task<PagedResult<QuotaIncreaseRequestResponse>> ListAsync(QuotaRequestQuery filter, PagedRequest paging, CancellationToken ct);
Task<QuotaIncreaseRequestResponse> GetAsync(int quotaIncreaseRequestId, CancellationToken ct);
Task<QuotaIncreaseRequestResponse> ApproveAsync(int quotaIncreaseRequestId, ReviewQuotaIncreaseRequest review, CancellationToken ct);
Task<QuotaIncreaseRequestResponse> RejectAsync(int quotaIncreaseRequestId, ReviewQuotaIncreaseRequest review, CancellationToken ct);
Task<int> CancelPendingForUserAsync(int userId, string note, CancellationToken ct);
```

**`Controllers/RequestsController`** → `/api/v1/requests`: `GET` (paged, `?status=`, `?userId=`),
`GET /{id}`, `POST`, `POST /for/{userId}` (admin), `POST /{id}/approve`, `POST /{id}/reject` (admin).

**Domain**: `QuotaRequestQuery(QuotaRequestStatusType? Status, int? UserId)` alongside the existing
request contracts (positional, no validation attributes — the `AuditLogQuery` precedent).

- **A request asks for a tier, not a number (D-013).** `RequestedQuota` is `null` (unlimited) or
  exactly a configured tier cap; `GatewayTierMapper.EnsureValidQuota` guards submission **and**
  approval (CONVENTIONS names "request approval" as a write path, and the tier table is configuration
  that can change between filing and review). It must also be a genuine increase — bigger finite tier
  or unlimited; an already-unlimited caller gets a `400` saying there is nothing larger to ask for.
- **Review actions are `POST`, not the spec's `PUT`.** Non-idempotent state transitions whose body is
  not the resource — the shape `POST /users/{id}/activate`, `POST /keys/{userId}/rotate` and
  `POST /quota/reset` already use. Re-sending one is a `409`, not a no-op. `reference/api.md` and
  `plans/08` record the deviation.
- **Someone else's request is `404`, not `403`.** Byte-for-byte the message an unknown id produces, so
  `GET /requests/{id}` cannot be used to enumerate other people's requests. A non-admin passing
  `?userId=` naming another user *is* `403` — they asked for a queue rather than probing an id.
- **Submit is the one path with two saves, inside one transaction.** The audit row's `TargetId` is the
  request's identity PK, which does not exist until the insert has run. CONVENTIONS' rule is that a
  mutation must never exist without its audit row, not that there must be exactly one save — so the
  insert and the audit row commit inside a single `BeginTransactionAsync`. Approve/reject keep the
  normal single save (the id is already known).
- **Deactivated users.** Self-submit by a deactivated developer is `403` (the same answer
  `GET /quota/allocations/me` gives, and for the same reason: submitting would mint them an allocation
  and, once #118 lands, a tier sync onto a product). Admin-on-behalf and approval for a deactivated
  subject are `409` — the admin is entitled, the target just isn't in a state to hold a budget.
  Rejection has no such guard: closing a request has no side effects.
- **Submission captures `currentQuota` from the current-period allocation**, resolving and creating it
  if this is the developer's first activity of the month — so the recorded "before" is the same number
  their dashboard shows, and the allocation commits with the request.
- **`CancelPendingForUserAsync` neither saves nor audits.** It marks pending requests `Rejected` with
  the caller's note and no `ReviewedByUserId` (no human decided them; the status enum has no
  "Cancelled") and returns the count, leaving the commit and the audit row to the orchestrating
  lifecycle service. **@users/lifecycle agent: this is the seam for #65's deprovision path** — it
  currently cancels pending requests inline; adopting this keeps the cancellations inside its own unit
  of work and described by its own audit row.
- **Demo-data fix.** `TestDataSeeder`'s "Asked for more" request was filed by the *unlimited*
  developer asking for Power — a request `POST /requests` would refuse twice over. It now comes from a
  Standard-tier developer, matching the landing page's 5M-token "Asked for more" row.
  `TestDataSeederTests` pins the invariant.

## Verification

- `dotnet test FoundryGate.Tests.Predeployment -c Release` — **538 passed, 0 failed** (63 new: 35
  service-level, 28 endpoint)
- `npm ci && npm run build` in `docs-site/` — 13 pages built

Covered: submit happy paths (finite→finite, finite→unlimited), `400` on a non-tier value listing the
tiers, `400` on "not an increase" and on "already unlimited", `409` on a duplicate pending request and
none on a decided or earlier-period one, `201` + lowercase `Location` that actually resolves; list
scoping admin vs owner with both filters and paging; `404` for a non-owner read; approve applies the
quota, re-resolves the period, records the `IGatewayTierSync` call and audits before/after; approve
and reject `409` on a decided request; reject leaves the quota alone and frees the slot;
`CancelPendingForUserAsync` is idempotent, scoped to one user, and persists nothing until the caller
saves; the full auth matrix (401/403).

## Deferred, filed

- #147 — the one-pending-request-per-period rule is a read-then-write check with no database
  constraint behind it, so two simultaneous submissions can both land. Proposed fix (filtered unique
  index + `DbUpdateException` → `ConflictException`) is in the issue.

## Files not touched, deliberately

`Support/RequestDtoBindingController` still binds `SubmitQuotaIncreaseRequest` /
`ReviewQuotaIncreaseRequest` even though its doc says those actions "can go" once a production
controller binds the records — removing them would collide with the users and groups waves editing the
same theory data in parallel. The coverage is harmless; a later wave can retire all of it at once.

### DIFF
--- a/docs-site/src/content/docs/reference/api.md
+++ b/docs-site/src/content/docs/reference/api.md
@@ -82,11 +82,54 @@ the same row count.
 
 | Method | Path | Auth | Description |
 |---|---|---|---|
-| `GET` | `/requests` | Any | Own requests; Admin sees all. Filter: `?status=Pending|Approved|Rejected` |
-| `POST` | `/requests` | Any | Submit a quota increase request |
-| `GET` | `/requests/{id}` | Owner or Admin | Request detail |
-| `PUT` | `/requests/{id}/approve` | Admin | Approve with optional notes — updates quota immediately |
-| `PUT` | `/requests/{id}/reject` | Admin | Reject with optional notes |
+| `GET` | `/requests` | Any | Own requests; Admin sees all. Paged (`?page=&pageSize=`), newest first. Filters: `?status=` (`0` Pending, `1` Approved, `2` Rejected) and, for admins, `?userId=` |
+| `POST` | `/requests` | Any | Submit own request. Body: `requestedQuota` (null = unlimited), `justification` (10–2000 chars). `201` + `Location` |
+| `POST` | `/requests/for/{userId}` | Admin | Submit on a user's behalf — same rules, `requestedByUserId` is the admin. `201` + `Location` |
+| `GET` | `/requests/{id}` | Owner or Admin | Request detail. `404` for anyone else — see below |
+| `POST` | `/requests/{id}/approve` | Admin | Approve. Applies the tier to the user and re-resolves the current period immediately |
+| `POST` | `/requests/{id}/reject` | Admin | Reject. Nothing about the user's quota changes |
+
+**A request asks for a tier, not a number.** `requestedQuota` must be `null` (unlimited) or exactly
+one of the configured tier caps from `GET /quota/tiers` — anything else is `400` listing the allowed
+values (D-013; the same `EnsureValidQuota` guard as `PUT /users/{id}/quota`). It must also be a
+genuine increase over the requester's currently resolved quota: a bigger finite tier, or unlimited.
+A developer who is already unlimited has nothing larger to ask for, so their submission is `400`
+too. Approval re-checks the stored value against the tier table, so a request filed before a tier was
+removed from configuration is `400` at review time rather than persisting a budget the gateway cannot
+enforce.
+
+**One open request per user per period.** A second submission while one is still `Pending` in the
+same UTC calendar month is `409`; a decided (approved or rejected) request frees the slot at once, so
+a rejected developer can re-ask with a better justification. The check is not yet backed by a database
+constraint, so two simultaneous submissions can both land ([#147](https://github.com/kolatts/foundry-gate/issues/147)).
+
+Submission captures `currentQuota` from the requester's current-period `QuotaAllocation`, resolving
+and creating it if this is their first activity of the month — so the recorded "before" is the same
+number `GET /quota/allocations/me` shows. Every submission writes `quota.requested`; approval writes
+`quota.approved` with the before/after quota, the resolved tier product and whether the gateway
+subscription was moved; rejection writes `quota.rejected`. All three carry target type `Request` and
+the request id.
+
+**Approval is the write path.** It sets the subject's `IsUnlimited` / `MonthlyTokenQuota`, then re-runs
+quota resolution for the current period — which upserts the `QuotaAllocation` and moves the subject's
+APIM subscription to the new tier product — before the response is written. No cron job, no lag. The
+subject's reconciled `tokensUsed` is untouched. Everything (request, user, allocation, audit row)
+commits in one transaction; a failed gateway move fails the approval.
+
+Status codes worth knowing: `403` when the caller has no user row yet (call `GET /users/me` first) or
+is deactivated; `403` when a non-admin passes `?userId=` naming someone else; `409` when the request
+was already decided, or when the subject user is deactivated (re-activate them or reject the request);
+`404` on `GET /requests/{id}` for a request that belongs to someone else — deliberately identical to
+the response for an id that does not exist, so the route cannot be used to enumerate other people's
+requests.
+
+Deprovisioning hooks into the same service: `IQuotaRequestService.CancelPendingForUserAsync` closes a
+departing developer's pending requests so they do not sit in an admin's queue
+([#65](https://github.com/kolatts/foundry-gate/issues/65)/[#66](https://github.com/kolatts/foundry-gate/issues/66)).
+
+Spec §4.4 writes the two review actions as `PUT`; they are `POST` here — non-idempotent state
+transitions whose body is not the resource, matching `POST /users/{id}/activate`,
+`POST /keys/{userId}/rotate` and `POST /quota/reset` elsewhere in this API.
 
 ## Keys
 
--- a/docs-site/src/content/docs/reference/api.md
+++ b/docs-site/src/content/docs/reference/api.md
@@ -94,19 +94,32 @@ one of the configured tier caps from `GET /quota/tiers` — anything else is `40
 values (D-013; the same `EnsureValidQuota` guard as `PUT /users/{id}/quota`). It must also be a
 genuine increase over the requester's currently resolved quota: a bigger finite tier, or unlimited.
 A developer who is already unlimited has nothing larger to ask for, so their submission is `400`
-too. Approval re-checks the stored value against the tier table, so a request filed before a tier was
-removed from configuration is `400` at review time rather than persisting a budget the gateway cannot
-enforce.
+too.
+
+**Both rules are re-checked at approval, against live resolution — a request can never lower a
+budget.** A user's quota is far more volatile than the tier table: `PUT /users/{id}/quota`, a group's
+quota, or a change in group membership can all raise it between filing and review. So approval asks
+the resolution service what the subject's budget is *now* and refuses with `409` when applying the
+stored `requestedQuota` would not raise it — naming the current budget, so the reviewer can reject the
+request instead. A request filed at 5M→20M and approved after an admin made that developer unlimited
+would otherwise have silently demoted them. The tier check is re-run for the same reason: a request
+whose value is no longer a configured tier is `400` at review time rather than persisting a budget the
+gateway cannot enforce.
+
+**Both submission and approval read live resolution, not the stored `QuotaAllocation` row.** The
+allocation row is whatever the last resolution wrote — a group change since then makes it stale — so
+the `currentQuota` recorded on the request, and the comparison at review time, come from re-walking
+the five-level chain. That read is side-effect-free: submitting creates **no** allocation row and makes
+no gateway call, so a refused submission leaves nothing behind. The developer's allocation still
+appears on their first `GET /quota/allocations/me` of the month, as it always did.
 
 **One open request per user per period.** A second submission while one is still `Pending` in the
 same UTC calendar month is `409`; a decided (approved or rejected) request frees the slot at once, so
 a rejected developer can re-ask with a better justification. The check is not yet backed by a database
 constraint, so two simultaneous submissions can both land ([#147](https://github.com/kolatts/foundry-gate/issues/147)).
 
-Submission captures `currentQuota` from the requester's current-period `QuotaAllocation`, resolving
-and creating it if this is their first activity of the month — so the recorded "before" is the same
-number `GET /quota/allocations/me` shows. Every submission writes `quota.requested`; approval writes
-`quota.approved` with the before/after quota, the resolved tier product and whether the gateway
+Every submission writes `quota.requested`; approval writes `quota.approved` with the before/after
+quota, the quota and level resolution reported at review time, the tier product and whether the gateway
 subscription was moved; rejection writes `quota.rejected`. All three carry target type `Request` and
 the request id.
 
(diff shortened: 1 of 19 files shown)


kolatts:
## Review — quota increase requests (#34/#35)

**REQUEST CHANGES**

Well-built wave. The state machine is tight, the decisions are argued rather than assumed, and 63 tests is not a token gesture. One real bug though: approval applies the requested tier unconditionally, so it can *lower* someone's budget — and that goes live the moment the users and groups waves merge.

Verified on `origin/claude/quota-requests-34-35`: `dotnet build -c Release` clean, `dotnet test FoundryGate.Tests.Predeployment -c Release` 538/538, `dotnet format --verify-no-changes` clean. Findings 1–3 are each confirmed by a throwaway test, not just by reading.

---

**Issues**

- **Major** `src/FoundryGate.Api/Services/Requests/QuotaRequestService.cs:172-190` — `ApproveAsync` re-validates the stored `RequestedQuota` against the tier table but never re-checks that it is still an *increase*, so approval can silently downgrade the subject. Confirmed: user files 5M→20M, an admin (or a group) then makes them unlimited, approve sets `IsUnlimited = false, MonthlyTokenQuota = 20M` and re-resolves the allocation down to 20M. Two things make this more than theoretical. The PR argues the tier table can go stale between filing and review — the user's own quota is far more volatile, and `PUT /users/{id}/quota` and the group membership endpoints are both in the PRs open right next to this one. And because `GetCurrentQuotaAsync:351` measures against the stored allocation row rather than live resolution, a group change that happened *before* filing is already invisible: the `currentQuota` the reviewing admin sees can be wrong at submit time, no race required. Re-resolving at approval and 409-ing on "no longer an increase" fixes both halves.

- **Minor** `src/FoundryGate.Api/Services/Requests/QuotaRequestService.cs:369-380` — `LoadForReviewAsync` is a read-then-write with no row claim, so two concurrent reviews both pass the `!= Pending` check. A simultaneous approve and reject leaves the quota raised, the subscription moved, the row reading `Rejected`, and both a `quota.approved` and a `quota.rejected` audit row. This is the same shape as the submit-side race you filed as #147 — worth either the `ExecuteUpdateAsync` claim `ApimKeyService.cs:73-84` already uses, or its own issue so the pair get fixed together.

- **Minor** `src/FoundryGate.Api/Services/Requests/QuotaRequestService.cs:320` — `SubmitCoreAsync` calls `BeginTransactionAsync` unconditionally, so it throws if the caller already owns one. `ApimKeyService.cs:74` guards this exact case with `Database.CurrentTransaction is null ? Begin : null` precisely because an orchestrator may have one open — and this PR ships `CancelPendingForUserAsync` explicitly as an orchestrator seam, so orchestrators calling in is the stated design. Confirmed: an outer `BeginTransactionAsync` makes `SubmitAsync` throw today.

- **Minor** `src/FoundryGate.Api/Services/Requests/QuotaRequestService.cs:287-300` — the "not an increase" and "already unlimited" refusals happen *after* `GetCurrentQuotaAsync` has run a full resolution, which calls `IGatewayTierSync.SyncAsync`. CONVENTIONS is explicit: "resolve the actor and do every refusal *before* the call." Harmless while `NullGatewayTierSync` is registered; once #118 lands, a submission that 400s will have moved the caller's APIM subscription and then discarded the allocation row that justified it. Confirmed with a recording sync: a refused submit persists nothing and still records one gateway call.

- **Minor** `src/FoundryGate.Data/Entities/QuotaIncreaseRequest.cs:18` — the new `GET /requests` orders by `CreatedDate DESC` and filters on `StatusType`, and neither column is indexed (only the `QuotaIncreaseRequestUnique` unique index plus the FK ones). CONVENTIONS: "Index the columns your default ordering and filters use (`AuditLog.OccurredDate` is the precedent) and mirror the index in `FoundryGate.Database/dbo/Tables/*.sql`." Cheap to do here given #147 is going to touch this table's indexes anyway.

- **Nit** `src/FoundryGate.Api/Services/Requests/QuotaRequestService.cs:192` — approval resolves `BillingPeriod.Current` and ignores the request's own `PeriodYear`/`PeriodMonth`, and nothing expires pending requests, so a months-old request stays approvable and applies its stale tier to today's period while the response reports the old one. Either scope approval to the request's period or expire stale pending rows — a follow-up issue is fine.

- **Nit** `src/FoundryGate.Api/Services/Requests/QuotaRequestService.cs:194-210` — the audit row and `SaveChangesAsync` after `ResolveAsync` (which may have already moved the APIM subscription) run on the request's `cancellationToken`, where CONVENTIONS' external-commit-point rule calls for `CancellationToken.None`. `QuotaAllocationService` and `ApimKeyService` have the same gap today, so this reads like a repo-wide follow-up issue rather than something to fix here — worth filing while it's in view.

- **Nit** `src/FoundryGate.Data/Seeding/TestDataSeeder.cs:159` — `users.Find(...)` plus `if (requester is not null)` turns a broken demo-data invariant into silently missing demo data; `.First(...)` would fail loudly. `TestDataSeederTests` catches it either way, so this is purely about which failure you'd rather read.

- **Nit** `src/FoundryGate.Api/Services/Requests/QuotaRequestService.cs:305-341` — worth noting `QuotaIncreaseRequestUnique` is assigned at construction, so keying all three audit rows on the Guid would collapse submit to a single save and drop the transaction entirely. Not obviously better: every other area uses the int PK as `TargetId`, and `AuditTargetTypes.QuotaIncreaseRequest`'s own doc says the same, so the trade is one transaction against being the odd area out. Current call is defensible — just flagging the alternative.

- **Nit** `src/FoundryGate.Api/Services/Requests/IQuotaRequestService.cs:104` — `CancelPendingForUserAsync` has no production caller yet, and "does not save, does not audit" is exactly the contract the next caller gets wrong. Documented and tested, so fine to keep — just make sure #65/#66 actually adopts it instead of leaving a second inline implementation behind.

---

**Checked and fine**

`POST` over the spec's `PUT` for approve/reject — agreed, and it matches how the rest of this API already behaves. The 404-not-403 on someone else's request is right and correctly byte-identical to the unknown-id message. `QuotaRequestQuery` as a positional record binds cleanly alongside `PagedRequest` through the real pipeline (`Admin_list_sees_other_users_requests_and_honours_the_status_and_user_filters` proves it), so no #128 exposure. `ReviewedDate` comes off `TimeProvider` and is asserted against a moved clock. `EnsureValidQuota` on both write paths per D-013. Deferring #147 is right — the filtered index needs `SchemaParityTests`' regex parser and the SQLite harness to learn about filtered indexes first, which is its own change. Demo data now depicts a request the API would actually accept and the landing page's 5M "Asked for more" row stays true. Auth surface is clean: global `AuthorizeFilter`, admin routes carry the policy per action, whole 401/403 matrix tested. Only collision with the open waves is the one-line `AddRequestsServices()` registration against #152 — the conflict CONVENTIONS deliberately chose.

Fix the approval downgrade, the transaction guard, and land the review race somewhere (fix or issue), and I'm happy.

kolatts:
## Author response — all findings addressed

Rebased on `main` (#143). Gates: `dotnet build -c Release` **0 warnings**, `dotnet test` **672 passed**
(538 → 672; +134 counting the merged CLI wave and 16 new here), `dotnet format --verify-no-changes`
clean, `docs-site` builds.

---

### Major — approval could silently downgrade the subject. **Fixed.**

You were right, and right about the second half being the worse one: reading the allocation row meant a
group change *before* filing was already invisible, no race required. Both halves now read live
resolution.

New `IQuotaResolutionService.PreviewAsync(userId, ct)` runs the five-level chain and writes nothing —
no allocation upsert, no `IGatewayTierSync` call, nothing tracked. `ResolveLevelAsync` was already the
pure half of `ResolveCoreAsync`, so this is an exposure, not a second copy of the rule.

- **Submission** measures `currentQuota` against `PreviewAsync`, so the number recorded on the request
  and shown to the reviewing admin is what the chain says today.
- **Approval** re-asks and refuses with `409` when applying the stored `RequestedQuota` would not raise
  the subject: *"User 7's budget is already unlimited, so approving this request for 20,000,000 tokens
  would lower it. Their quota changed after the request was filed; reject it instead."* The finite case
  names the current number. The shared `IsIncrease(current, requested)` helper is the same predicate
  submission uses, so the two can't drift.

`409` rather than `400`: nothing about the request is malformed — the world moved under it.

Your exact scenario is now `QuotaRequestServiceTests.ApproveAsync_refuses_with_409_when_the_subject_is_now_unlimited_so_approval_would_downgrade_them`
(asserts the user is still unlimited, the row still `Pending`, no tier sync, no audit row), plus the
group variant, the endpoint variant, and — the case that would have made a naive "reject everything
stale" fix wrong — `ApproveAsync_still_succeeds_when_the_subject_moved_up_but_the_request_asks_for_more_again`.

### Minor — `LoadForReviewAsync` read-then-write. **Fixed here, not deferred.**

Taking the `ApimKeyService` route: the transition out of `Pending` is now a conditional
`ExecuteUpdateAsync` on `WHERE ... AND StatusType == Pending` inside the transaction, setting status,
reviewer, date and notes in one statement. Zero rows → re-read → `409` naming how it actually went.

One wrinkle worth flagging: `ExecuteUpdate` bypasses the change tracker, so the tracked copy is stale
the moment it succeeds. Patching it would have made the later `SaveChangesAsync` issue a second,
*unconditional* UPDATE of the same columns — undoing the claim entirely — so the entity is detached
instead and reads after that point go to the database. Tested with a second `DbContext` on the same
database deciding the row in the window, for both approve and reject; the winner's decision and notes
survive, no quota is raised behind a rejection, no audit row is written by the loser.

#147 (submit-side) stays open — the filtered index still needs `SchemaParityTests`' parser and the
SQLite harness to learn about filtered indexes, which is its own change.

### Minor — unconditional `BeginTransactionAsync`. **Fixed.**

All three writers now use `BeginTransactionIfNoneAsync` (`Database.CurrentTransaction is null ? Begin :
null`) and commit only what they began — your `ApimKeyService.cs:74` guard. Rejection got a transaction
it didn't have before, for the same reason: claim and audit row are one unit.
`SubmitAsync_joins_a_transaction_the_caller_already_owns_instead_of_opening_its_own` drives submit *and*
`CancelPendingForUserAsync` inside an outer transaction and commits it — the #65 shape, end to end.

### Minor — refusal after a side effect. **Fixed, and the side effect is gone entirely.**

Since `PreviewAsync` replaced `GetCurrentQuotaAsync`, submission no longer resolves at all: no
allocation is created, no gateway call is made, and a refused submit leaves nothing behind.
`SubmitAsync_that_is_refused_touches_neither_the_gateway_nor_the_database` seeds an
`ApimSubscriptionId` so a resolution *would* sync, and asserts `_tierSync.Calls` is empty and both
tables are.

Behaviour change worth knowing: the allocation row no longer appears on submit. It appears where it
always did — the developer's first `GET /quota/allocations/me` of the month, or a reset — and `api.md`
says so.

### Minor — missing indexes. **Fixed.**

`[Index(nameof(CreatedDate))]` and `[Index(nameof(UserId), nameof(StatusType))]`, mirrored into
`QuotaIncreaseRequests.sql`. Two rather than one composite: the default ordering applies to the admin's
unfiltered list (no user to lead with), and the filter shape applies with no ordering advantage — no
single column order serves both. Leading with `UserId` means EF drops its conventional FK index as
subsumed, so `IX_QuotaIncreaseRequests_UserId` came out of the `.sql` too — `SchemaParityTests` caught
that and now passes.

### Nit — approval scoped to `BillingPeriod.Current`, nothing expires stale requests. **Filed as #159.**

Agreed on both halves, and the downgrade guard narrows it but doesn't close it: a stale request that is
still technically an increase applies a months-old justification to today's budget. Issue lays out
expire-on-review plus a sweep in the monthly reset, and `api.md` states the current behaviour rather
than leaving it implied.

### Nit — `CancellationToken.None` past the external commit point. **Fixed here; sweep filed as #158.**

Approval's audit row, save and commit now run on `CancellationToken.None` — every refusal happens before
`ResolveAsync`, so there is nothing left to abandon after it. #158 covers `QuotaAllocationService.ResetAsync`
and the four `ApimKeyService` paths (rotation is the ugly one: the old key is dead the moment APIM
returns), with `FoundryDeploymentService.AuditAfterCommitAsync` as the shape to copy.

### Nit — seeder failing quietly. **Fixed.** `First(...)`, with a comment saying why a loud
`InvalidOperationException` on the local seed beats demo data that is one row short.

### Nit — the `*Unique` Guid alternative. **Considered, keeping the transaction.**

Approve and reject key `TargetId` on the int PK because they have it, and `AuditTargetTypes.QuotaIncreaseRequest`
documents that. Keying submit on the Guid would put one request's three audit rows in two different id
spaces under the same `TargetType`, so the audit viewer's `targetId` filter could never show a request's
full history — a worse trade than one transaction. That reasoning is now a comment at the call site
rather than only in a PR body.

### Nit — `CancelPendingForUserAsync` has no production caller. **Filed as #160.**

#156 implements it inline with a comment naming this method. #160 is the swap, scoped to
`UserLifecycleService` (take `IQuotaRequestService`, drop the inline query, keep the existing lifecycle
tests green since they assert behaviour), post-merge given the #148 → #152 → #156 order. The method
composes as-is now that it joins an ambient transaction.

---

**New follow-up issues:** #158 (CancellationToken sweep), #159 (stale requests), #160 (deprovision
adopts the seam). #147 unchanged.

**Test delta:** +16 on this branch — 4 `PreviewAsync`, 3 downgrade-guard, 2 review races, 2
ambient-transaction, 3 live-vs-stale-resolution, 2 endpoint.

Issue #151, opened 2026-09-02
Schedule the Entra syncs: nightly users sync then group sync as a timer-triggered Function


kolatts:
Discovered while implementing #41.

Both Entra syncs are admin-triggered only: `POST /users/sync` (#40) and now `POST /groups/sync-entra` (#41). Both are pull-only and idempotent — designed from the start to be safe on a schedule — but nothing schedules them, so a fork's roster drifts from the directory until somebody remembers to press the button.

Epic #10's background services cover the monthly reset (#38) and usage sync (#39); directory sync is not in either.

## Ask

Add a timer-triggered Function (the #39 shape: isolated worker, `TimeProvider`, its own audit row) that runs, in order:

1. `POST /users/sync`'s reconciliation — so people who joined the tenant have `User` rows.
2. `POST /groups/sync-entra`'s reconciliation for every linked group — which is the order that matters, because group sync skips directory members with no `User` row and reports them as `skippedUnknownUserCount`.

Details to settle:

- **Cadence** and whether it is configurable (`Entra:SyncSchedule`?). Daily overnight is the obvious default.
- **Actor.** Both services currently attribute their audit row to the calling admin via `IAuditService`. A scheduled run has no caller, so it needs `IAuditWriter.AddSystem(…)` — which means the sync services need to take the actor rather than resolve it, or grow a system-actor entry point. Same problem the monthly reset already solved; follow whatever #38 landed.
- **Off switch.** `Entra:Enabled` already gates the directory client (a disabled host returns 400); the scheduled job must no-op quietly on such a host rather than logging an error every night.
- **Departure detection is currently suspended** whenever the directory reports group-principal app-role assignments (#121). Scheduling the sync before #121 lands means the nightly run adds and updates but never deactivates — acceptable, but the job's logging should make that obvious rather than looking like a clean run.

kolatts:
The pattern this issue asks to follow now exists — from the #38/#39/#84/#119 Functions wave. For whoever picks this up:

- **Where the code goes.** `src/FoundryGate.Functions/Functions/<Name>Function.cs` is a thin `TimerTrigger` that delegates to a service in `Services/<Area>/`, so the behaviour is unit-testable without a Functions host. `MonthlyQuotaResetFunction` / `MonthlyResetJob` is the model.
[...]

Issue #150, opened 2026-09-02
Decide whether a group's Entra link should be changeable after creation (link / unlink / re-point)


kolatts:
> Parent: #223 — Product decisions pending

Discovered while implementing #30.

`CreateGroupRequest` accepts an optional `EntraGroupId`; `UpdateGroupRequest` deliberately does **not** carry one, so `PUT /groups/{id}` cannot link, unlink, or re-point a group at a different Entra group. The reasoning is on `IGroupService.UpdateAsync`: re-pointing a synced group at a different directory group would silently rewrite its entire roster on the next `sync-entra` run — an admin who mistypes a GUID would find a team's whole membership (and therefore its budget) replaced by a stranger's, with the only trace being one `group.entra-synced` audit row. Today the answer is "delete and recreate", which at least makes the destruction explicit.

That is a defensible default but not obviously the right permanent one — an admin whose directory group was itself recreated (new object id, same people) has no way to fix the link without losing the group's identity, its `GroupUnique`, and its audit history.

## Ask

Decide, and implement whichever way it goes:

- **Allow it**, with the consequence made visible: a separate `PUT /groups/{id}/entra-link` (or a `PATCH`) that takes the new `entraGroupId`, returns a *preview* of the resulting add/remove counts, and requires a confirmation flag before it writes — so "you are about to remove 14 people and add 9" is on screen before the roster changes.
- **Or keep it immutable**, and say so in `reference/api.md`'s Groups section as a rule rather than as an aside, plus add the unlink case (clearing `EntraGroupId` to convert a synced group into a native one) if that is wanted — unlinking is not destructive and could be allowed on its own.

Also worth settling in the same breath: `Group.IsEntraSynced` is currently just "`EntraGroupId` is non-empty" — if linking becomes mutable, decide whether the flag should mean something independent (e.g. "linked but sync paused").

kolatts:
Partly answered in #152's fix pass, so this issue is now narrower.

**Settled and shipped:**

[...]

Issue #149, opened 2026-09-02
Entra group sync: hoist the user map out of the per-group loop and decide SyncAllAsync's failure semantics


kolatts:
Discovered while implementing #41.

`EntraGroupSyncService.SyncAllAsync` loops the linked groups and calls the same per-group reconciliation `POST /groups/{id}/sync-entra` uses. Two things that are fine for one group and wasteful (or unhelpful) for many:

## 1. The whole `Users` table is read once per group

`SyncGroupAsync` projects every `User` into an oid → `(UserId, EntraObjectId, IsActive)` map so a large Entra group can be resolved without a `WHERE EntraObjectId IN (…)` that would blow the provider's parameter limit (the same reason `EntraUserSyncService` loads the table). With *N* linked groups that is *N* full reads of the table in one request.

**Ask:** hoist the map to the run — build it once in `SyncAllAsync` and pass it into the per-group reconciliation — while keeping `SyncAsync(groupId)` building its own. Newly-provisioned users cannot appear mid-run (nothing in this path inserts users), so one snapshot per run is correct.

## 2. One group's failure ends the run

A Graph failure (or any exception) on group 3 of 5 propagates: groups 1-2 stay committed — each group is its own unit of work — but 4-5 are never attempted and the caller gets a 500 with no summary of what did happen. Re-running is idempotent, so nothing is corrupt; it is just an unhelpful answer.

**Ask:** decide between (a) keep fail-fast, but return the partial summaries with the failure named, or (b) isolate per group — record the failure against that group and continue. Either way the response shape needs a per-group status, which is a `GroupSyncResult` change, so it is worth doing deliberately rather than by accident.

Neither is urgent: a fork has a handful of linked groups, not hundreds.

kolatts:
Claiming this on branch `claude/hardening-entra-groups-config` (hardening wave, alongside #121/#164/#123/#145).

PR #152, opened 2026-09-02
feat(groups): group CRUD, membership and Entra group sync

### LINKED ISSUE #30: Implement group CRUD: create, list, get, update, and delete
Part of #6.

All endpoints require the admin role. `POST /groups` creates a group with optional `EntraGroupId` for Entra linking. `GET /groups` returns a paged list with member count. `GET /groups/{id}` returns full detail including members. `PUT /groups/{id}` updates name, description, quota, or unlimited flag — a quota change must re-resolve allocations for all current members. `DELETE /groups/{id}` removes the group and all `GroupMember` rows (does not delete users or affect their quotas beyond group override removal).

### LINKED ISSUE #31: Implement group membership management and Entra group sync
Part of #6.

`POST /groups/{id}/members` adds a user by `userId`, records the `AddedBy` admin reference, and re-resolves quota for the new member. `DELETE /groups/{id}/members/{userId}` removes them and re-resolves. `POST /groups/sync-entra` calls Microsoft Graph `GET /groups/{entraGroupId}/members`, resolves FoundryGate users by `EntraObjectId`, adds new members, removes departed ones, and re-resolves allocations for all changed memberships. Returns a sync summary.

### LINKED ISSUE #41: Implement Entra group member sync (POST /groups/sync-entra)
Part of #11.

`POST /groups/sync-entra` requires the group to have `EntraGroupId` set; returns 400 otherwise. Call Graph `GET /groups/{entraGroupId}/members?$select=id` to retrieve current Entra member IDs. Resolve them to FoundryGate `User` records by `EntraObjectId`. Add rows to `GroupMember` for new members (with a system actor for `AddedBy`); remove rows for departed members. Re-run quota resolution for all affected users. Return `{ added, removed }`. Write a `sync.group-members` audit log entry per sync run.

kolatts:
Implements the whole `/api/v1/groups` surface: CRUD (#30), membership (#31) and Entra group sync (#41). All admin-only, all behind `Services/Groups/`.

| Method | Path | What it does |
|---|---|---|
| `GET` | `/groups` | Paged, ordered by name, `memberCount` projected; `?search=` over name + description |
| `POST` | `/groups` | `201` + `Location`; optional `entraGroupId` link; tier-validated quota |
| `GET` | `/groups/{id}` | Group + full roster |
| `PUT` | `/groups/{id}` | Name/description/quota; a quota change re-resolves every active member |
| `DELETE` | `/groups/{id}` | `204`; `409` if populated unless `?force=true`; former members re-resolved |
| `GET`/`POST` | `/groups/{id}/members` | Paged roster; add by `userId`, attributed to the calling admin |
| `DELETE` | `/groups/{id}/members/{userId}` | `204`; re-resolves |
| `POST` | `/groups/{id}/sync-entra`, `/groups/sync-entra` | Reconcile one / every linked group |

**Quota resolution now reads group state through the change tracker.** This is the one change outside `Services/Groups`, and it is the reason everything else could stay a single unit of work. `QuotaResolutionService` resolved levels 3-4 with a projection query straight to the database (`gm.Group.IsUnlimited`, `gm.Group.MonthlyTokenQuota`), which cannot see pending changes — so "edit the group's quota → re-resolve → save once", which is exactly what `PUT /groups/{id}` does, would have resolved against the policy the admin just replaced. Same for adding a member (the `GroupMember` isn't in the database yet) and deleting a group (its memberships still are). It now reads the persisted `(UserId, GroupId)` pairs, overlays this unit of work's Added/Deleted `GroupMember` entries and Deleted `Group`s, and loads the referenced groups **tracked** so EF's identity resolution hands back the instance the caller mutated. Same "pending state first, database second" rule `FindExistingAsync` and `CurrentUserAccessor.TryGetUserAsync` already follow, and four new tests in `QuotaResolutionServiceTests` pin it. User-level resolution was already correct (`ResolveAsync` loads the `User` tracked), so nothing else changes behaviour.

**Create is the one two-save path, and it is wrapped in a transaction.** `GroupId` is an `IDENTITY` value, so it does not exist until the insert runs, and the audit row's `TargetId` has to be that id rather than a placeholder the audit viewer cannot filter on. Insert → save → audit → save, inside `BeginTransactionAsync`, keeps the guarantee the single-save convention exists for. Every other mutation is a genuine single `SaveChangesAsync` with its audit row added first.

**Re-resolution covers active members only.** A deactivated developer holds no enforceable key, and `GET /quota/allocations/me` already refuses to mint them an allocation — minting one here would contradict it. Entra sync still *records* the membership for an inactive user (it is a fact about the directory), it just does not allocate.

**Directory-added memberships carry `AddedByUserId = null`.** The directory chose the membership, not the admin who pressed the button; crediting them would make the audit trail lie. The nullable column was already there for this.

**Orphan Entra members are skipped, counted and logged** — `GroupSyncResult` gains `SkippedUnknownUserCount` (positional record; the Blazor client only consumes it, so it still compiles). A non-zero count means those people have never signed in and were not imported: run `POST /users/sync` first.

**`Group.Name` gains a unique index** (mirrored in `Groups.sql` for `SchemaParityTests`). The service check is case-insensitive on both providers via `ToLower()` — SQL Server's default collation is CI and SQLite's is not, and a rule that behaves differently in the harness than in production is worse than one that cannot use the index at group scale. A `DbUpdateException` on save is translated to the same `409` only when the name really is taken; anything else rethrows.

**`EntraGroupId` is not updatable, is unique, and owns the roster.** Re-pointing a synced group at a different directory group would silently rewrite its whole roster on the next sync, so the link is set at creation only (#150 holds the "should a guarded re-link exist" question). A filtered unique index `IX_Groups_EntraGroupId` (`WHERE [EntraGroupId] <> ''` — filtered because the column is non-nullable and empty means "not linked") stops two groups backing onto one directory group. And a linked group's membership routes answer **409**: a manual add would be applied and then silently undone by the next sync, which is a worse answer than refusing. Group *policy* — name, description, quota — stays editable.

**Commit-point discipline.** Re-resolution can move a developer's APIM subscription between tier products, so every write path resolves the actor and makes every refusal *before* it, and writes its audit row and saves on `CancellationToken.None` when `QuotaResolution.TierSyncRequested` says the gateway was actually touched — an unchanged tier or an empty member list is not a commit point. `CreateAsync` joins an ambient transaction rather than opening a nested one (the `ApimKeyService` guard).

**`IsEntraSynced` is derived, not stored.** The column is gone from the entity and `Groups.sql`; the response field is computed from `EntraGroupId` in the read projection, so the two cannot drift.

**The duplicate-name 409 is collation-agnostic.** Recognised by the identifier the provider names in the violation (`IX_Groups_Name` on SQL Server, `Groups.Name` on SQLite), not by a `LOWER(Name)` re-query that an accent-insensitive database would miss. Anything else rethrows.

**A host with Entra disabled answers 503, not 400** (`FeatureNotConfiguredException`): the request is well-formed and the caller can do nothing about it. `POST /users/sync` moves with it, and it keeps "this host has no Graph" distinguishable from "this group has no Entra link", which is a genuine 400.

## Corrections to existing docs

`architecture/overview.mdx` described group sync as bulk provisioning and deprovisioning — creating `User` rows for directory members, deleting APIM subscriptions for departed ones. That is bulk *user* sync. Group sync reconciles memberships only: it never creates, deactivates or deletes a user, and never revokes a key. The pipeline is rewritten to what ships, and the "departures are hard" caution is split so it attaches to the account-level event it actually describes.

The Blazor typed client's group methods were written against a surface that never shipped (`GetGroupsAsync` returning an unpaged list); they now match the real one, plus the member roster and per-group sync.

## Verification

- `FoundryGate.Tests.Predeployment` — **665 passed, 0 failed**
- `npm run build` in `docs-site/` — 13 pages, complete
- `plans/06-groups.md` and `plans/11-entra-sync.md` verification items ticked, with an "As landed" note on each approach section

New tests: `GroupServiceTests`, `EntraGroupSyncServiceTests` and `GroupsEndpointTests` (including the 401/403 matrix over all ten routes), four pending-change cases in `QuotaResolutionServiceTests`, and filtered-index support in `SchemaParityTests` — its own docstring requires extending it in the same PR that adds a filtered index. The Entra tests run against `FakeEntraDirectoryClient` — including a 250-member group with ids duplicated across "pages" — and against `DisabledEntraDirectoryClient` for the 503. No live Graph; the live items are on #120.

## Deferred, filed rather than left inline

- #149 — `SyncAllAsync` reads the whole `Users` table once per group, and one group's directory failure ends the run
- #150 — decide whether `EntraGroupId` should be re-linkable, with a preview + confirmation
- #151 — schedule the Entra syncs (users, then groups) as a timer-triggered Function
- #163 — sweep the commit-point rule across the quota-resolution callers outside this area (`QuotaAllocationService.ResetAsync` has the same shape); should land in #118's wave rather than after it
- #120 — live-validation items added in a comment: `transitiveMembers` against the documented Graph roles (`GroupMember.ReadBasic.All` may not cover it), >100-member paging, guest/B2B members, `skippedUnknownUserCount` after a users sync

Closes #30
Closes #31
Closes #41


kolatts:
**APPROVE**

Best-argued PR in the wave. The `QuotaResolutionService` overlay is genuinely correct — I probed the two multi-group cases it has no tests for and both resolve right — and it's an O(1)-query change, not the N+1 I went hunting for. One docs invariant got missed and the commit-point ordering slipped in three of four write paths; neither is deep.

---

**Issues**

- **Major** `docs-site/src/content/docs/getting-started/why-foundrygate.mdx` lines 50, 65, 198 — You corrected `overview.mdx` to say group sync never creates a `User` or issues a key, but the *simplest* level still says "when a developer joins a FoundryGate-linked Entra group, they're automatically provisioned: user record created, key issued... when they leave, the key is revoked" and "Group sync → key + quota". CLAUDE.md's docs invariant #2 wants every behavioral change at ALL levels, and right now two levels of the progressive docs flatly contradict each other.

- **Minor** `src/FoundryGate.Api/Services/Groups/GroupService.cs` lines 177/181, 234/236, 319/321 and `EntraGroupSyncService.cs` lines 123/136 — `audit.LogAsync` resolves the actor internally (`GetRequiredUserAsync` → 403 for an unprovisioned admin), so in `UpdateAsync`, `DeleteAsync` and `RemoveMemberAsync` that refusal lands *after* `ResolveManyAsync` has already moved APIM. CONVENTIONS: "resolve the actor and do every refusal *before* the call." `AddMemberAsync` line 278 gets it right, which is what makes it look accidental rather than considered.

- **Minor** same lines — and the other half of that rule: the audit row and `SaveChangesAsync` after the tier move run on the request's `CancellationToken` rather than `CancellationToken.None`, so a client that hangs up mid-`PUT /groups/{id}` moves a whole team's APIM products with nothing in the database to show for it. `NullGatewayTierSync` makes this latent today; #156 lands the real one. `QuotaAllocationService.ResetAsync` has the same shape, so this may be #118's to own — but it wants an issue either way.

- **Minor** `src/FoundryGate.Api/Services/Groups/GroupService.cs` line 83 — `CreateAsync` calls `BeginTransactionAsync` unconditionally where `ApimKeyService` line 74 uses `CurrentTransaction is null ? … : null`. I confirmed with a throwaway test that an ambient transaction turns this into an `InvalidOperationException`, and the `CommitAsync` would commit somebody else's transaction if it didn't. Nothing composes it today; #156 lands orchestrators that do.

- **Nit** `src/FoundryGate.Api/Services/Groups/GroupService.cs` line 404 — `SaveWithNameConflictAsync` claims the 409 only when `LOWER(Name)` matches, but the index that raised the error is collation-based; on an accent-insensitive database "Résumé" vs "Resume" trips `IX_Groups_Name` and then rethrows as a raw 500. Worth knowing too: under SQLite the index is case-sensitive, so no test ever reaches this catch block.

- **Nit** `src/FoundryGate.Data/Entities/Group.cs` line 39 — nothing reads `IsEntraSynced`; `SyncAllAsync` filters on `EntraGroupId != string.Empty` instead, so it's derived state that can only drift. #150 already asks the question — worth answering there rather than keeping two sources of truth.

- **Nit** `src/FoundryGate.Api/Services/Groups/EntraGroupSyncService.cs` line 69 — two groups can link to the same `entraGroupId` (verified; no index, no check), giving their members `GroupMax` of both quotas off one directory group. Deterministic, so not a bug — but probably not what an admin who double-pasted a GUID meant.

- **Nit** `src/FoundryGate.Api/Services/Groups/EntraGroupSyncService.cs` line 61 — "Entra is disabled on this host" comes back as a 400 next to "this group has no link", which is a real caller error. CONVENTIONS now has `FeatureNotConfiguredException` → 503 for exactly this and `FoundryDeploymentService` uses it; `DisabledEntraDirectoryClient` predates the rule so it isn't yours, but this PR pins the 400 into two more endpoints, the docs and a test.

---

**Things I went looking for and didn't find**

- `LoadGroupPoliciesAsync` is two set-based queries regardless of user count, and `ChangeTracker.Entries<T>()` runs once per resolution, not per user — `ResetAsync` over every active user is still O(1) queries. EF Core 10 translates the `Contains` lists to `OPENJSON`, so no parameter-limit cliff either.
- The overlay is right for the cases you didn't test: deleting one of two groups leaves the other's policy standing, and editing one of two takes the max including the pending edit. Both verified against a real SQLite context.
- `GroupQuery` as an attribute-free positional `[FromQuery]` record is fine — #128 is about validation attributes on request *bodies*, and `AuditLogQuery` is the same shape.
- `Groups.sql` matches the entity (SchemaParity green); `TestDataSeeder` no-ops on existing users, so the new unique index doesn't break re-seeding; the Web client compiles against the real routes.
- 401/403 matrix covers all ten routes, `Location` is lowercase per #129 and actually resolves, 537/537 predeployment green, build clean at 0 warnings, `dotnet format` clean.

---

**Merge order** — #148, then this, then #156.

#148 has the smallest shared surface: one DI line, its own `reference/api.md` section, and `TestDataSeeder` which nobody else touches. This one next — #156 doesn't touch `QuotaResolutionService` at all, so the overlay lands clean, and #156 then rebases onto the corrected `overview.mdx` it's meant to build on (its "the subscription deletion and hard stop land with the user lifecycle work" aside is literally handing itself the baton). #156 last because it owns the widest blast radius — `GlobalExceptionHandler`, `ApiTestFactory`, `QuotaAllocationService`, and the real `ApimGatewayTierSync` that turns the commit-point Minors above from latent into live. Expect trivial conflicts in `ApiServiceCollectionExtensions.cs` (adjacent registration lines, three ways) and `reference/api.md` (different sections); the only real one is `overview.mdx` between this and #156.

kolatts:
Fix pass pushed as `4244217`. All eight taken; point by point.

---

**Major — `why-foundrygate.mdx` contradicts the corrected `overview.mdx`.** Fixed, and you were right that it was the *simplest* level that mattered most — it was the one making the flatly false promise. Rewrote all three places in the plain register, and the fix was to separate the two syncs rather than to soften the wording:

- The card is now "Entra-synced access": being given access to FoundryGate in Entra is what creates the account and issues the key, losing it is what revokes them — and team groups are called out as a *separate* sync that moves a person's budget without touching their account.
- The comparison table splits the one row into two: "Assigned in Entra → account + key" and a new "Team budgets from the directory | ❌ Nothing to sync | ✅ Entra group membership → that team's budget tier". The deprovisioning row now says "Access removed in Entra", not "Departure detected", because leaving a *team* group detects nothing.
- The "Entra sync" prose is now two bullets under one sentence — "who has an account" and "which team's budget they get" — ending with the explicit line that moving somebody between teams "only changes the size of their budget — it never creates or closes their account, and never touches their key."

---

**Minor — actor resolved after the side effect.** Fixed in all five paths (`UpdateAsync`, `DeleteAsync`, `RemoveMemberAsync`, `SyncAsync`, `SyncAllAsync`). It *was* accidental: `AddMemberAsync` needed the actor's `UserId` for `AddedByUserId`, so it happened to do the right thing, and the others let `audit.LogAsync` resolve it implicitly. There is now an explicit `EnsureActorAsync` first line, and both sync entry points resolve before the directory read — an unprovisioned admin should not even cost a Graph call. `Write_paths_refuse_an_unprovisioned_caller_before_re_resolving_anything` pins all four group paths at once and asserts `_tierSync.Calls` is empty and the allocation unmoved; `SyncAsync_refuses_before_reading_the_directory_when_the_caller_has_no_user_row` does the sync side.

---

**Minor — commit token after the tier move.** Fixed, and the signal is `QuotaResolution.TierSyncRequested` rather than "did we call resolution": an empty member list, or a re-resolve where nobody's tier actually changed, is not a commit point and shouldn't opt out of cancellation. So each path computes `gatewayMoved` from the resolutions and passes `CommitToken(gatewayMoved, cancellationToken)` to both the audit call and `SaveChangesAsync`.

You were right that this is bigger than my area — `QuotaAllocationService.ResetAsync` has exactly the same shape and `GetMyAllocationAsync` a smaller version of it. Filed **#163** for the sweep, with the suggestion that the "audit + save on `None`" tail belongs to the resolution seam itself rather than to each caller, since resolution already reports its own commit point. Flagged there that it should land in #118's wave, not after it.

---

**Minor — unconditional `BeginTransactionAsync`.** Fixed with the `ApimKeyService` guard (`CurrentTransaction is null ? Begin… : null`), and the commit is now conditional on having opened one, so we can't commit somebody else's. `CreateAsync_joins_an_ambient_transaction_instead_of_opening_a_nested_one` runs it under an outer transaction and asserts the outer is still ours to commit afterwards.

---

**Nit — collation-dependent 409.** Good catch, and the fix went the way you suggested rather than fixing the re-query: detection is now by the identifier the provider names in the violation (`IX_Groups_Name` on SQL Server, `Groups.Name` on SQLite — both identifiers, so a localized server message doesn't break it), with anything else rethrown untouched. That also closes a hole the re-query had: a loser whose winner then rolled back would have re-queried clean and thrown a 500.

On "no test ever reaches this catch block" — there is one now, and it doesn't need a second connection. Adding a conflicting `Group` to the change tracker without saving produces exactly the situation the index exists for: the pre-check's query can't see it, both rows insert in one `SaveChanges`, the index rejects the pair. `CreateAsync_maps_a_unique_index_violation_the_pre_check_could_not_see_to_409` and its negative twin on `IX_Groups_GroupUnique` (which must stay a `DbUpdateException`, not become a misleading 409) both go through the real catch.

---

**Nit — `IsEntraSynced` is derived state.** Agreed, and rather than answering it on #150 I deleted it: the column is gone from the entity and `Groups.sql`, and `GroupResponse.IsEntraSynced` (unchanged on the wire) is now `group.EntraGroupId != string.Empty` in the read projection. Left a note on the entity that a future "linked but sync paused" needs its own honestly-named column rather than a re-purposed derived flag. #150 is updated and re-titled to just the link-mutability question that's actually still open.

---

**Nit — two groups can link to one `entraGroupId`.** Fixed rather than documented: filtered unique index `IX_Groups_EntraGroupId` (`WHERE [EntraGroupId] <> ''` — filtered because the column is non-nullable and empty means "not linked", so an unfiltered unique index would allow exactly one native group in the whole fork), plus a friendly pre-check and the same identifier-based 409 on the race.

That meant honouring `SchemaParityTests`' own rule about extending it in the same PR: it now parses an optional `WHERE` clause off the index DDL and compares it to `index.GetFilter()`, normalising whitespace and balanced outer parens so DacFx's `([EntraGroupId]<>'')` and the model's `[EntraGroupId] <> ''` agree. Text-level, not semantic — documented as such, alongside moving "filtered indexes" out of the deliberate-limits list. The green run is real: had the regex not captured the filter, the model's non-empty filter would have compared against `""` and failed.

---

**Nit — 400 for a disabled host.** Taken; you're right that it pins the wrong answer into two more endpoints. `DisabledEntraDirectoryClient` now throws `FeatureNotConfiguredException` → **503**, which also keeps "this host has no Graph" distinguishable from "this group has no Entra link" — a genuine 400 that sits right next to it. Updated `DisabledEntraDirectoryClientTests`, `UsersSyncEndpointTests` (so `POST /users/sync` moves too), the `[ProducesResponseType]` sets on both controllers, `reference/api.md`'s Users and Groups sections, and the fork guide's troubleshooting entry, which now names both endpoints and says why it's a 503.

---

**Your "below threshold" observation — manual membership on a linked group.** Agreed it's a trap, and it's now a 409 on both `POST /groups/{id}/members` and `DELETE /groups/{id}/members/{userId}` when the group has an `entraGroupId`, naming the directory group and the sync endpoint to use instead. Group *policy* stays editable — the directory owns who is in the group, not what the group is worth. `EntraGroupSyncService` writes `GroupMember` rows directly, so it is unaffected. Documented in `reference/api.md` and covered at both service and endpoint level.

---

**Merged `origin/main` after #148 landed** (`8cf5ac4`). Three conflicts, all additive on both sides: the adjacent DI registration lines; `LoadGroupPoliciesAsync` next to the new `PreviewAsync`; and their two test blocks. Kept everything. `PreviewAsync` deliberately keeps its straight `AsNoTracking` projection rather than routing through the tracker-aware loader — it is a pure read with nothing pending in the unit of work, and staying untracked is the point of it — with a note in place pointing at #163 for the day a caller previews *after* editing a group in the same request.

**Verification after the pass and the merge:** build 0 warnings / 0 errors; predeployment **743 passed, 0 failed**; `dotnet format --verify-no-changes` clean; docs build 14 pages.

**Merge order:** understood that it's now #148 → #156 → #152, with me last so #156's `MoveToProductAsync` change lands first — which matters here, because a group quota change fans the tier sync out across every member and each one committing its own unit of work would be exactly the half-finished state this PR's commit-point work is trying to avoid. #148 is already merged in. I'll take #156 on your signal; the likely conflicts there are `overview.mdx` (its aside hands itself the baton from mine), `reference/api.md`, and `ApiServiceCollectionExtensions.cs` again.

Issue #154, opened 2026-09-02
GET /users/me: absorb the first-login race instead of returning 409


kolatts:
Discovered while implementing #28 / #65.

`GET /users/me` auto-provisions on first login. Two requests from the same person arriving together (two browser tabs, a UI that fires the profile call twice on load) both find no `User` row and both run `IUserLifecycleService.ProvisionAsync(FirstLogin, …)`. `Users.EntraObjectId` is unique, so the loser's `SaveChangesAsync` fails; `UserLifecycleService.CreateFromFirstLoginAsync` catches the `DbUpdateException`, detaches its entity and throws a `ConflictException` — **the caller gets a `409` telling them to retry.**

That is correct and safe (the winner's row, allocation and APIM subscription are intact; the loser's transaction rolled back and left no orphan), but it is a poor first impression: the very first request a new developer's UI makes can fail, and only a retry fixes it.

## Ask

Make the race invisible to the caller: on the `ConflictException` from a first-login provision, re-read the row (`ICurrentUserAccessor.TryGetUserAsync` after clearing the failed entity) and return the winner's profile instead of surfacing a 409. Points to be careful about:

- The failed save happened inside the pipeline's transaction, which is already rolled back — the retry needs a clean change tracker, not just a second query.
- Only the unique-index violation on `EntraObjectId` should be swallowed; any other `DbUpdateException` must still surface.
- Keep the 409 path reachable for the genuinely-repeated case (`ProvisionAsync(FirstLogin)` called for an oid that already has a row) — that one is a programming error, not a race.
- Test it by forcing the conflict deterministically (seed the row between `TryGetUserAsync` and the save, e.g. via a second context) rather than by racing two real requests.

kolatts:
Note from the #156 fix pass, for whoever picks this up: the losing tab does not fail fast. The provision pipeline holds the transaction — and therefore the X lock on the new `Users` row — across the ARM round trip, so the second request blocks on the unique index for the whole APIM call before it gets its 409. A naive retry pays that latency twice. Worth measuring before choosing between (a) absorbing the conflict and re-reading, and (b) shortening the lock window by claiming the oid in a short transaction of its own before the ARM call.

kolatts:
Claimed by an implementation agent — working this together with the rest of the keys/lifecycle/quota hardening wave (#163, #158, #168, #154, #136, #167) on branch `claude/hardening-keys-lifecycle-quota`.

Issue #153, opened 2026-09-02
Expose the gateway model alias map to the control plane so GET /users/me can return it


kolatts:
Discovered while implementing #28.

`GET /users/me` returns a `GatewayConnectionInfo` (`cliConfig`) so a developer can configure Claude Code / Codex without reading the docs. Two of its four fields are not yet sourced from anything the control plane knows:

- **`modelAliases` is always empty.** The alias map (`alias -> { deployment, pool }`) lives only in `infra/modules/ai-gateway.bicep`'s `modelAliases` parameter; nothing surfaces it to the API. `UserService.BuildCliConfig` therefore returns `[]` rather than guessing, and developers get model names from `getting-started/cli-setup.mdx` instead.
- **`anthropicBasePath` / `openAiBasePath` are C# constants** (`GatewayOptions.AnthropicBasePath` = `/anthropic`, `GatewayOptions.OpenAiBasePath` = `/openai/v1`), mirroring the bicep's hard-coded `anthropicApiPath` / `openaiApiPath`. They are constants on purpose — a setting the data plane does not honour would be worse — but the two definitions can drift.

## Ask

- Decide how the alias map reaches the control plane: a `Gateway__ModelAliases__*` env var set by `infra/modules/control-plane.bicep` from the same `modelAliases` object the gateway module consumes (one source, two consumers), or an APIM Management read of the named value the policy uses. The env var is simpler and needs no ARM call on a profile read.
- Populate `GatewayConnectionInfo.ModelAliases` from it (alias, deployment name, provider), filtered to what the caller's tier product actually allows if that ever differs per product (plans/25).
- If the front-door paths ever become bicep parameters, promote them to `Gateway__*` settings in the same change and add a test that the shipped `appsettings.json` matches `infra/main.bicep` (the precedent is `GatewayOptionsTiersTests`).
- Update `reference/api.md`'s `GET /users/me` note, which currently says the alias list is empty.

kolatts:
Claiming this as part of a backend sweep (with #194, #180, #181, #130) on `claude/backend-sweep-b`.

PR #156, opened 2026-09-02
feat(api): user lifecycle pipelines, the /users surface, and the real gateway tier sync

### LINKED ISSUE #28: Implement GET /users/me with first-login auto-provisioning flow
Part of #5.

On `GET /users/me`, extract the caller's Entra `oid` claim. Look up by `User.EntraObjectId`; if not found call Microsoft Graph `GET /users/{oid}` to fetch `displayName`, `mail`, and `employeeId`, then INSERT the user with the system default quota, provision an APIM key (see #9), and INSERT a `QuotaAllocation` for the current period. If found, update display fields and `LastSyncedAt`. Return a `UserProfileDto` that includes quota info and masked key status.

### LINKED ISSUE #29: Implement admin user management endpoints: list, detail, quota, activate/deactivate
Part of #5.

Implement the remaining `/users` endpoints behind `[Authorize(Policy = AdminPolicy)]`: `GET /users` returns a paged `UserDto` list with optional search. `GET /users/{id}` returns full detail. `PUT /users/{id}/quota` sets `MonthlyTokenQuota` or `IsUnlimited` and re-runs quota resolution. `PUT /users/{id}/activate` re-enables the user; `PUT /users/{id}/deactivate` sets `IsActive = false`, revokes the APIM key, and hard-stops the current allocation. Write audit log entries for quota changes and activate/deactivate actions.

### LINKED ISSUE #65: Wire IUserLifecycleService as single pipeline orchestrator; fix Entra departure deprovision

### LINKED ISSUE #66: Implement re-activation with full APIM re-provision and orphan subscription detection

### LINKED ISSUE #118: Implement ApimGatewayTierSync: move the APIM subscription to the user's resolved tier product

### LINKED ISSUE #160: Deprovision should call IQuotaRequestService.CancelPendingForUserAsync instead of cancelling pending requests inline
Two waves landed the same behaviour in parallel, and the duplicate is deliberate but should not survive
both merges.

`UserLifecycleService.DeprovisionAsync` (#156, closing #65/#66) rejects a departing developer's pending
`QuotaIncreaseRequest`s inline — `StatusType = Rejected`, `ReviewNotes = "User deactivated"`, no
reviewer — with a comment naming the replacement, because the requests wave was in flight at the time.

`IQuotaRequestService.CancelPendingForUserAsync(int userId, string note, CancellationToken ct)` (#148,
closing #34/#35) is that replacement and was built for exactly this caller:

- neither saves nor audits — the cancellations land in the orchestrator's own unit of work and are
  described by its `user.deactivated` audit row;
- joins an ambient transaction rather than opening one, so it composes inside the deprovision
  transaction as-is;
- idempotent, returns the number of requests it closed (useful in the deprovision audit details);
- covered by service tests including "only that user's rows", "decided rows untouched", "persists
  nothing until the caller saves" and the ambient-transaction case.

**Scope:** in `Services/Lifecycle/UserLifecycleService`, take `IQuotaRequestService` on the constructor,
replace the inline query-and-mutate with a `CancelPendingForUserAsync` call, drop the now-stale comment,
and keep the existing lifecycle tests (`deactivation ... rejects only that user's pending requests`)
green — they assert behaviour, not implementation, so they should pass unchanged. Consider adding the
returned count to the `user.deactivated` audit details.

Both PRs merge before this can be done (merge order #148 → #152 → #156), so this is a small follow-up on
`main`.

kolatts:
Plan 21's provision and deprovision pipelines land as a single orchestrator, the `/api/v1/users` surface is built on top of it, and the `IGatewayTierSync` seam gets its real implementation.

> Updated after review. The three Majors are fixed and pinned; see the [point-by-point response](https://github.com/kolatts/foundry-gate/pull/156#issuecomment-5505384460). The design notes below describe the shipped shape, not the first draft.

## What's here

**`Services/Lifecycle/IUserLifecycleService` (#65, #66)** — the one place the sequence lives:

```csharp
Task<User> ProvisionAsync(ProvisionTrigger trigger, ProvisionContext context, CancellationToken ct);
Task<ApiKeyRevealResponse> ProvisionKeyForUserAsync(int userId, CancellationToken ct);
Task DeprovisionAsync(DeprovisionTrigger trigger, int userId, CancellationToken ct);
```

`ProvisionTrigger` is `FirstLogin | AdminProvision | Reactivate`; `DeprovisionTrigger` is
`AdminDeactivation | EntraDeparture`; `ProvisionContext` is `FirstLogin()` or `ForUser(id)`.

- **Provision**: `[insert User from claims, enriched via Graph when Entra:Enabled]` → `IQuotaResolutionService.ResolveAsync(current period)` → `IApimKeyService.ProvisionAsync(user, resolution tier)` → audit (`user.provisioned` / `user.activated`) → save.
- **Deprovision**: `RevokeAsync` / `RevokeAsSystemAsync` → `IsActive = false` → current allocation `IsHardStopped = true` → pending `QuotaIncreaseRequest`s closed via #148's `IQuotaRequestService.CancelPendingForUserAsync` → audit (`user.deactivated`) → save.
- `EntraUserSyncService`'s departure path calls it (Trigger B, system-attributed audit rows), keeping the group-assignment suspension from #121 intact.

**`Services/Users` + `UsersController` (#28, #29)** — `GET /users/me` (auto-provisions on first call, refreshes display fields from the token afterwards, returns quota + masked key + `cliConfig`), `GET /users` (paged, `?search=` on name/email, `?isActive=`), `GET /users/{id}` (groups + allocation + masked key), `PUT /users/{id}/quota`, `POST /users/{id}/activate`, `POST /users/{id}/deactivate`. `POST /users/sync` keeps its route and gains `failedCount`.

**`Services/Quota/ApimGatewayTierSync` (#118)** — delegates to `IApimKeyService.MoveToProductAsync`, registered in place of `NullGatewayTierSync` when `GatewayOptions.IsApimConfigured`, and translates ARM failures to 502. The keys service already audits `key.tier-changed`, so this writes no audit row of its own.

## Design decisions worth a reviewer's attention

**The two pipelines are shaped differently, because their irreversible step points the other way.**
*Provision* holds one transaction across the APIM create: `IApimKeyService` joins it rather than opening its own, so an ARM failure rolls back the claim, the `User` row and the allocation together — "a failed first login leaves no `User` row" is a test, not a hope. Its residue is the safe direction: a subscription minted but not committed becomes an orphan the next provision adopts by name.
*Deprovision* cannot do that — an APIM `DELETE` has no undo, and rolling back after it would leave a deleted key the database knows nothing about. So it deletes **first, outside** any transaction it owns, then takes the database steps in a transaction of their own. The residue there is one-directional and recoverable (key revoked and audited, user still active); re-running the deactivation completes it, because revocation tolerates a missing subscription. Both directions are now a compensation table in `plans/21`.

**Everything after an accepted external change runs on `CancellationToken.None`**, with an Error log naming what needs reconciling — the `FoundryDeploymentService.AuditAfterCommitAsync` pattern. Fixed at the source in `ApimKeyService.ProvisionAsync`/`RevokeCoreAsync` (so `DELETE /keys/{userId}` benefits too) and in every caller that can reach the gateway through quota resolution. The token switches to `None` only when something actually reached APIM, so a request that touched nothing external stays cancellable.

**`MoveToProductAsync` does not save.** It runs inside quota resolution, mid-way through its caller's unit of work, so a save there committed the caller's half-finished mutation — a quota in the database with no audit row describing it. The `key.tier-changed` row now joins the caller's change tracker; every call site saves exactly once (verified across the users, quota, requests and groups waves).

**Each Entra departure is its own unit of work.** N serial ARM deletes in one transaction meant a 502 on the last one erased the record of every earlier deletion. Failures are counted in `UserSyncResult.FailedCount`, logged, and retried next run; the Graph enumeration happens before any transaction exists.

**New `UpstreamDependencyException` → 502**, distinct from `FeatureNotConfiguredException`'s 503 ("never wired up" vs. "wired up and failing"), selected by an **allowlist** of ARM/transport types so our own `InvalidOperationException`s stay the 500s they deserve.

**Re-activation is a provision *trigger*, not a `ReactivateAsync` method.** It is the same sequence with `IsActive = true` in front of it. It also clears the `IsHardStopped` its own deactivation set, and keeps a key the user still holds rather than re-minting one a configured CLI is already using.

**`POST /users/{id}/activate` does not return the new key**, and **`POST /keys/{userId}/provision` no longer takes `?tier=`.** A budget *is* a tier, so the product comes from the user's resolved quota and nothing else; to mint on a different tier you set the quota, which moves the gateway too. That entry point moved to the orchestrator (`ProvisionKeyForUserAsync`) because the key service cannot depend on resolution — resolution already depends on it through the tier sync.

**`DELETE /keys/{userId}` is deliberately *not* routed through deprovision** (#116). `DeprovisionTrigger` has no member for it: revoking a key should not hard-stop someone's quota or reject their pending requests.

**`cliConfig` is sourced honestly.** `gatewayBaseUrl` comes from the new `GatewayOptions.ApimGatewayUrl` (infra already sets `Gateway__ApimGatewayUrl`; empty locally rather than invented). The two front-door paths are constants mirroring `ai-gateway.bicep`. `modelAliases` is **empty**: the alias map lives only in Bicep, and guessing would be a lie (#153).

**`GET /users/{id}` defines its own `UserGroupMembershipResponse`** rather than reusing `Groups.Contracts.GroupResponse` — the user view needs identity and join date, not the group's quota policy and member count.

## Verification

- `dotnet build FoundryGate.sln -c Release` — succeeded, **zero warnings**
- `dotnet test src/FoundryGate.Tests.Predeployment -c Release` — **799 passed**, 0 failed (475 on main when this wave started)
- `npm ci && npm run build` in `docs-site/` — 14 pages built
- Merged `origin/main` after #143/#144, #148 and #152 — all clean

Tests run against real services with only the two Azure seams faked. Highlights:

- **Commit-point probes** — `FakeApimManagementClient.AfterMutation` cancels the request token the instant APIM accepts a create or a delete; both pipelines still commit fully and audit fully.
- **`A_quota_change_whose_audit_row_cannot_be_written_is_not_committed`** — gateway moved, database unchanged, no orphan `key.tier-changed` row.
- **`A_deactivation_whose_database_half_fails_leaves_a_recoverable_state_not_a_silent_one`** — the documented residue, plus the retry that clears it.
- **`One_departure_the_gateway_refuses_is_counted_and_does_not_undo_the_others`** — per-user isolation and `failedCount`.
- Plus first-login atomicity, orphan adoption on re-activation, the full auth matrix, paging/search, tier validation, and the deactivate→reactivate round trip.

Plan verification checklists in `plans/05-users.md` and `plans/21-provision-deprovision.md` are ticked, with plan 21's implementation notes and both compensation tables amended to the shipped shape.

## Deferred (issues filed, nothing left as a TODO)

- **#153** — expose the gateway's model alias map so `cliConfig.modelAliases` stops being empty.
- **#154** — absorb the first-login race instead of returning 409 (commented with the lock-window measurement it needs first).
- **#167** — `User.LastLoginDate`, so "has this account been used?" has an honest column now that `/users/me` no longer overloads `LastSyncedDate`.
- **#168** — `ApimKeyService.RotateAsync` has the same cancellation gap, and its compensation is guarded by `is not OperationCanceledException`, so a cancellation skips it.
- **#132** — live checklist extended with the lifecycle items a fake cannot prove, headed by the one #118 flags: *does an APIM re-scope preserve the primary key?* `api.md` now carries a caution saying that is unproven.
- **#122** — commented: the departure trigger is live, so an `accountEnabled = false` account keeps a working gateway key for the whole disable-to-delete window. Owner decision still needed.

Closes #28
Closes #29
Closes #65
Closes #66
Closes #118
Closes #160


kolatts:
**REQUEST CHANGES**

This is the best-argued PR in the repo so far — the "APIM inside the transaction" call is the right one, and the write-up is honest enough that I could check every claim. But the transaction only buys atomicity in the direction that was already safe, and both pipelines hand `cancellationToken` to the save *after* APIM has accepted the change, which is the one thing CONVENTIONS.md explicitly tells you not to do. Two provable data-integrity bugs; the rest is small.

---

**Issues**

- **Major** `Services/Lifecycle/UserLifecycleService.cs:216,220` (and `123,127`) — The audit row, `SaveChangesAsync` and `CommitAsync` all run on the request's `cancellationToken` after `RevokeKeyAsync` has already made APIM delete the subscription. CONVENTIONS.md: *"once the external system has accepted the change, the audit row and `SaveChangesAsync` run on `CancellationToken.None` — a client that disconnects mid-call must not turn an accepted change into an unaudited one."* I wrote a throwaway probe that cancels the token the instant the fake APIM returns from the delete: the subscription is gone, the row still says `IsActive = true` with a stale key ciphertext, and there are **zero** audit rows — not `user.deactivated`, not `key.revoked`. `FoundryDeploymentService.AuditAfterCommitAsync` is the precedent and it's twelve lines, including the Error log for the save-failed-anyway case that's also missing here.

- **Major** `Services/Quota/ApimGatewayTierSync.cs` → `Services/Keys/ApimKeyService.cs:MoveToProductAsync` — `MoveToProductAsync` ends with its own `SaveChangesAsync`, and `QuotaResolutionService.ResolveCoreAsync` calls the sync *after* mutating the allocation, so making this seam real turned it into a service that commits its caller's half-finished unit of work. Second probe, `PUT /users/{id}/quota` with the audit write forced to fail: the gateway is on `power`, `MonthlyTokenQuota = 20M` is **committed** to the database, and there is no `user.quota-changed` row describing it. `UpdateQuotaAsync` opens no transaction, so the "audit before your own save, commits atomically" rule can't hold. Same shape in `QuotaAllocationService.ResetAsync`, which also opens no transaction and whose docs still promise "in one transaction" — each tier sync now flushes every allocation Added so far, so a mid-loop ARM failure leaves the reset half-applied with rows re-resolved but never stamped. Once #152 lands this fires on every group edit too.

- **Major** `Services/Lifecycle/UserLifecycleService.cs:162-220`, `Services/Entra/EntraUserSyncService.cs:50-157` — The transaction argument works for provision, where the residue is a harmless orphan the next provision adopts by name. Deprovision is the other direction: an APIM delete can't be rolled back, so *any* rollback after it — commit failure, deadlock, cancellation, or a later user's failure in the sync loop — leaves the key deleted and the database completely unaware. The sync makes it worse by spanning N serial ARM DELETEs in one transaction: a 502 on the last departure undoes the DB record of all of them. It self-heals on the next sync run, which is real mitigation, but the PR body's "the one residue this leaves is the *safe* direction" is only true of provision. Worth considering whether deprovision should commit the row state first and delete after, or just drop to per-user transactions.

- **Minor** `Controllers/UsersController.cs` (quota + activate) — `UpstreamDependencyException` is only ever thrown from `UserLifecycleService`, and the wrap there covers `keys.ProvisionAsync`/`Revoke*`, not `quotaResolution.ResolveAsync`. So the tier-move path — `UpdateQuotaAsync` → `ResolveAsync` → `ApimGatewayTierSync` → `MoveToProductAsync` → `ArmApimManagementClient`, which only translates 404 — lets a real ARM failure (403 from a missing role, a 429, a 500) escape as a raw `RequestFailedException` and fall through the handler as a generic **500**. Both `PUT /users/{id}/quota` and `POST /users/{id}/activate` declare `ProducesResponseType(502)` and `api.md` promises the same. And per the issue above, "nothing was saved" isn't true either.

- **Minor** `Services/Entra/EntraUserSyncService.cs:50` — The transaction opens before `directory.ListAssignedUsersAsync` on line 60, so a paged Graph enumeration of the whole org runs inside the lock window for no reason. Moving `BeginTransactionAsync` below line 60 costs nothing and cuts the longest read out of it.

- **Minor** `Services/Users/UserService.cs:83-84,249` — `RefreshFromClaims` stamps `LastSyncedDate` unconditionally and saves, so every `GET /users/me` is a row UPDATE on `Users` — the same table the sync's long transaction holds write locks on. `UserContracts.cs` still documents that column as *"When this user was last touched by an Entra sync"*, which it no longer means.

- **Minor** `Services/Lifecycle/LifecycleTriggers.cs:20` — `ProvisionTrigger.AdminProvision` has zero production callers; only tests reach `LoadForAdminProvisionAsync`. Meanwhile `POST /keys/{userId}/provision` still takes `?tier=` defaulting to `standard`, so an admin minting a key for a `power` user silently under-scopes them at the gateway — and `KeysController`'s own doc comment says *"#118 … after which the user's resolved tier drives the product and this parameter becomes an override at most; the lifecycle orchestrator (epic #64) then owns the provision call itself."* This PR closes #118 and builds the orchestrator. Either wire it up or drop the enum member.

- **Minor** `Services/Lifecycle/UserLifecycleService.cs:179` — Reactivation never clears the `IsHardStopped` its own deactivation set, and nothing else does until the next fork-wide `POST /quota/reset`. `isHardStopped` is on the wire and the docs call it the dashboard mirror, so a deactivate-by-mistake leaves that developer showing hard-stopped for the rest of the month with no per-user remedy.

- **Minor** `Services/Users/UserService.cs:116-128` — `GET /users` orders by `DisplayName` and filters on `DisplayName`/`Email`/`IsActive`; `Users.sql` carries only the two unique indexes. CONVENTIONS.md: *"Index the columns your default ordering and filters use … and mirror the index in `FoundryGate.Database/dbo/Tables/*.sql`."* Following `ListCurrentPeriodAsync`'s precedent is following another violation. Fine to defer, but per CLAUDE.md it needs an issue rather than a PR-body paragraph.

- **Minor** `docs-site/src/content/docs/reference/api.md:59` — "their key is unchanged, so nothing needs reconfiguring" is stated as fact, and it's exactly the assumption #118/#132 flag as unproven against real ARM. `ApimGatewayTierSyncTests.Moves_the_subscription_to_the_tier_product_without_changing_the_key` proves the fake preserves the key, not that APIM does. The PR body is honest about this; the public docs should hedge the same way.

- **Minor** `Configuration/GatewayOptions.cs:67` — The seeded `SystemConfiguration` row `ApimGatewayUrl` is documented as *"APIM gateway base URL shown to developers on `/me`"* and `GatewayConnectionInfo.GatewayBaseUrl`'s param doc points at it; nothing reads it now. Sourcing it from infra's env var is the right call, but that leaves an admin-editable row on `/config` that lies about what it does.

- **Nit** `Services/Lifecycle/UserLifecycleService.cs:352` — `IsUpstreamFailure` is a denylist, so anything not on it becomes a 502. `ApimKeyService`'s own `InvalidOperationException("The user must be saved…")` and `"APIM returned an empty primary key"` get reported to the caller as the gateway's fault. An allowlist of the ARM/APIM exception types blames the right party.

- **Nit** `Services/Lifecycle/UserLifecycleService.cs:69` — Worth noting on #154: because the transaction holds the X lock on the new `Users` row across the ARM call, the losing tab doesn't fail fast — it blocks on the unique index for the whole ARM round trip and *then* gets its 409. The retry #154 adds pays that latency twice.

---

Things I went looking for and didn't find a problem with: no admin response carries anyone else's key material, the audit `details` objects are clean, the 502/503 split is a genuinely good taxonomy, `EnsureValidQuota` runs before anything is written, the 409 semantics are right, `UserListQuery` matches the `AuditLogQuery` precedent, the DI is one line per area, and the plan/docs amendments are more honest than most. Two methods instead of three was the right call, and so was keeping `DELETE /keys/{userId}` out of the pipeline.

kolatts:
Fix pass pushed (`1c87ebb`, `7aae1ca`, on top of merges of #148 and #152). All three Majors are fixed and pinned, all seven Minors and both Nits are addressed. Point by point.

---

### Major 1 — commit point on the request token

Fixed at the source rather than only in the orchestrator, because the same bug was one layer down. `ApimKeyService.ProvisionAsync` and `RevokeCoreAsync` now run everything after APIM has accepted a change on `CancellationToken.None`, wrapped in the `AuditAfterCommitAsync` shape (try / Error log naming the row to reconcile / rethrow) — which means `DELETE /keys/{userId}` gets the fix too, not just my pipelines. Above them, `UserLifecycleService`, `UserService.UpdateQuotaAsync` and `QuotaAllocationService` (both call sites) compute a `completionToken` that is `None` **only when something actually reached the gateway** (`key is not null || resolution.TierSyncRequested`), so a request that touched nothing external is still cancellable.

Your probe is now a test — twice. `FakeApimManagementClient.AfterMutation` fires the instant APIM accepts a create or a delete, and the tests cancel the request token from it:

- `A_client_that_disconnects_the_instant_APIM_deletes_the_subscription_still_gets_a_full_deactivation` — subscription gone, `IsActive = false`, key fields cleared, allocation hard-stopped, pending request rejected, **both** `user.deactivated` and `key.revoked` present.
- `A_client_that_disconnects_the_instant_APIM_mints_the_key_still_gets_a_saved_user` — the same guarantee on the provision side.

I left `RotateAsync` alone (no lifecycle path reaches it) but it has the same gap and a nastier one behind it: its compensation is guarded by `when (exception is not OperationCanceledException)`, so a cancellation is precisely the failure that skips it. Filed as **#168** with the probe shape to use.

### Major 2 — `MoveToProductAsync` committing its caller's unit of work

Removed the save. The `key.tier-changed` row is now *added* and commits with whatever the caller commits; the interface documents that a caller which does not save is a bug, because the gateway would be enforcing a tier the audit trail never recorded. Verified every path that can reach the seam saves exactly once: `UserService.UpdateQuotaAsync`, `QuotaAllocationService.GetMyAllocationAsync` / `ResetAsync`, the lifecycle pipeline, #148's `ApproveAsync`, and — after merging #152 — `GroupService` and `EntraGroupSyncService`, both of which already had the `commitToken` discipline.

Your second probe is pinned as `A_quota_change_whose_audit_row_cannot_be_written_is_not_committed` (a `FailingAuditService` that fails only `user.quota-changed`, so the setup calls still work). It asserts the honest split: the gateway **is** on `power` — that residue is the accepted, idempotent direction — and the database has no new quota *and* no orphan `key.tier-changed` row.

`ResetAsync`'s "in one transaction" promise is true again, for the reason you identified: with the mid-loop flush gone, a failed ARM move aborts before the single save, so nothing is half-applied.

### Major 3 — deprovision compensation and N deletes in one transaction

You were right that my transaction argument only held for provision. Deprovision now does the APIM `DELETE` **first and outside** any transaction it owns, then takes the database steps in a transaction of their own on `CancellationToken.None`. The residue is one-directional and recoverable: key revoked and audited, user still active — logged at Error naming the user, and re-running the deactivation finishes it because revocation tolerates a missing subscription. `A_deactivation_whose_database_half_fails_leaves_a_recoverable_state_not_a_silent_one` walks exactly that, including the successful retry.

`EntraUserSyncService` no longer spans anything: the Graph enumeration happens before any transaction exists (your other Minor), each departure is its own unit of work in a `try`/`catch (UpstreamDependencyException)`, failures are counted in a new `UserSyncResult.FailedCount`, and adds/updates commit afterwards with the `users.synced` row. `One_departure_the_gateway_refuses_is_counted_and_does_not_undo_the_others` seeds two departures, fails one at the ARM layer, and asserts the other stays deprovisioned while the failed one keeps its key for the next run.

The whole ordering rationale is now a compensation table of its own in `plans/21`.

### Minors

1. **502 escaping as 500 on the tier-move path** — `ApimGatewayTierSync` wraps ARM failures in `UpstreamDependencyException`. Dropped the false "nothing was saved" wording: what is true is that the caller will not commit, because the exception aborts it before its save.
2. **Transaction spanning the Graph enumeration** — gone with the rest of that transaction.
3. **`LastSyncedDate` on every `/me`** — `RefreshFromClaims` returns whether anything actually differs and only then saves, and it no longer touches `LastSyncedDate` at all, so the column means what `UserContracts` says again. `LastLoginDate` is **#167**, with the "don't reintroduce the write-per-load problem" constraint written into the ask.
4. **`AdminProvision` had no callers** — wired up: `POST /keys/{userId}/provision` now runs the trigger via a new `IUserLifecycleService.ProvisionKeyForUserAsync`, and `IApimKeyService.ProvisionForUserAsync` is deleted. I **dropped `?tier=`** rather than making it an override: a caller-supplied tier can disagree with the allocation the database records, which is the exact drift the tier sync exists to prevent. To mint on a different product you set the quota, which moves the gateway too. (The entry point has to live on the orchestrator — the key service can't depend on resolution, since resolution depends on it through the sync.)
5. **Reactivation leaving `IsHardStopped`** — cleared in the pipeline.
6. **Missing indexes** — added, and not as two singles: `IX_Users_DisplayName` serves the unfiltered ordering and `IX_Users_IsActive_DisplayName` serves the far more common `?isActive=` view with filter and ordering from one seek. Mirrored in `Users.sql`. Documented on the entity that `?search=`'s `LIKE '%…%'` scans and that this is the right trade at fork scale.
7. **`api.md` stating the key survives a re-scope as fact** — now a `:::caution` block that says it is proven against the in-memory APIM only, points at #132, and spells out what changes if real ARM disagrees (create-on-new-product + delete-old, and the tier change becomes `key.rotated`).
8. **The `ApimGatewayUrl` config row lying** — marked legacy in `SystemConfigurationKeys` in the same register as `ApimProductId`, noting the editor (#161) treats legacy keys as read-only.

### Nits

- **`IsUpstreamFailure` denylist** → allowlist (`RequestFailedException`, `ApimSubscriptionNotFoundException`, `HttpRequestException`, `TimeoutException`), shared by the lifecycle service and the tier sync. Our own `InvalidOperationException`s now surface as the 500s they are, and the two tests that faked ARM with an `InvalidOperationException` were changed to throw what ARM actually throws.
- **#154 lock window** — commented on the issue: the loser blocks on the unique index for the whole ARM round trip before its 409, so a naive retry pays it twice; worth measuring against shortening the lock window instead.

---

**Gates:** zero-warning Release build; **799 predeployment tests pass** (was 723 before merging #148/#152, 475 on main at the start of this wave); `dotnet format --verify-no-changes` clean; docs site builds (14 pages). Merged `origin/main` twice — #148 and #152 both landed clean, and #152's `GroupService`/`EntraGroupSyncService` already had the `commitToken` discipline, so the `MoveToProductAsync` change makes their tier rows commit *with* their saves rather than ahead of them.

**New issues from this pass:** #167 (`LastLoginDate`), #168 (`RotateAsync` cancellation gap). Still open from the first round: #153, #154, and the extended live checklist on #132. `Closes #160` added for the `CancelPendingForUserAsync` swap.

Issue #165, opened 2026-09-02
SystemConfiguration.ResetDayOfMonth is stored but nothing reads it


kolatts:
Discovered while implementing #161.

`ResetDayOfMonth` is seeded (`"1"`), listed by `GET /config`, and — as of #161 — validated on write (a whole number from 1 to 28, so the reset can never be scheduled on a day some month lacks). But `grep` finds no reader: the monthly reset is a timer trigger on its own schedule, and `POST /quota/reset` is manual. Changing the value today does nothing at all, which is exactly the failure mode #161's 409 branch exists to prevent for the other two dead keys.

Pick one:

1. **Honour it** — the monthly reset job (#38) reads the key and only resets on that day of month, or the timer schedule is derived from it. This is what the seed comment and spec §6 imply ("always \"1\" for v1").
2. **Retire it** — same treatment as #164: drop it from the seed set and refuse edits.

Option 1 is the honest reading of the spec, but it is a behavioural change to the reset job, so it belongs with that job rather than with the config editor.

Issue #164, opened 2026-09-02
Retire the two dead SystemConfiguration keys (ApimProductId, EntraTenantId)


kolatts:
Discovered while implementing #161 (`PUT /config/{key}`).

Two of the eight seeded `SystemConfiguration` keys are wired to nothing:

- `ApimProductId` — superseded by per-tier APIM products (`Gateway:Tiers` / `quotaTiers`); a developer's subscription is issued against their tier's product, not this one id.
- `EntraTenantId` — unused since #123; Graph access comes from the `Entra` options section over the host's `AzureAd:TenantId`.

`PUT /config/{key}` now refuses both with `409` and a message naming the replacement (`SystemConfigValidator.EnsureEditable`), so an operator can no longer be fooled into "editing" them. That is the safe half of the fix; the rows themselves are still seeded, still listed by `GET /config`, and still take up space in the admin editor.

**Work:** drop both from `SystemConfiguration.GetSeedData()` and `SystemConfigurationKeys.All`, keep the constants (or remove them with the callers), and let the seeder's `deleteFilter` clean the existing rows out of deployed databases. Then delete the read-only branch in `SystemConfigValidator` and its tests — an unknown key is already a `404`.

Deferred rather than done in #161 because deleting seeded rows on the next deploy of every fork is a data change, not an API change, and deserves its own review.

kolatts:
Claiming this on branch `claude/hardening-entra-groups-config` (hardening wave), together with #123 and `ApimGatewayUrl`.

Issue #55, opened 2026-06-04
Build /config MudTable editor and /audit MudDataGrid with MudDateRangePicker filters


kolatts:
Part of #18.

`/config` fetches `GET /config` and renders each `SystemConfiguration` entry as an editable row (display key read-only, value text input with an inline Save button that calls `PUT /config/{key}`). Show the last `UpdatedAt` and `UpdatedBy` alongside each row. `/audit` fetches `GET /audit` with filter inputs: actor name (type-ahead from user list), action (dropdown of known action strings), and date-range pickers. Results are paged with a detail JSON expansion per row.

kolatts:
Claimed — implementing on `claude/web-developer-dashboard-49-50-54-55` alongside #49, #50, #54.

PR #166, opened 2026-09-02
feat(api): add the /config and /dashboard admin endpoints

### LINKED ISSUE #161: Implement GET /config and PUT /config/{key} admin endpoints (SystemConfiguration editor API)
Spec §4.6 lists GET /config and PUT /config/{key}; #55 needs them; no backend issue existed. ConfigController + IConfigService: list, update (404 unknown key; per-key validation incl. DefaultMonthlyTokenQuota must be a tier cap; legacy keys read-only 409), UpdatedByUserId/UpdatedDate via TimeProvider, audit ConfigUpdated before/after, single save, tests, reference/api.md.

### LINKED ISSUE #162: Implement GET /dashboard admin summary endpoint
Spec §4.6 lists `GET /dashboard` (admin: summary stats) and the Blazor `/dashboard` page (#54) needs it — but no backend issue existed. Domain already has `DashboardSummaryResponse(TotalUserCount, ActiveUserCount, UnlimitedUserCount, PendingQuotaIncreaseRequestCount, TotalTokensUsedThisPeriod, TopConsumers)` and `TopConsumerResponse`.

Scope: `DashboardController` (`/api/v1/dashboard`, admin) + `IDashboardService`: counts via set-based queries, current-period (`BillingPeriod.Current`) token totals from `QuotaAllocations`, top 10 consumers by `TokensUsed` with `PercentUsed` (null when unlimited), `AsNoTracking` projections, short `IMemoryCache` (30 s) since the page auto-refreshes every 60 s. Tests + `reference/api.md`.

Filed by the orchestrator while planning wave 3 (frontend #54 depends on it).

kolatts:
Implements the two remaining spec §4.6 admin endpoints — the `SystemConfiguration` editor API and the dashboard summary — so the wave-3 frontend (#54, #55) has a backend to call.

Closes #161
Closes #162

## What landed

**`/api/v1/config` (#161)** — `ConfigController` + `Services/Config/{IConfigService, ConfigService, SystemConfigValidator}`.

- `GET /config` → every row ordered by key, `AsNoTracking` projection, with the editing admin's name joined in.
- `PUT /config/{key}` → `404` unknown key, `409` retired key, `400` invalid value, `200` with the updated row. Stamps `UpdatedByUserId` + `UpdatedDate` (TimeProvider), writes one `config.updated` audit row with `{ key, before, after }`, one `SaveChangesAsync`.
- `SystemConfigEntryResponse` gains a nullable `UpdatedByDisplayName` (last positional member) for the #55 editor. `FoundryGate.Web` compiles unchanged — it only deserializes the record.

**`/api/v1/dashboard` (#162)** — `DashboardController` + `Services/Dashboard/{IDashboardService, DashboardService}`.

- Six set-based `AsNoTracking` queries over `BillingPeriod.Current(TimeProvider)`, behind a 30 s `IMemoryCache` keyed by period; `?fresh=true` bypasses it.

1. **One validation table, in the API, not Domain.** `SystemConfigValidator` is a singleton over `GatewayTierMapper` because `DefaultMonthlyTokenQuota` must satisfy the same D-013 tier rule as every other quota write path, and the tier table lives in `GatewayOptions`. Rules: tier-cap quota; reset day 1–28; boolean; absolute-https-or-empty URL; ARM-resource-id-or-empty for the two resource id keys; free text for a row a fork operator added themselves (the seeder's `deleteFilter` deliberately preserves those, so the editor must not refuse to edit what the seeder refuses to delete). Values are stored **normalized** — trimmed, booleans lower-cased, trailing slash removed.
2. **The two dead keys are `409`, not `400` or a silent no-op.** `ApimProductId` (superseded by per-tier APIM products) and `EntraTenantId` (unused since #123) still exist because the seed data references them; editing either would do nothing, so the refusal names the replacement. Removing the rows is a *data* change and got its own issue (#164).
3. **Key lookup is in memory, matched with `OrdinalIgnoreCase`.** `Key == key` is case-insensitive under SQL Server's default collation and case-sensitive under the SQLite the tests run on. Rather than ship an endpoint that 404s on one provider and succeeds on the other, `UpdateAsync` materializes the table (nine rows) and matches in C#. The canonical key comes back in the response, not the caller's casing.
4. **`UpdatedDate` is set explicitly as well as by the interceptor.** Re-saving an unchanged value must still record who touched it and when, and EF never hands an unmodified entity to `TimestampInterceptor`. Both read the same `TimeProvider`, so they cannot disagree.
5. **`unlimitedUserCount` counts active users only**; `totalTokensUsedThisPeriod` includes deactivated users' allocations (tokens spent before offboarding are still tokens this month spent); `topConsumers` lists active users only. Each choice is asserted in a test rather than left to the reader.
6. **`PUT` returns the updated row (`200`), not `204`.** The `Web` client's `SendActionAsync` accepts any 2xx, and the editor needs the new `updatedDate`/`updatedByDisplayName` without a re-list.
7. **`?fresh=true` rather than a cache-busting header.** Two callers need it — the page after a mutation, and the tests after seeding — and a query parameter is the one the Blazor client can express.

## Verification

- `FoundryGate.Tests.Predeployment` — **749 passed, 0 failed** (78 new)
- `docs-site`: `npm ci && npm run build` — passes
- Merged `origin/main` (with #148's requests wave) before pushing; `ApiServiceCollectionExtensions` resolved to two added lines, `reference/api.md` sections do not overlap.

New tests: `SystemConfigValidatorTests` (the whole rule table, both `409` messages, the free-text fallback), `ConfigEndpointTests` (401/403 matrix on both actions, 400/404/409 mapping, actor/timestamp stamping, audit before/after, case-insensitive key, **and that a re-run of `ReferenceDataSeeder` never reverts an edit** — the `[DoNotUpdate]` guarantee the whole editor rests on), `DashboardServiceTests` (each number pinned against a private database, including period scoping, top-10 ordering, the percent-used branches and the per-period cache key), `DashboardEndpointTests` (auth matrix, wire shape, `?fresh=true`).

## Deferred, as issues

- **#164** — retire the `ApimProductId` / `EntraTenantId` rows from the seed set and drop the read-only branch.
- **#165** — `ResetDayOfMonth` is now validated on write but nothing reads it; either the monthly reset job honours it or it joins #164.

### DIFF
--- a/CONVENTIONS.md
+++ b/CONVENTIONS.md
@@ -47,7 +47,13 @@
 - Temporal: `DateTimeOffset`, names end in `Date` (`CreatedDate`, `ReviewedDate`).
   **[improvement]** All `CreatedDate`/`ModifiedDate` set by ONE
   `SaveChangesInterceptor` using injected `TimeProvider` — never inline
-  `DateTimeOffset.UtcNow` in services.
+  `DateTimeOffset.UtcNow` in services. **One exception**: a service may assign a
+  `ModifiedDate`-style field itself, from the injected `TimeProvider`, when the row must
+  record a re-save that changed no other column — EF never hands an unmodified entity to
+  the interceptor, so "who touched this and when" would otherwise be lost on a same-value
+  save (precedent: `ConfigService.UpdateAsync`). Both read the same `TimeProvider`, so the
+  explicit assignment and the interceptor cannot disagree; `DateTimeOffset.UtcNow` stays
+  banned either way.
 - **[improvement]** `TimeProvider` injected everywhere time is read. No naked
   `UtcNow` outside the interceptor's TimeProvider.
 - Strings: non-nullable, `[Required]` + `[StringLength(n)]`, default
--- a/docs-site/src/content/docs/reference/api.md
+++ b/docs-site/src/content/docs/reference/api.md
@@ -127,7 +127,61 @@ Rules the mutation paths enforce (CLAUDE.md "Anthropic deployments are create-on
 
 | Method | Path | Auth | Description |
 |---|---|---|---|
-| `GET` | `/config` | Admin | All SystemConfiguration keys |
-| `PUT` | `/config/{key}` | Admin | Update a configuration value |
+| `GET` | `/config` | Admin | Every `SystemConfiguration` key, ordered by key: `{ key, value, updatedDate, updatedByUserId, updatedByDisplayName }`. The last two are `null` for a seeded key no admin has edited |
+| `PUT` | `/config/{key}` | Admin | Set one key's value. Returns the row as it now stands |
 | `GET` | `/audit` | Admin | Audit log, paged. Filter: `?actor=&action=&from=&to=` |
-| `GET` | `/dashboard` | Admin | Summary stats for the dashboard |
+| `GET` | `/dashboard` | Admin | Summary stats for the dashboard. `?fresh=true` bypasses the 30-second cache |
+
+### `PUT /config/{key}`
+
+Body: `{ "value": "..." }`. The key is matched case-insensitively; an unknown one is `404`. The
+value is validated **per key** before it is stored — a configuration editor that accepts a value
+the system cannot use only moves the failure somewhere harder to find — and stored **normalized**
+(trimmed, booleans lower-cased, URLs and resource ids without a trailing slash), so two admins
+typing `True` and `true` leave the same row behind.
+
+| Key | Rule |
+|---|---|
+| `DefaultMonthlyTokenQuota` | A non-negative whole number that is exactly one of the configured tier caps (`Gateway:Tiers`; `GET /quota/tiers` lists them). Same D-013 rule as every other quota write path — the gateway enforces a tier, not a number. Unlimited is not expressible fork-wide: it is set per user or per group |
+| `ResetDayOfMonth` | A whole number from 1 to 28 (28 is the last day every month has) |
+| `EntraGroupSyncEnabled` | `true` or `false` |
+| `ApimGatewayUrl` | An absolute `https` URL, or empty |
+| `ApimResourceId`, `FoundryResourceId` | An ARM resource id (`/subscriptions/{id}/resourceGroups/{group}/providers/{namespace}/{type}/{name}`), or empty. Shape only — whether the resource exists is Azure's answer, reported as `503` by the endpoints that use it |
+| Any other row a fork operator added | Free text, up to 4000 characters |
+
+Two seeded keys are **read-only (`409`)**, because editing them would be a silent no-op. They stay
+in the table only so re-seeding is idempotent, and the refusal names the replacement:
+
+- `ApimProductId` — quota tiers are APIM *products* now, so a developer's subscription is issued
+  against the product for their tier (`Gateway:Tiers` in the API, `quotaTiers` in `infra/main.bicep`),
+  not against one fork-wide product id.
+- `EntraTenantId` — Graph access is configured by the `Entra` options section over the host's
+  `AzureAd:TenantId` ([#123](https://github.com/kolatts/foundry-gate/issues/123)).
+
+Every accepted edit stamps `updatedByUserId` (the calling admin) and `updatedDate`, and writes one
+`config.updated` audit row with `{ key, before, after }` — in the same transaction as the change, so
+a value can never move without a trail. The calling admin must already have a FoundryGate user row
+(`GET /users/me` provisions one) or the write is `403`; reading `/config` needs no such row.
+
+**A deploy never reverts an edit.** `SystemConfiguration`'s value, timestamp and editor columns are
+`[DoNotUpdate]`, so the reference-data seeder that runs after every database deploy only *inserts*
+keys that are missing.
+
+### `GET /dashboard`
+
+Returns `{ totalUserCount, activeUserCount, unlimitedUserCount, pendingQuotaIncreaseRequestCount,
+totalTokensUsedThisPeriod, topConsumers }` for the current UTC calendar month.
+
+- `unlimitedUserCount` counts **active** unlimited users — a deactivated account consumes nothing.
+- `totalTokensUsedThisPeriod` sums every allocation in the period, deactivated users included: the
+  tokens they spent before offboarding are still tokens this month spent.
+- `topConsumers` is the ten busiest **active** users this period, each with `userId`, `userUnique`,
+  `displayName`, `tokensUsed`, `allocatedTokens` and `percentUsed` — `null` for an unlimited user,
+  `100` for a zero quota with any usage.
+
+Every usage figure here is a reconciliation number from the Log Analytics sync, refreshed on that
+job's cadence — not a live view of gateway enforcement.
+
+The summary is computed once and served to every admin for **30 seconds**, keyed by billing period
+(so the first read after a month boundary can never show last month's figures). The admin page
+refreshes itself every 60 s; pass `?fresh=true` to recompute immediately after a change.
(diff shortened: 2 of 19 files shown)


kolatts:
**REQUEST CHANGES**

Good bones — the D-013 tier rule, the retired-key 409s, the `[DoNotUpdate]` guarantee are all correctly implemented and actually tested (`A_later_reference_data_seed_never_reverts_an_edited_value` is exactly the test I'd have asked for). Build's clean, 749/749 tests pass, docs build passes, `dotnet format` is quiet. But there's one real bug: two of the six documented validation rules are unreachable through the actual endpoint.

---

**Issues**

- **Major** `src/FoundryGate.Domain/Config/Contracts/ConfigContracts.cs` line 30 — `UpdateSystemConfigRequest.Value` carries a bare `[Required]`, which defaults to `AllowEmptyStrings = false`. `SystemConfigValidator.NormalizeHttpsUrl`/`NormalizeArmResourceId` explicitly accept `""` so an admin can clear `ApimGatewayUrl`, `ApimResourceId`, and `FoundryResourceId` back to empty — but `[ApiController]`'s automatic model validation rejects the empty-string body with a 400 ("The Value field is required") before the action, let alone `SystemConfigValidator.Normalize`, ever runs. I reproduced it: `PUT /config/ApimGatewayUrl` with `{"value": ""}` returns 400, not 200. The docs (`reference/api.md`: "An absolute https URL, **or empty**") and the validator's own unit tests both promise a capability the real HTTP pipeline blocks. No endpoint test sends an empty body, which is exactly why this shipped. Swap to `[Required(AllowEmptyStrings = true)]` (StringLength + Normalize's own null-check already cover the rest) and add the endpoint test.

- **Minor** `src/FoundryGate.Api/Services/Config/ConfigService.cs` line 50 — `UpdateAsync` materializes the whole table, mutates in memory, and saves with no concurrency token on `SystemConfiguration`. Two admins racing to edit the same key is a silent last-write-wins with no `DbUpdateConcurrencyException` and no signal to the loser. Table's tiny (nine rows) and this is a low-traffic admin surface, so I won't block on it, but worth a follow-up issue if the config page ever gets busier.

- **Nit** `src/FoundryGate.Api/Services/Config/ConfigService.cs` line 70 — `entry.UpdatedDate = timeProvider.GetUtcNow()` is a deliberate, well-commented deviation from CONVENTIONS.md's "All `CreatedDate`/`ModifiedDate` set by ONE `SaveChangesInterceptor`" rule (justified: EF won't hand an unmodified entity to the interceptor on a same-value resave, and it reads the same `TimeProvider` so the two can't disagree). Reasonable call, but since the convention doesn't carve out the exception in writing, it's worth a one-line addition to CONVENTIONS.md so the next person doesn't have to rediscover the reasoning from a code comment.

---

**FYI, not a blocker:** merging current `origin/main` (now includes #156) into this branch produces exactly one conflict, `ApiServiceCollectionExtensions.cs` — the adjacent-line collision CONVENTIONS.md already calls out as expected for parallel waves. `reference/api.md` merges clean, confirming the PR description's claim. Nothing to do here now, just flagging it'll need a trivial rebase before merge.

kolatts:
Thanks — the Major is a real bug and a fair catch about *why* it shipped: I had unit tests asserting `""` is accepted and docs promising it, and no endpoint test that actually sent one. The validator was tested in isolation from the pipeline that was blocking it.

Pushed as `8042c17` (fixes) + `030aa57` (merge of current `origin/main`). Point by point:

---

**Major — `UpdateSystemConfigRequest.Value` bare `[Required]` — fixed.**

Now `[Required(AllowEmptyStrings = true)]`. The attribute stays on rather than coming off entirely, so a literal `{"value": null}` is still a field-level 400 instead of a null reference deeper in; emptiness is the per-key rule's decision (`SystemConfigValidator`), not model binding's to make for all six keys at once. The XML doc on the property now says that, so the next person doesn't "tidy" it back.

Four new endpoint tests, all going over the real pipeline:

- `An_empty_value_clears_a_key_whose_rule_allows_it` — `ApimGatewayUrl` and `FoundryResourceId` → 200, and the row actually reads back empty. This is the exact repro you ran; it was 400 before the fix.
- `An_empty_value_is_still_refused_where_the_keys_rule_forbids_it` — `DefaultMonthlyTokenQuota`, `ResetDayOfMonth`, `EntraGroupSyncEnabled` → 400, value unchanged. Worth having both directions: the fix must not turn into "empty is fine everywhere", which would let someone blank the quota every fall-through user depends on.
- `An_over_long_value_is_a_field_level_400_not_a_500` — see the next point.

**One consequence worth flagging, since it changes a file outside my area.** With `AllowEmptyStrings = true` and the `= string.Empty` initializer the convention requires, `{}` and `{"value": ""}` are now the *same request* — so `RequestDtoBindingTests`' `{ "config", "{}", ... }` row could no longer be a 400. I took the lifecycle `RequestDtoBindingController` documents for itself ("once a production controller binds one of these records, its own endpoint tests take over and the matching action here can go"): removed the `config` action and its theory row, and replaced the coverage in `ConfigEndpointTests` with an over-long value that asserts a 400 `ApiError` naming `Value` — same #128 guard (never a 500), now against the real endpoint. The other six actions are untouched.

The alternative was `string? Value` so that omitted (null → 400) stays distinguishable from explicitly-empty (`""` → clear). It is more precise, and I decided against it: it needs a nullable string on a request DTO against the convention's "nullable only when null means something", plus null-forgiveness at the one call site, to buy a distinction only a hand-written client can even make. Happy to revisit if you'd rather have it.

---

**Minor — no concurrency token, last-write-wins — filed as [#170](https://github.com/kolatts/foundry-gate/issues/170), not fixed here.**

Taking your "won't block on it" at face value, and with a reason beyond that. The cheap fix is an optional `ExpectedUpdatedDate` on the request, 409 on mismatch — but shipping it now means a public request field with no caller: the admin config page (#55) doesn't exist, nothing would send it, and the shape would be frozen before the UI that has to use it is designed. Because it is optional it is purely additive, so deferring costs nothing in compatibility — which is the argument for doing it *with* #55, where the interesting question is what "Save all changes" does with a partial conflict across several dirty rows, not what one PUT returns.

#170 carries the full plan (including using the `updatedByDisplayName` this PR already returns, so the UI can say *who* got there first without another round trip) and the reasoning for preferring a request field over a `rowversion` column on a reference-data table whose every column is `[DoNotUpdate]`. `ConfigService`'s class doc now points at it, so the gap is visible from the code.

---

**Nit — CONVENTIONS.md doesn't carve out the exception — done.**

Added to §EF Core, right after the interceptor rule:

> **One exception**: a service may assign a `ModifiedDate`-style field itself, from the injected `TimeProvider`, when the row must record a re-save that changed no other column — EF never hands an unmodified entity to the interceptor, so "who touched this and when" would otherwise be lost on a same-value save (precedent: `ConfigService.UpdateAsync`). Both read the same `TimeProvider`, so the explicit assignment and the interceptor cannot disagree; `DateTimeOffset.UtcNow` stays banned either way.

Scoped deliberately narrowly — the licence is "the interceptor structurally cannot see this save", not "set timestamps wherever convenient".

---

**FYI — merge done.** `origin/main` now has #152 *and* #156, so the merge took the one expected `ApiServiceCollectionExtensions.cs` conflict (my two lines vs. `AddLifecycleServices`/`AddUsersServices`); resolved by keeping all four, mine last. `reference/api.md` merged clean again — the users/groups waves and the Admin section don't overlap.

**Gates after the merge:** build Release 0 warnings / 0 errors · Predeployment **881 passed, 0 failed** · `dotnet format --verify-no-changes` clean · docs-site `npm ci && npm run build` passes · CI green on `030aa57` (`build-test`, `docs-build`, Astro build).

Also documented the empty-value semantics in `reference/api.md`, since the old wording ("or empty") was describing something the API didn't actually do:

> Emptiness is one of those per-key decisions, not a blanket rule: `{ "value": "" }` **clears** a key whose rule allows empty (the URL and resource-id keys — how you unwire a resource) and is a `400` for one that does not.

Issue #172, opened 2026-09-02
The read-only SystemConfiguration key list lives in two places (Api validator + Web config page)


kolatts:
The `/config` admin editor (#55) disables a row and states the reason for keys that cannot be edited — `ApimProductId`, `EntraTenantId` and `ApimGatewayUrl`, whose rows nothing reads any more. That list is currently written twice:

- `FoundryGate.Api/Services/Config/SystemConfigValidator.EnsureEditable` (on PR #166) throws `ConflictException` → `409` for the retired keys, with the replacement in the message.
- `FoundryGate.Web/Pages/Config.razor`'s `ReadOnlyReason` repeats the same keys so the field is disabled instead of offering an edit that can only fail.

They already disagree by one entry: `SystemConfigurationKeys.ApimGatewayUrl`'s doc comment (landed with #156) says "the admin config editor treats it as read-only", but `EnsureEditable` on #166 does not `409` it — so today the Web page disables a key the API would happily write.

Fix: put the retired-key set and its reason in `FoundryGate.Domain.Constants` (both projects reference Domain), have `EnsureEditable` and the Web page read the same map, and add a test that the two agree. Web already degrades correctly if a key is added only on the server — the save reports the `409` next to the field — so this is about removing the duplicate, not about a broken path.

Related: `SystemConfigEntryResponse` gains `UpdatedByDisplayName` on #166; `Config.razor`'s `ChangedBy` helper is the single line that should switch to it once that merges.

kolatts:
Half of this is now moot: #166 merged, so `SystemConfigEntryResponse` carries `UpdatedByDisplayName` and `/config` renders the editor's name (PR #174). What remains is the duplicated retired-key list — `SystemConfigValidator.EnsureEditable` (`ApimProductId`, `EntraTenantId`) versus `Config.razor`'s `ReadOnlyReason` (those two plus `ApimGatewayUrl`, following that key's Domain doc comment). The Web page is the stricter of the two today; one map in Domain that both read would settle it.

kolatts:
Picking this up in claude/frontend-followups (frontend follow-ups wave: #190, #191, #171, #173, #172).

Issue #169, opened 2026-09-02
Web API client drifted from the API surface: wrong verbs on activate/deactivate/approve/reject and the wrong type for GET /users/{id}


kolatts:
Found while building the developer/admin pages (#49/#50/#54/#55). `src/FoundryGate.Web/Services/FoundryGateApiClient.cs` was written against the spec before the controllers landed, and several methods no longer match what the API actually serves. Every one of them would fail at runtime (405 / deserialization) the first time a page calls it:

| Client method | Client sends | API expects (`docs-site/src/content/docs/reference/api.md`, controllers) |
|---|---|---|
| `ActivateUserAsync` | `PUT /users/{id}/activate` | `POST` (`UsersController.ActivateAsync`) |
| `DeactivateUserAsync` | `PUT /users/{id}/deactivate` | `POST` (`UsersController.DeactivateAsync`) |
| `ApproveRequestAsync` | `PUT /requests/{id}/approve` | `POST` (`RequestsController.ApproveAsync`) |
| `RejectRequestAsync` | `PUT /requests/{id}/reject` | `POST` (`RequestsController.RejectAsync`) |
| `GetUserAsync` | deserializes `UserResponse` | returns `UserDetailResponse` (user + allocation + key + groups) — since #156 |
| `UpdateUserQuotaAsync` / activate / deactivate | `ApiCallResult<bool>` (body discarded) | all three return the updated `UserResponse` |
| `ProvisionUserKeyAsync` | no tier | `?tier=` was removed in #156 — the client never sent it, so this one is already correct, listed only for completeness |

Not fixed in the #49/#50/#54/#55 branch on purpose: every affected method belongs to the admin management pages being built concurrently (#51/#52/#53/#62/#63), and changing them under that agent would collide. Whoever lands those pages should fix the client alongside them, and `FoundryGateApiClientTests` should grow a case per verb so the next drift fails the build rather than the browser.

Issue #51, opened 2026-06-04
Build admin user MudDataGrid (/users) and tabbed detail page (/users/{id})


kolatts:
Part of #17.

`/users` renders a searchable, paged `<UserTable>` (columns: name, email, quota, tokens used, status, actions). Clicking a row navigates to `/users/{id}`. The detail page shows all user fields, a quota override form (set tokens or toggle unlimited), inline activate/deactivate buttons with confirmation, and a key management panel (provision, rotate, revoke buttons that call the `/keys` endpoints). Trigger a page reload after any mutation to keep displayed state fresh.

kolatts:
Claimed — implementing on `claude/web-admin-51-52-53-62-63` alongside #52, #53, #62, #63 (admin Web pages).

PR #174, opened 2026-09-02
feat(web): developer /me and /me/request pages, admin dashboard, config editor and audit viewer

### LINKED ISSUE #49: Build /me page: MudProgressLinear quota gauge, key display/rotate, and request history
Part of #16.

`/me` fetches `GET /quota/allocations/me` and `GET /keys/me`. Render a `<QuotaGauge>` component (linear progress bar, color-coded green < 80% / amber 80–95% / red > 95%) with the token count and percentage. Below it, a `<KeyDisplay>` component shows the last-4 masked key, a "Reveal" button (calls a separate API action and shows the full key once in-memory — never in a cookie or localStorage), and a "Rotate Key" button with a confirmation modal. At the bottom, a request history table fetches `GET /requests` (own only).

### LINKED ISSUE #50: Build /me/request MudForm with validation and submission feedback
Part of #16.

`/me/request` renders a form with a required justification textarea and a requested token count input (nullable — empty = requesting unlimited). Use Blazor's `EditForm` with `DataAnnotationsValidator`. On success, show an inline confirmation and redirect back to `/me`. Disable the submit button if a pending request already exists for the current period (derive from the request history loaded on the same page). Graceful error display for API failures.

### LINKED ISSUE #54: Build /dashboard with MudCard stat cards, top consumers MudDataGrid, and pending MudBadge
Part of #18.

`/dashboard` fetches `GET /dashboard` and renders stat cards (total users, active users, hard-stopped users, pending increase requests). A "Top Consumers" table shows the 10 highest `TokensUsed` users this period with a mini progress bar. The pending request count is highlighted as a badge/link that navigates to the filtered requests queue. The page auto-refreshes every 60 seconds using a `PeriodicTimer` in the component lifecycle.

### LINKED ISSUE #55: Build /config MudTable editor and /audit MudDataGrid with MudDateRangePicker filters

kolatts:
Builds the four pages a person actually uses FoundryGate through: the developer's `/me` and `/me/request`, and the admin's `/dashboard`, `/config` and `/audit`. Every stub is replaced; the admin-management pages (#51/#52/#53/#62/#63) are untouched.

## What landed

**`/me` (#49)** — one `GET /users/me` supplies the page (the resolved quota, the masked key and the gateway addressing arrive together, so no second and third round trip), alongside `GET /foundry/models` and the caller's own `GET /requests`.

- `<QuotaGauge>` — green below 80%, amber from 80% through 95%, red above 95%; an "Unlimited" state instead of a bar; the tier name beside the number **because the tier is what the gateway enforces**, and an explicit warning when `IsGatewayCapped` says the two disagree. It also names which level of the resolution chain produced the budget, so "why do I have this number" is answerable on the page.
- `<KeyDisplay>` — masked by default. *Reveal* calls `POST /keys/me/reveal` and holds the plaintext in a private field for that render only: no `localStorage`, no cookie, no session storage (spec §11). *Rotate* is behind a confirm dialog. A developer with no key yet gets "your key is being set up", not an error.
- **Configure your AI CLI** — Claude Code, Codex CLI and `curl` snippets filled in with this gateway's URL and this developer's deployment names, and with the real key while it is revealed. Every variable, header and config key is copied from `getting-started/cli-setup.mdx` per CLAUDE.md invariant 3 — including `CLAUDE_CODE_USE_FOUNDRY`/`ANTHROPIC_FOUNDRY_BASE_URL` rather than the plain `ANTHROPIC_BASE_URL` the issue sketched, because the doc is the verified source and it says Foundry mode. A test asserts the snippets still match the doc.
- Request history table and a "Request more" CTA.

**`/me/request` (#50)** — since D-013 the request is for a **tier**, not a number: APIM's `token-quota` is a per-product literal, so an arbitrary figure is one the gateway could never enforce. The dropdown is `GET /quota/tiers` filtered to tiers strictly above the caller's (a developer already on top is told there is nothing bigger to ask for). The Domain `SubmitQuotaIncreaseRequest` record is the bound model so its `DataAnnotations` are the only validation rules — its properties are init-only, so each field rebuilds the record with `with` and validation runs through `Validator.TryValidateProperty` against the record itself rather than restating the limits in the UI. Submit is disabled in flight and while a request is pending; a `409` warns and locks the form rather than navigating away.

**`/dashboard` (#54)** — four stat cards, a pending-request badge linking to the filtered queue, and a top-ten consumers grid with a per-row bar ("Unlimited" where there is no cap). Refreshes every 60 s from a `PeriodicTimer` that is cancelled and disposed with the page; a failed background refresh keeps the last good numbers rather than nagging. A scoped `DashboardStateService` publishes the pending count so the nav's "All Requests" link carries the same badge without a second admin-only fetch.

**`/config` + `/audit` (#55)** — the config editor stages edits with per-row dirty tracking and makes you confirm a before/after diff before any write; each key is then written on its own, so a value the API refuses stops that row alone and reports next to its own field. Retired keys are disabled with the reason rather than offering an edit that could only `409`. The audit viewer is server-paged and server-filtered (the log is unbounded — client-side filtering would filter the wrong 25 rows), with action and target-type choices reflected off the Domain constants the writers use, a date range inclusive of its end day, and a details blob pretty-printed when it is JSON and shown verbatim when it is not.

`/` now sends an admin to `/dashboard` and everyone else to `/me` instead of duplicating half of `/me`.

## Verification

- `dotnet test FoundryGate.Tests.Predeployment` — **1010 passed, 0 failed**, of which **123 are new** (125 in `Web/` including the two that were already there)
- `dotnet run --project src/FoundryGate.Web` — starts and serves: `/` → 200 with `<title>Foundry Gate</title>`, `/css/app.css` → 200, `/_framework/blazor.webassembly.js` → 200
- `docs-site`: `npm ci && npm run build` — 15 pages built, complete

### plans/16 — Developer UI

- [x] `dotnet build` passes
- [x] Quota gauge renders at correct colour thresholds (0%, 80%, 95%, 100%) — `QuotaDisplayTests` pins the boundaries as a table; `MePageTests.Colours_the_bar_by_how_much_of_the_budget_is_gone` checks the rendered class
- [x] "Unlimited" chip shows when allocation has no token cap — `An_unlimited_developer_gets_a_chip_not_a_bar`
- [x] Rotating a key opens confirmation dialog, then shows the new key revealed once — `Rotating_asks_for_confirmation_before_it_touches_the_api`
- [x] CLI setup section shows the correct APIM gateway URL — from `UserProfileResponse.CliConfig.GatewayBaseUrl`; `The_cli_panel_shows_the_gateway_paths_and_the_deployment_names`
- [x] CLI setup section lists available model deployment names — same test, plus an empty-list state
- [x] Copy-paste snippets render the user's actual key (revealed) and correct gateway URL — `Revealing_the_key_fills_it_into_the_cli_snippets`
- [x] Request form disables submit while in-flight and re-enables on response — `Submit_is_disabled_while_the_request_is_in_flight`
- [x] Duplicate pending request attempt shows a warning without navigating away — `A_409_warns_and_locks_the_form_without_navigating_away`
- [x] Successful submission navigates back to `/me` — `Navigates_back_to_me_after_a_successful_submission`

### plans/18 — Admin dashboard, config, audit

- [x] `dotnet build` passes
- [x] Dashboard stat cards show correct counts from the API — `Shows_the_four_stat_cards_from_the_summary`
- [x] Pending request badge on the nav item reflects the live count — `The_dashboard_and_the_nav_agree_on_the_pending_count` renders both against the one scoped service
- [x] Config editor shows only dirty rows in the diff preview dialog — `The_diff_lists_only_the_dirty_rows`
- [x] Saving config calls PUT for each changed key and shows a success snackbar — `Confirming_writes_one_put_per_dirty_key`
- [x] Audit log date range filter works across a month boundary — `A_date_range_covers_the_whole_of_its_last_day_including_across_a_month_boundary` (30 Aug → 2 Sep)
- [x] Audit log details render JSON without crashing on malformed input — `Details_that_are_not_json_still_render_instead_of_throwing`

## Review fix pass (2026-09-02)

Everything in the REQUEST CHANGES review is addressed; the point-by-point reply is a comment on this PR. The three Majors, in short:

- **`/me/request`'s single-tier auto-select submitted unlimited.** The shortcut assigned the product id without going through `OnTierChanged`, so `RequestedQuota` stayed `null` — which the request record documents as "requesting unlimited". A two-tier fork (Standard + Power, no Unlimited product) showed "Power" and submitted unlimited. Fixed, and the reviewer's probe is now a test.
- **Admins saw, and were blocked by, other people's requests.** `GET /requests` returns everyone's to an admin, so `/me`'s history and `/me/request`'s pending pre-check both name the caller's own `UserId` from the profile now. Covered by tests rendered under an admin identity.
- **Harness collision with #176.** This PR's harness is now the canonical one and #176 rebases onto it: `WebTestContext` matches #176's shape, `FakeFoundryGateApiClient` is `partial` and carries #176's result properties, capture lists and `Arrange*` helpers, and `WebTestData` gains its `User`/`UserDetail`/`Group`/`Member`/`Deployment` factories. Two deliberate differences for #176 to adjust: `Tiers` puts Power at **20M** (mirroring `infra/main.bicep`, not 25M), and `UserDetailResult` is spelled `UserResult` here because `GET /users/{id}` returning `UserDetailResponse` is #169's fix, not this PR's.

Also in this pass: `cli-setup.mdx` corrected to say **pin the alias, not the deployment name** — proved by `infra/main.bicep`'s `productModelAliases`, where the alias map *is* the per-tier allowlist and anything else is `403 model_not_permitted` — with a new `CliSetupDocFidelityTests` that parses that doc's fenced blocks instead of asserting against a hard-coded copy. `/audit` resets to page one on a filter change and reads the picker's wall-clock dates in the admin's own zone through an injected `TimeProvider`. `/config` drops its duplicated retired-key list in favour of the API's `409`, which #186 immediately vindicated by deleting those keys from `SystemConfigurationKeys` — the hard-coded list would have been a build break on that merge rather than a no-op. The fabricated `409` response object and the last three naked `DateTimeOffset.UtcNow` are gone; pages thread a page-lifetime `CancellationToken` and `Dashboard` cancels its in-flight request on dispose so a late reply cannot publish into the nav badge; `KeyDisplay` no longer assigns to its own `[Parameter]` and awaits the hide that clears the plaintext; every page heading is an `<h1>` so `FocusOnNavigate` has something to focus; and both plan files' verification boxes are ticked.

Two substitutions the review flagged are now issues rather than silent gaps: **#190** (the dashboard's hard-stopped card — `DashboardSummaryResponse` carries no such count) and **#191** (the audit actor name type-ahead — `AuditLogQuery` takes only `actorUserId`). **#188** rules that #176's `TierDisplay` wins over this PR's `QuotaDisplay`, so `QuotaDisplay` is left exactly as it was and #176 replaces its usages and deletes it on rebase.

## Notes for the reviewer

**bUnit is new here, and `WebTestContext` is meant to be reused** by the admin-management wave. Two things in it are load-bearing and commented in place: the context is torn down through `IAsyncLifetime` because MudBlazor registers `IAsyncDisposable`-only services and a container holding one refuses synchronous disposal; and `RenderPage<T>` wraps the page in MudBlazor's popover and dialog providers and returns the **whole** tree, because an inline `MudDialog` renders as a sibling of the page — a result scoped to the page alone can neither see the confirm dialog nor click its buttons.

**Package bump.** bUnit 2.9.0 requires the ASP.NET Core WASM packages at ≥ 10.0.10, so the four `Microsoft.AspNetCore.Components.WebAssembly*` / MSAL versions move 10.0.8 → 10.0.11 — the same servicing band the rest of the repo already pins (EF Core, `Mvc.Testing`, OpenApi).

**Shared files, kept minimal.** New client methods live in `FoundryGateApiClient.Me.cs` / `IFoundryGateApiClient.Me.cs` (the existing types became `partial` in one commit) so the concurrent frontend wave adds methods without colliding. `NavMenu.razor` changed only the "All Requests" link (badge) plus an inject and a `@code` block; `_imports.razor` is append-only; `Program.cs` gained one registration line.

**Confirm dialogs are inline `MudDialog`s**, not `IDialogService.Show<T>` — the confirmation is part of the component that owns the action, which keeps it testable without a dialog-service double.

**Deferred, filed as issues:**
- **#169** — the Web API client has drifted from the API: `ActivateUserAsync`/`DeactivateUserAsync`/`ApproveRequestAsync`/`RejectRequestAsync` send `PUT` where the controllers are `POST`, and `GetUserAsync` deserializes `UserResponse` where `GET /users/{id}` now returns `UserDetailResponse`. Every one belongs to a page the other agent is building, so fixing them here would collide; left for that wave with a note to add a client test per verb.
- **#172** — the retired-config-key list is written twice (`SystemConfigValidator.EnsureEditable` and `Config.razor`'s `ReadOnlyReason`) and already disagrees by one entry: `ApimGatewayUrl`'s Domain doc comment says the editor treats it as read-only, but the API does not `409` it. The Web page follows the Domain comment and disables it; the issue proposes one map in Domain that both read.

Closes #49
Closes #50
Closes #54
Closes #55


kolatts:
**REQUEST CHANGES**

This is the best-documented PR in the repo so far — the D-013 tier form, the "reveal lives in a private field and nowhere else" discipline, and 115 bUnit tests that assert behaviour instead of markup are all exactly right. Blocking on three scoping/selection bugs on the pages a developer touches every day, one of which quietly turns a request for "Power" into a request for unlimited.

Verified on the branch: build 0 warnings, 996 tests pass, `dotnet format` clean, docs-site builds 15 pages, `/` serves 200.

---

**Issues**

- **Major** `Pages/MeRequest.razor:157` — `_selectedTierProductId = _availableTiers[0].ProductId` when there's exactly one option, but that path never goes through `OnTierChanged`, so `_model.RequestedQuota` stays `null` — which `SubmitQuotaIncreaseRequest` documents as "requesting unlimited". I probed it with a two-tier fork config (Standard + Power, no Unlimited product): the dropdown shows Power, the submitted `RequestedQuota` is `null`. The repo's three-tier default dodges it because the single option there really is Unlimited; a fork without an Unlimited product does not. `OnTierChanged(_availableTiers[0].ProductId)` closes it, and the two-tier case is worth a test — the whole point of D-013 is that the number and the tier are the same thing.

- **Major** `Pages/MeRequest.razor:131` — the pending pre-check sends `QuotaRequestQuery(Pending, UserId: null)`, and `QuotaRequestService.ListAsync` returns *every* user's pending requests to an admin. So any admin who opens `/me/request` is told "you already have a request waiting for review, submitted [a stranger's date]" and the form locks — for as long as anyone in the fork has something pending. `profile.Value.UserId` is right there on the response.

- **Major** `Pages/Me.razor:149` — same root cause on the history table: `GetRequestsAsync(paging)` unfiltered means an admin's "Your quota increase requests" lists other people's. #49 says "own only", and `web-ui.md` repeats the claim. `MeRequestPageTests.Asks_the_api_only_for_pending_requests_when_checking` asserts `query.Status` and never `query.UserId` — worth asserting both once this is fixed.

- **Major** cross-PR — #176 independently creates `Web/WebTestContext.cs`, `Web/WebTestData.cs` and `Web/FakeFoundryGateApiClient.cs` with materially different implementations (its fake is 379 lines to this one's 238; its context is abstract and registers no `DashboardStateService`). That's a guaranteed add/add conflict on ~700 lines of harness, and whoever resolves it second will be tempted to take one side wholesale and quietly break the other suite. **My vote: land this one first** — it owns `Home.razor`, `Program.cs` and the harness this PR advertises as reusable — then #176 rebases and folds its extra fake methods into this fake instead of shipping a second one. Worth agreeing that before either merges.

- **Minor** `Components/CliSetupPanel.razor:84` — "Pin your CLI to the alias, not the deployment name" and `ModelFor`'s alias-wins ordering aren't in `cli-setup.mdx`, which still says to pin deployment names. The file's own header comment claims the content is copied verbatim from that doc, and CLAUDE.md invariant 3 says the doc leads. `infra/main.bicep:119` suggests the panel is the one that's right (the alias map *is* the per-tier allowlist), which makes this a stale doc rather than a bad snippet — but the fix is to update `cli-setup.mdx` with the evidence in this PR, not to let the app and the doc disagree about what goes in `model=`.

- **Minor** `Tests/Web/MePageTests.cs:186` — `The_cli_snippets_match_the_verified_instructions_in_cli_setup_mdx` doesn't read `cli-setup.mdx`; it hard-codes five `Assert.Contains` strings, none covering the model-pinning lines. The PR body says "a test asserts the snippets still match the doc" — it asserts they match a copy of the doc that drifts silently. Parsing the fenced blocks out of the `.mdx` would make the claim true, and would have caught the item above.

- **Minor** `Pages/Audit.razor:214` — `ReloadServerData()` doesn't reset the grid's page. Confirmed it: navigate to page 20, change the Action filter, and the page still requests page 20 of the narrowed set and renders "No audit entries match these filters" for a filter that has plenty. Reset to page 0 before reloading.

- **Minor** `Pages/Audit.razor:171` — the picker hands back local dates, they're wrapped as `new DateTimeOffset(start.Date, TimeSpan.Zero)`, and the grid renders `OccurredDate.ToLocalTime()`. For any admin not sitting in UTC the filter window is offset from the timestamps on screen by their UTC offset. The month-boundary test passes because it only checks the date part.

- **Minor** `Pages/MeRequest.razor:263-276` — the 409 branch fabricates a whole `QuotaIncreaseRequestResponse` with zeroed ids just to flip a flag, and the alert then tells the user their request was "submitted [today]", which the page has no way to know. A `bool _lockedByConflict` says the same thing without inventing data — and it drops the three naked `DateTimeOffset.UtcNow` reads, which CONVENTIONS.md bans outright ("No naked `UtcNow` outside the interceptor's TimeProvider"). They're the only ones in the diff.

- **Minor** `Pages/Dashboard.razor:180-208` — `RefreshLoopAsync` has a `CancellationToken` and `LoadAsync` doesn't take one, so `Dispose()` cancels the timer but never the in-flight `GET /dashboard`, which still lands and writes to `DashboardState`. Same omission on every other page's calls (`Audit.razor`'s `ServerData` is the one that gets it right). Also: `StartRefreshLoop()` runs *after* the first `await LoadAsync(...)` — dispose during that window and `Dispose()` sees nulls, then a timer and a fire-and-forget loop get created that nothing ever cancels, polling an admin endpoint for the life of the tab.

- **Minor** `Pages/Config.razor:275` — `ReadOnlyReason` disables `ApimGatewayUrl`, but `SystemConfigValidator.EnsureEditable` doesn't 409 it, so the editor blocks a key the API would happily write. #172 covers the duplication; worth deciding now which of the two is correct rather than shipping an editor that refuses a legal edit.

- **Minor** `Services/QuotaDisplay.cs` — #176 ships `Shared/TierDisplay.cs` for the same concept, with its own `FormatTokens`, and the two resolve tier names differently (hard-coded switch here vs. lookup from `GET /quota/tiers` there). They'll disagree the first time a fork adds a tier product. One of them should win in the merge.

- **Minor** `docs-site/.../reference/web-ui.md:98` — calls `/requests` and `/requests/{id}` admin routes that are "already routed and role-gated". They're `@attribute [Authorize]` with no role, and `NavMenu` links `/requests` as "My Requests" under **Developer**. `/users/{id}` is missing from the list, and line 7's "every route requires an authenticated caller" isn't true of `/`.

- **Minor** #54 asked for a hard-stopped-users stat card and #55 for an actor-*name* type-ahead; what shipped is a tokens-used card and a raw "Actor user id" numeric field. Both look like reasonable calls given `DashboardSummaryResponse` has no hard-stopped count and `AuditLogQuery` only takes an id — but CLAUDE.md says the gap becomes an issue, not a silent substitution in the PR that closes the issue.

- **Nit** `Components/KeyDisplay.razor:166,192` — `RotateAsync` assigns to the `[Parameter] Key` property (works only because `Me.razor` handles `KeyChanged`; any other parent, or a re-render before the callback lands, reverts to the stale mask), and `Hide()` is `void` discarding the task that clears the plaintext out of the CLI snippets. Both one-liners, and neither is a leak today — the storage discipline itself is clean: no `localStorage`, no cookie, no URL, clipboard is the only interop that ever sees the value.

- **Nit** There's no `<h1>` anywhere in the app — every page title is `Typo.h4` — so `App.razor`'s `FocusOnNavigate Selector="h1"` has never matched anything. Pre-existing, but this PR adds five more page headings to it.

- **Nit** `plans/16-developer-ui.md` and `plans/18-admin-dashboard-ui.md` still have every verification box unchecked. The PR body ticks them; the plan files are what CLAUDE.md asks for.

kolatts:
All fifteen addressed. Re-verified on the branch after merging `origin/main` (which now carries #156, #166, #184 and #186): build 0 warnings, **1010 tests pass** (125 in `Web/`, up from 117), `dotnet format --verify-no-changes` clean, docs-site builds 15 pages, `/` serves 200.

The two-tier probe was the right call — thank you for actually running it.

---

### Majors

**1. `MeRequest.razor:157` — single-tier auto-select submitted unlimited.** Fixed: the shortcut now calls `OnTierChanged(_availableTiers[0].ProductId)` instead of assigning the id, so there is exactly one path that keeps the number and the tier the same thing. Your probe is now a test — `A_fork_with_one_option_left_submits_that_tiers_cap_not_unlimited` arranges a two-tier fork with no Unlimited product and asserts the submitted `RequestedQuota` is `20_000_000`. It fails on the old code.

**2. `MeRequest.razor:131` — admins locked out by a stranger's pending request.** Fixed: the pre-check now sends `QuotaRequestQuery(Pending, UserId: profile.Value.UserId)`. That reorders the loads — the profile has to land before the check can be made — which the comment explains rather than leaving as a mystery serialization. Two tests: `Asks_the_api_only_for_the_callers_own_pending_requests` (asserts both `Status` **and** `UserId`, as you suggested) and `An_admin_is_not_told_a_strangers_pending_request_is_their_own`, which renders under an admin identity.

**3. `Me.razor:149` — the history listed other people's requests.** Same fix, same reason: `GetRequestsAsync(new QuotaRequestQuery(null, _profile.UserId), …)`. `The_history_asks_for_the_callers_own_requests_only` and `An_admins_own_page_does_not_list_other_peoples_requests` cover it. The profile and the model list still go out in parallel; only the history waits, because only the history needs the id.

**4. Cross-PR harness collision with #176.** Agreed, and taken: this PR's harness is now the canonical one, shaped so #176's rebase is mostly deletion.

- `WebTestContext` matches #176's shape exactly where the two can agree — `abstract`, an `Authorization` property, `void SignInAsAdmin/Developer/SignOut`, `RenderPage<TPage>(params (string, object?)[])` returning `IRenderedComponent<IComponent>`, `AddMultipleAttributes` (which also retired my `ASP0006` pragma), and an empty `Dispose(bool)` override rather than relying on bUnit's guard. It keeps the `DashboardStateService` registration the nav badge needs, adds a swappable `TimeProvider`, and exposes `Snackbars`.
- `FakeFoundryGateApiClient` is now **`partial`**, so the next wave adds its new interface methods in a `FakeFoundryGateApiClient.Admin.cs` beside it rather than editing this file at all. It has folded in #176's result properties (`UsersResult`, `UserDetailResult`→`UserResult`, `GroupsResult`, `GroupDetailResult`, `GroupMembersResult`, `CreateGroupResult`, `GroupSyncResult`, `RequestResult`, `MutationResult`, …), its capture lists (`RequestListCalls`, `GroupListCalls`, `ActivatedUserIds`, `DeactivatedUserIds`, `QuotaUpdates`, `CreatedGroups`, `GroupUpdates`, `DeletedGroupIds`, `AddedGroupMembers`, `RemovedGroupMembers`, `EntraSyncedGroupIds`, `ApprovedRequests`, `RejectedRequests`, `ProvisionedKeyUserIds`, `RotatedKeyUserIds`, `RevokedKeyUserIds`) and its fluent `Arrange*` helpers, all under the same names.
- `WebTestData` gains `User`, `UserDetail`, `Membership`, `Group`, `Member`, `Deployment`; `Tiers` is now a property, and `Allocation`/`Request` take #176's parameter names (`userId`, `level`, `requestId`, `userDisplayName`).

**Two things #176 will still have to touch**, both mechanical: `WebTestData.Tiers` puts Power at **20M**, mirroring `infra/main.bicep`'s `quotaTiers`, not 25M — a fixture that matches the shipped caps is one a reader can sanity-check; and its `UserDetailResult` is spelled `UserResult` here since `GET /users/{id}` returning `UserDetailResponse` is #169's fix, not this PR's.

---

### Minors

**`CliSetupPanel.razor:84` — panel and doc disagreed on alias vs deployment name.** You were right that the panel was correct and the doc was stale, so I fixed the doc, with the evidence: `infra/main.bicep`'s `productModelAliases` comment says the alias map *is* the allowlist and anything else is `403 model_not_permitted`. `cli-setup.mdx` now leads with a caution saying to put the alias in `model`, and its snippets use `<sonnet-alias>` / `<openai-alias>` placeholders throughout. The panel matches: `ModelFor` no longer falls back to a deployment name at all (substituting one would produce a snippet that is refused), the model table is retitled "Deployments behind the gateway" with the aliases above it as the thing you actually pin, and the `403` is called out on both.

**`MePageTests.cs:186` — the fidelity test asserted against a copy.** Replaced by `CliSetupDocFidelityTests`, which reads `docs-site/.../cli-setup.mdx`, parses every fenced block, strips prose comments, and asserts each configuration line appears on the rendered page with this developer's values substituted. A third test guards the guard (the parse must yield ≥ 8 lines and include `export CLAUDE_CODE_USE_FOUNDRY=1`) so it can't pass vacuously, and a fourth pins the alias-vs-deployment guidance by name in both. I checked it fails on the pre-fix doc.

**`Audit.razor:214` — filter change kept the page number.** Fixed: `ReloadAsync` calls `NavigateTo(Page.First)` when the grid is not already on page one. `Changing_a_filter_goes_back_to_the_first_page` seeds a 500-row total, pages to the last page, changes the Action filter, and asserts the next query asks for page 1 with the new filter.

**`Audit.razor:171` — picker dates read as UTC.** Fixed: `ToInstant` applies the reader's own offset, taken from an injected `TimeProvider.LocalTimeZone` rather than `TimeZoneInfo.Local` so a test can pin it. `Program.cs` registers `TimeProvider.System` (also the sanctioned clock for any component that later needs one — CONVENTIONS.md bans naked `UtcNow`). The month-boundary test now runs under a `FixedZoneTimeProvider(-05:00)` and asserts the instant, not just the date part: `2026-08-30T00:00-05:00` == `05:00Z`.

**`MeRequest.razor:263-276` — fabricated response and three naked `UtcNow`.** Both gone: a `_lockedByConflict` flag, and the alert only names a submission date when it has a real request to name one from. The 409 test now asserts the alert says "waiting for review" and does **not** say "submitted". There are no `DateTimeOffset.UtcNow` reads left in the diff.

**`Dashboard.razor:180-208` — no cancellation, and a loop that could outlive a dispose.** Both fixed. `_pageLifetime` is a field initializer (so `Dispose` during the first request has something to cancel), `StartRefreshLoop()` runs *before* the first `await` and no-ops if already disposed, `LoadAsync` passes the token and returns without publishing when cancellation was requested. `A_reply_that_lands_after_disposal_is_not_published_to_the_nav_badge` gates the request, disposes, releases, and asserts `DashboardState.PendingRequestCount` is still 0. `Me`, `MeRequest` and `Config` now carry the same page-lifetime token through their calls.

**`Config.razor:275` — editor blocked a key the API would write.** Decided in favour of the API, and the duplicate is gone: `ReadOnlyReason` is deleted, every row is editable, and a refusal is reported next to its own field using the API's own message. This turned out to be load-bearing within the hour — #186 merged and removed `ApimProductId`, `EntraTenantId` and `ApimGatewayUrl` from `SystemConfigurationKeys` entirely, so the hard-coded list would have been a build break on the rebase rather than a no-op. Two tests: the API's reason is what renders, and a rejected value stays in its box to be corrected.

**`Services/QuotaDisplay.cs` vs #176's `TierDisplay`.** Filed as #188 and ruled in `TierDisplay`'s favour — it names a tier from the live `GET /quota/tiers` catalogue instead of a switch over the shipped ids. So `QuotaDisplay` is unchanged here (its remarks now point at #188 and note why the developer pages resolve from a product id: `/me` makes one profile call and holds no tier list), and #176 replaces the usages and deletes it on rebase. I'd briefly folded the two the other way round and reverted it — that revert is its own commit.

**`web-ui.md:98` — wrong claims about `/requests` and `/`.** Fixed. The "still being built" section is now a table with the real auth for each route, `/requests` explicitly called out as **not** admin-only ("a developer sees their own here, an admin sees everyone's — the API scopes the list, not the route"), `/users/{id}` added, and line 7 corrected to "every route except `/`".

**#54's hard-stopped card and #55's actor type-ahead.** Fair — filed rather than substituted silently: **#190** (add `hardStoppedUserCount` to `DashboardSummaryResponse` and give it its own alerting card) and **#191** (replace the numeric actor field with a `MudAutocomplete` over `GET /users?search=`, once #176 lands that client overload). Both plan files now record the substitution and link the issue, and `web-ui.md` links them from the relevant sections.

---

### Nits

**`KeyDisplay.razor:166,192.**` Both fixed. `RotateAsync` writes to a private `_rotatedKey` and lets `OnParametersSet` drop it once the parent's `Key` catches up, so the panel no longer depends on the parent handling `KeyChanged` to keep showing the right mask — `The_rotated_mask_survives_the_next_render` rotates, then hides, then asserts the new mask. `Hide()` is now `HideAsync()` returning the task, since that task is what clears the plaintext out of the CLI snippets.

**No `<h1>` anywhere.** Fixed for the six pages this PR owns: each heading is `<MudText Typo="Typo.h4" HtmlTag="h1">`, so `FocusOnNavigate Selector="h1"` finally has something to focus. The pages #176 owns still need the same one-word change.

**Plan verification boxes.** Ticked in `plans/16-developer-ui.md` and `plans/18-admin-dashboard-ui.md`, with the CLI-setup line rewritten to say aliases (and why), and plan 18 carrying a short "substituted, with an issue rather than a silent gap" section pointing at #190 and #191.

Issue #175, opened 2026-09-02
docs: add a reference/web-ui page describing the Blazor routes


kolatts:
## Problem

`docs-site/src/content/docs/reference/` documents the API, configuration,
infrastructure, CI/CD, rate limits and cost — but nothing describes the UI. Two
frontend waves have now landed pages (#49–#55 developer surface, #51/#52/#53/#62/#63
admin surface) and the only place the route list exists is `foundrygate-spec.md`
§8.1, which is a design document rather than user-facing reference.

A fork operator opening the deployed app has no page telling them what each route
does, who can reach it, or which API endpoints sit behind it.

## What's needed

`docs-site/src/content/docs/reference/web-ui.md` with a route table:

| Route | Who | What it does |
|---|---|---|
| `/` | Any | Home — signed-in greeting and this month's usage |
| `/me` | Any | Developer account: quota gauge, key display and rotate, request history |
| `/me/request` | Any | Submit a quota increase request |
| `/dashboard` | Admin | Summary cards, top consumers, pending-requests badge |
| `/users` | Admin | Searchable, server-paged user list; row opens the detail page |
| `/users/{id}` | Admin | Overview, Quota, Groups and Keys tabs; activate/deactivate, budget tier, key provision/rotate/revoke |
| `/users/sync` | Admin | Run the Entra user reconcile and read its counts |
| `/groups` | Admin | Group list with member count, budget tier and roster source |
| `/groups/new` | Admin | Create a group, optionally linked to an Entra group |
| `/groups/{id}` | Admin | Policy editor, roster, add/remove members (manual groups only), sync from Entra, delete |
| `/requests` | Any | Admin: every request with a status filter and a review drawer. Developer: their own, read-only |
| `/requests/{id}` | Owner or Admin | The same review panel as a page |
| `/foundry` | Admin | Foundry model deployments; create (OpenAI format) and delete |
| `/config` | Admin | System configuration editor |
| `/audit` | Admin | Audit log with filters |

Worth stating alongside it, because each is a thing an operator will otherwise
discover by surprise:

- Admin routes carry `[Authorize(Roles = "FoundryGate.Admin")]`; an authenticated
  non-admin gets the AccessDenied page, an anonymous caller gets the sign-in redirect.
- Every quota is chosen as a **tier**, never typed as a number
  (`fable-refactor-log.md` D-013) — so the UI never shows a token count where a tier
  name belongs.
[...]

Issue #173, opened 2026-09-02
Serve the Foundry create form a real model/SKU catalogue instead of a hardcoded list


kolatts:
## Problem

The `/foundry` create dialog shipped in #62 offers model names and SKUs from a
hardcoded array in `src/FoundryGate.Web/Shared/FoundryDeploymentDialog.razor`,
seeded from the #60 direction update (`gpt-5.3-codex`, `gpt-5.1-codex-max`,
`gpt-5.1-codex-mini`, `gpt-5-codex`, …). `plans/20-foundry-provisioning.md` itself
hedged on this: *"populated from a hardcoded list of known Azure AI Foundry model
IDs, or fetched from a `GET /foundry/available-models` endpoint if we add it later"*.

A hardcoded list goes stale the week after it ships, and the model catalogue is
exactly the thing that moves fastest. The dialog mitigates it — both fields are
`CoerceValue` autocompletes, so anything typed is accepted and the list is only a
shortcut — but an admin still has no way to see what the account can actually serve
without leaving for the Azure portal.

Same for SKUs: the dialog offers `GlobalStandard`, `DataZoneStandard`, `Standard`,
`ProvisionedManaged` from a constant.

## What's needed

`GET /foundry/available-models` (admin), listing what the configured accounts can
deploy — model name, versions, and the SKUs available for each. ARM exposes this:
`CognitiveServicesAccount.GetModels()` on the account resource returns the
deployable model list with SKU information, which is what the create call needs
anyway.

Then point the dialog's two autocompletes at it and keep coercion on, so a model
Azure lists before the endpoint does is still deployable.

## Acceptance

- [ ] An admin opening the create dialog sees models the selected account can
      actually serve, not a list from a plan document.
- [ ] SKU choices come from the same place.
- [ ] Typing a model the endpoint didn't list still works (coercion stays).

Related: #130 (capacity resize via PATCH), #126 (lifting the Anthropic create/delete
refusal). Found while building #62.

kolatts:
Picking this up in claude/frontend-followups (frontend follow-ups wave: #190, #191, #171, #173, #172).

Issue #171, opened 2026-09-02
Persist last-Entra-sync metadata so /users/sync can show it across sessions


kolatts:
## Problem

Issue #63 (and `plans/17-admin-management-ui.md`) specify that `/users/sync` shows
"the timestamp and result of the last sync run (read from
`SystemConfiguration[\"LastUserSyncAt\"]` and `[\"LastUserSyncResult\"]`)".

Those keys do not exist. `FoundryGate.Domain/Constants/SystemConfigurationKeys.cs`
declares `DefaultMonthlyTokenQuota`, `ApimResourceId`, `ApimGatewayUrl`,
`ApimProductId`, `FoundryResourceId`, `EntraTenantId`, `EntraGroupSyncEnabled` and
`ResetDayOfMonth` — nothing about user sync — and `POST /users/sync` writes only a
`users.synced` audit row plus its `UserSyncResult` response body. Nothing durable
records *when* the last run happened or what it did.

So the `/users/sync` page shipped in #63 shows the result of the run you just
triggered in **this** browser session, and nothing on first load. Reload the page and
the history is gone.

## What's needed

Pick one and make the page read it:

1. **`SystemConfiguration` keys** — `LastUserSyncDate` and `LastUserSyncResult`,
   written inside the same transaction as the sync (the run already commits
   atomically, so this is a free addition). Matches what #63 assumed. Note the
   naming convention is `*Date` for temporal values (CONVENTIONS.md), so
   `LastUserSyncDate`, not `LastUserSyncAt`.
2. **Query the audit log** — `GET /audit?action=users.synced&pageSize=1` already
   returns the last run's actor, timestamp and serialized `details`. No schema
   change, no new write path; the page just needs the call. The audit row's
   `details` would have to carry the counts (check whether it does).

Option 2 is probably right: the audit row is already the durable record of the run,
and a second copy in `SystemConfiguration` is a thing that can disagree with it.

## Acceptance

- [ ] `/users/sync` shows when the last Entra sync ran and what it did, on first
      load, without having run one in the current session.
- [ ] Whatever it reads is written by `POST /users/sync` itself (including runs
      triggered outside the UI).

Found while building #63.

kolatts:
Picking this up in claude/frontend-followups (frontend follow-ups wave: #190, #191, #171, #173, #172).

Issue #130, opened 2026-09-01
Add PATCH /foundry/deployments/{accountName}/{deploymentName}/capacity for live TPM rebalancing (sku.capacity is PATCH-able; never PUT)


kolatts:
From the #60 direction update ("Deployment capacity is PATCH-able at runtime → add a 'resize' action alongside create/delete for live TPM rebalancing between deployments"). #61 shipped list/get/create/delete only; there is deliberately **no** update path there because the only SDK update that touches `model` is a PUT (`CreateOrUpdate`), which must never be pointed at an existing deployment (CLAUDE.md; E-006/E-007).

Capacity is different: `sku.capacity` (thousands of TPM) can be changed in place with a **PATCH** — `CognitiveServicesAccountDeploymentResource.UpdateAsync(WaitUntil, PatchResourceTagsAndSku)` in `Azure.ResourceManager.CognitiveServices` 1.5.2 — without re-sending `model`/`modelProviderData`, so it should be safe for Anthropic deployments too (validate live before trusting that: E-007 saw re-PUTs wedge Claude deployments; a PATCH has not been tried).

## Ask

- `PATCH /api/v1/foundry/deployments/{accountName}/{deploymentName}/capacity` (admin) with `{ capacity: <thousands of TPM> }` → `200` with the updated `FoundryDeploymentResponse`; `404` when absent; `400` when out of `[1, ValidationConstants.FoundryDeploymentMaxCapacity]`.
- Add `ResizeDeploymentAsync(accountName, deploymentName, capacity, ct)` to `IFoundryManagementClient` (PATCH only — never PUT) + the fake, and to `IFoundryDeploymentService` with a new `AuditActions.FoundryDeploymentResized` (`foundry.deployment.resized`, details `{ before, after }`).
- Reference page row in `reference/api.md`; #62's grid gets a capacity editor.
- Live-validate on an **OpenAI** deployment first, then a Claude one, and record the E-00x finding in `fable-refactor-log.md`.

kolatts:
Claiming this as part of a backend sweep (with #194, #153, #180, #181) on `claude/backend-sweep-b`.

PR #176, opened 2026-09-02
feat(web): admin users, groups, requests, Foundry and Entra sync pages

### LINKED ISSUE #51: Build admin user MudDataGrid (/users) and tabbed detail page (/users/{id})

### LINKED ISSUE #52: Build admin group pages with MudDataGrid, MudFab, and Entra sync MudDialog
Part of #17.

`/groups` lists groups with member count, quota, and an Entra-linked indicator. `/groups/new` is a create form. `/groups/{id}` shows group detail, a member table (with remove-member button per row), an "Add Member" search widget (search existing users by name/email), a quota/unlimited editor, and — if `EntraGroupId` is set — a "Sync from Entra" button that calls `POST /groups/sync-entra` and displays the sync result summary inline.

### LINKED ISSUE #53: Build admin requests queue with MudDataGrid and approve/reject MudDrawer panel
Part of #17.

`/requests` shows a tabbed or filtered table: pending, approved, rejected. Clicking a pending row expands an inline review panel with the requester info, current and requested quota, justification text, review notes textarea, and Approve/Reject buttons that call the respective endpoints. On completion, collapse the panel and refresh the row status. Non-admin users see only their own requests in a read-only table showing status, reviewer notes, and timestamps.

### LINKED ISSUE #62: Build admin /foundry page with MudDataGrid, create dialog, and delete confirmation
Part of #60.

Create `Pages/Admin/Foundry/Index.razor` (route `/foundry`, Admin role). `MudDataGrid` shows deployment name, model, version, SKU, capacity, and a provisioning-state `MudChip`. A `MudFab` opens a `MudDialog` create form with `MudSelect` for model name, `MudTextField` for deployment name, `MudSelect` for SKU, and `MudNumericField<int>` for capacity. Per-row delete `MudIconButton` requires `MudDialog` confirmation. Add a `/foundry` `MudNavLink` to the Admin section in `NavMenu.razor`.

### LINKED ISSUE #63: Build /users/sync admin page to trigger Entra bulk sync and view results
Part of #17.

Create `Pages/Admin/Users/Sync.razor` (route `/users/sync`, Admin role). A "Run Entra sync" `MudButton` calls `POST /users/sync` and displays the result `{ added, updated, deactivated }` in a `MudAlert`. Show the timestamp and result of the last sync (read from `SystemConfiguration["LastUserSyncAt"]` and `["LastUserSyncResult"]`). Add a `MudNavLink` for `/users/sync` in the Admin Users section of `NavMenu.razor`. This page was listed in the spec §8.1 but was not captured in any prior issue.

### LINKED ISSUE #169: Web API client drifted from the API surface: wrong verbs on activate/deactivate/approve/reject and the wrong type for GET /users/{id}

### LINKED ISSUE #188: Consolidate QuotaDisplay and TierDisplay — one token format across the UI
## Problem

The two frontend waves landed two overlapping quota-presentation helpers, because they were
written in parallel:

| | `FoundryGate.Web/Services/QuotaDisplay.cs` (#49–#55) | `FoundryGate.Web/Shared/TierDisplay.cs` (#51–#53/#62/#63) |
|---|---|---|
| Gauge colour by percent | `GaugeColor`, `GaugeValue` | — |
| Tier name from a **product id** | `TierDisplayName(string?)` | — |
| Tier name from a **stored quota** | — | `Describe(bool, long?, tiers)` |
| Which tier a stored quota is | — | `MatchProductId(...)` |
| Where a budget came from | — | `LevelLabel`, `LevelColor` |
| Token formatting | `FormatTokens(long?)` → `"5,000,000"` | `FormatTokens(long)` → `"5M tokens"` |

The last row is the one that matters: **the same token count is formatted two different ways in
the same app** depending on which page you are looking at. `/me`'s gauge says `5,000,000` and
`/users`' budget column says `5M tokens`. Neither is wrong; having both is.

The rest is duplication rather than contradiction — two files answering adjacent questions about
the same domain concept, in two different folders (`Services/` and `Shared/`), which is exactly
how a third one gets written next.

## What's needed

One helper, in one place, covering all of it:

- gauge colour and clamped value (from `QuotaDisplay`),
- tier name from a product id **and** from a stored quota (both, they are different lookups),
- resolved-level wording and colour,
- **one** token format, picked deliberately.

On the format: `"5M tokens"` reads better in a dense grid cell and a chip, `"5,000,000"` is right
where the exact number matters (a gauge's "x of y"). If both are wanted, they should be two
explicitly-named methods on one type — `FormatTokensCompact` and `FormatTokensExact`, say — not
the same name meaning different things in two files.

Both helpers are already directly unit-tested (`QuotaDisplayTests`, `TierDisplayTests`), so the
merge is safe to do mechanically; the tests move with the methods.

## Acceptance

- [ ] One type owns quota presentation; the other is gone.
- [ ] A given token count renders identically on `/me`, `/dashboard`, `/users` and `/groups`.
- [ ] The existing tests for both still pass, against the merged type.

Filed from PR #176; the duplication was known and deliberate while both branches were open.

### LINKED ISSUE #195: Flaky: MePageTests.The_rotated_mask_survives_the_next_render fails ~2 in 5 full-suite runs
Found while merging `origin/main` into #187 after [#174](https://github.com/kolatts/foundry-gate/pull/174) landed.

`FoundryGate.Tests.Predeployment.Web.MePageTests.The_rotated_mask_survives_the_next_render` fails intermittently in a full-suite run — **2 of 5 runs** on this machine, on a clean `origin/main` (`7382004`) with no other changes:

```
Assert.Equal() Failure: Strings differ
        ↓ (pos 0)
Expected: "••••••••alue"
Actual:   "rotated-key-value"
        ↑ (pos 0)
  at MePageTests.The_rotated_mask_survives_the_next_render() in MePageTests.cs:line 175
```

It **passes 100% of the time** when the class is run alone (`--filter FullyQualifiedName~MePageTests`, 24/24), so this is an ordering or parallelism interaction with the rest of the suite, not a bug in the assertion itself.

## Reading of the failure

The test clicks rotate → confirm → hide, then asserts the key input shows the *masked* form of the rotated key. Getting the unmasked `rotated-key-value` back means the hide click either did not take effect or was undone by a re-render before the assertion — i.e. the very regression the test's own comment describes ("the component used to write the new value onto its own `[Parameter]`, which the framework owns and overwrites"), surfacing only under some interleaving.

Worth checking, roughly in order of suspicion:

- **Shared mutable state across the Web test classes.** `WebTestContext` / `FakeFoundryGateApiClient` / `DashboardStateService` — if any of them is a static or a singleton reused across bUnit contexts, a neighbouring class's render can win a race with this one's.
- **`FixedZoneTimeProvider`** or anything else registered statically for the Web suite.
- xUnit runs test *classes* in parallel by default. If the Web classes share a collection-level fixture (or an ambient `AsyncLocal`, which CONVENTIONS already warns about for `HttpContextAccessor`), the fix is a collection definition or per-test isolation rather than a retry.

## Ask

- [ ] Reproduce (run the full suite a handful of times; it is not rare)
- [ ] Find the shared state and isolate it — **not** a `[Retry]` or a sleep: an intermittently-red gate on every PR is worse than a slower one
- [ ] Confirm with 10 consecutive green full-suite runs

## Why it matters now

`ci.yml` runs the whole suite on every PR, so this is a ~40% chance of a spurious red on any PR that touches anything. It will be blamed on whatever change happens to be in flight — it was nearly blamed on #187.

kolatts:
Builds the five admin surfaces of the Blazor UI: the user list and detail, the Entra
user sync page, the group pages, the quota-request review queue, and the Foundry
deployment page — and fixes the API client drift that would have broken half of them.

Closes #51
Closes #52
Closes #53
Closes #62
Closes #63
Closes #169
Closes #188
Closes #195

## Routes

Rows and sections added to `reference/web-ui.md`, which PR #174 created — its
developer sections are untouched.

| Route | Who | What it does |
|---|---|---|
| `/users` | Admin | `MudDataGrid` over the server-paged `GET /users`; 300 ms debounced search, active/deactivated filter, tier chip, key indicator. Row click → detail |
| `/users/{id}` | Admin | `MudTabs`: Overview (fields, activate/deactivate), Quota (resolved level, tier, usage, budget editor), Groups (memberships), Keys (masked state, provision/rotate/revoke) |
| `/users/sync` | Admin | Runs `POST /users/sync` and reads its counts, including the two that need explaining |
| `/groups` | Admin | Grid with member count, budget tier and roster source; `MudFab` → create |
| `/groups/new` | Admin | Create form: name, description, optional Entra group id, budget tier |
| `/groups/{id}` | Admin | Policy editor, roster grid, add/remove members, sync from Entra, delete |
| `/requests` | Any | Admin: status filter + end-anchored review `MudDrawer`. Developer: their own, read-only |
| `/requests/{id}` | Owner or Admin | The same review panel as a page |
| `/foundry` | Admin | Deployment grid, create dialog, delete |

## #169 — the client had drifted from the API

Every one of these failed at runtime the first time a page called it, and nothing in
the build noticed:

| Method | Was | Now |
|---|---|---|
| `ActivateUserAsync` | `PUT /users/{id}/activate` | `POST` |
| `DeactivateUserAsync` | `PUT /users/{id}/deactivate` | `POST` |
| `ApproveRequestAsync` | `PUT /requests/{id}/approve` | `POST` |
| `RejectRequestAsync` | `PUT /requests/{id}/reject` | `POST` |
| `GetUserAsync` | deserialized `UserResponse` | `UserDetailResponse` (#156) |
| quota / activate / deactivate | discarded the body | return the updated `UserResponse` |

`FoundryGateApiClientRouteTests` is the tripwire — a recording handler asserts the
verb and path of each corrected method, plus the query strings for the filtered user
and request lists. The next drift fails CI rather than the browser.

- **No `MudNumericField` anywhere a budget is set.** D-013 says a budget *is* a
  gateway tier, so `QuotaTierPicker` is a pick from `GET /quota/tiers` plus an
  unlimited switch, and `TierDisplay` renders every stored quota as its tier's
  display name — `/users` shows "Standard", never "5,000,000". A legacy value
  matching no tier is shown as itself rather than silently mapped, matching the
  API's own `IsGatewayCapped` behaviour.
- **The bUnit harness is PR #174's, byte-for-byte.** `WebTestContext`, `WebTestData`,
  `FakeFoundryGateApiClient`, the `bunit` and WASM package lines, `Program.cs`'s
  `DashboardStateService` registration and the `*.Me` client partials are identical
  to that branch, so those merges are no-ops. The one intended edit is the fake's
  admin section, which #174 left with a comment naming this wave; admin-only
  fixtures live in a separate `AdminTestData` rather than colliding on
  `WebTestData`'s `Request`/`Allocation` names.
- **The client's admin half is its own partial file** (`*.Admin.cs`), matching #174's
  `*.Me.cs`, so the two waves never rewrote one interface. The duplicate
  `GetQuotaTiersAsync` and filtered `GetRequestsAsync` are dropped from mine — #174
  declares both.

Deliberate deviations from `plans/17-admin-management-ui.md`, recorded in the plan:

- Flat `Pages/*.razor` files (replacing the spec §8.1 stubs) rather than a
  `Pages/Admin/<Area>/Index.razor` tree.
- **No per-user quota reset button.** `POST /quota/reset` resets the whole tenant for
  the period; on one person's page it would misrepresent what it does.
- **Membership is edited on the group, not on the user.** A roster is what an Entra
  sync owns, and the API refuses membership edits on a linked group with a 409. The
  user's Groups tab lists and links; it doesn't edit.

## Things the UI says out loud rather than leaving to be discovered

- An Entra-linked group hides the add/remove controls entirely and explains that the
  next sync would undo the change — instead of offering a button the API answers 409 to.
- Deleting a group with members explains what `?force=true` costs before sending it.
- `/foundry` disables delete on Anthropic rows and never offers an Anthropic create;
  Claude deployments belong to infra end to end (#126 would lift that).
- `/users/sync` explains `skippedGroupAssignmentCount` — departure detection is
  suspended while group-based app-role assignments exist (#121) — and treats a
  non-zero `failedCount` as an error, because it means someone still holds a working
  gateway key.
- A gateway key's plaintext is shown once, in a dialog that says so.

## Verification

- `FoundryGate.Tests.Predeployment` — **988 passed, 0 failed** (109 new: 97 bUnit
  component tests, 12 client route tests, and the route-authorization theory)
- `npm run build` in `docs-site/` — 15 pages, clean

Against `plans/17-admin-management-ui.md`'s Verification list:

- [x] `dotnet build` passes — zero warnings
- [x] User table search and server-side pagination work correctly — 300 ms debounce, filter changes reset to page 1 (`UsersPageTests`)
- [x] Status toggle requires MudDialog confirmation before calling the API; cancelling calls nothing (`UserDetailPageTests`)
- [x] Group Entra sync shows the added/removed/skipped-unknown counts in a dialog (`GroupDetailPageTests`)
- [x] Approving a request from the drawer updates the row status without a full grid reload (`RequestsPageTests`)
- [x] Non-admin navigating to `/users` is redirected to AccessDenied — every admin page carries `[Authorize(Roles = RoleNames.Admin)]`, which `App.razor` renders as `AccessDenied` (`AdminPageAuthorizationTests`); a 403 mid-page does the same

Against `plans/20-foundry-provisioning.md`'s #62 items:

- [x] Grid with account, name, model, version, SKU, capacity as thousands of TPM, colour-coded state chip
- [x] Provisioning state chip updates on refresh (explicit refresh button; ARM finishes asynchronously)
- [x] Delete behind confirmation, disabled on Anthropic rows
- [x] Non-admin cannot access `/foundry`
- [ ] Live: creating and deleting through the UI against real Azure — #125

## Follow-ups filed

- **#171** — no durable last-sync metadata exists; #63 assumed `SystemConfiguration["LastUserSyncAt"]`/`["LastUserSyncResult"]` keys that were never added. The page says so and points at the audit log.
- **#173** — the Foundry create dialog's model and SKU lists are hardcoded from the #60 direction update; they need an endpoint listing what an account can serve.
- **#175** — superseded by #174's `reference/web-ui.md`; close it.
- **#188** — `QuotaDisplay` (#174) and `TierDisplay` (this PR) format the same token count two different ways. Known and deliberate while both branches were open; one type should own it.

Existing and untouched: #130 (capacity resize, no API route yet), #126 (Anthropic
create/delete refusal), #83 (backend pool membership, which `/foundry` links to),
#121 (group app-role expansion).


kolatts:
**REQUEST CHANGES**

Nine admin surfaces, 97 bUnit tests that assert behaviour instead of markup, and the #169 client fix with a tripwire behind it — and the "say the consequence before you send it" discipline (force-delete's cost spelled out, the Entra-linked group's 409 explained rather than demonstrated, the key that says it's only shown once) is the thing I'd want copied into every page that comes after this. Blocking on one real bug that puts one person's review notes on someone else's request, and two coordination items with open PRs that are cheap to settle now and expensive to discover later.

Verified on the branch: build 0 warnings, 988 tests pass, `dotnet format` clean, `docs-site` builds 15 pages. I probed four of the PR's claims and all four hold — the tier picker really emits the selected tier's cap, `PUT /users/{id}/quota` really sends `MonthlyTokenQuota = 20,000,000` with `IsUnlimited = false`, `/users` really resets to page 1 when you search from page 2 (unlike #174's audit grid), and an Entra-linked group really hides both roster controls.

**On the harness and merge order.** The shared files are byte-identical to #174's: `WebTestContext.cs`, `WebTestData.cs`, `Program.cs`, `DashboardStateService.cs` and both `*.Me` partials show zero diff against that branch. Only `FakeFoundryGateApiClient.cs` grows, in its own admin section, exactly as advertised. #174's review said land it first and that still stands — but three files here are *not* clean additions on top of it: `FoundryGateApiClient.cs` (−10/+10), `IFoundryGateApiClient.cs` (+21/−7) and `_Imports.razor`. Those ten changed client lines **are** #169. Resolve that conflict by taking one side wholesale and you quietly put `PUT /users/{id}/activate` back, with nothing failing. Same care on `web-ui.md`: this branch appends to #174's copy, so #174's fix to line 7 ("Every route requires an authenticated caller" — not true of `/`, which this file's own table lists as `Any`) has to survive the rebase rather than be overwritten by it.

---

**Issues**

- **Major** `Shared/RequestReviewPanel.razor:144-150` — `OnParametersSet` clears `_reviewNotes` only when the incoming request is *not* pending, so draft notes survive a hop from one pending request to the next. Reproduced it: open request 5, type "Only for request 5.", dismiss the temporary drawer by clicking the scrim (which sets `_drawerOpen = false` but never nulls `_selected`, so the panel is never unmounted), open request 6, press Approve — request 6 is approved with request 5's notes, and the requester reads them. The comment on line 146 says this is exactly what the guard is for. `@key="selected.QuotaIncreaseRequestId"` on the panel, or clearing on any id change, closes it; `/requests/{id}` → `/requests/{id}` has the same shape.

- **Major** `Pages/UsersSync.razor:67` (and the matching `web-ui.md` section) — the `skippedGroupAssignmentCount` explainer says group-based access "isn't expanded to its members yet… assign developers to the application individually if you need departures caught". PR #186 is open and inverts that: expansion works, and a non-zero count is the groups whose expansion **failed** — a Graph permission problem. Whichever of you merges second owns rewriting this, because as written it tells an admin to restructure their tenant to fix a missing role. Worth agreeing who, before either lands.

- **Major** — nothing in the UI calls `POST /groups/sync-entra`. `SyncAllGroupsFromEntraAsync` ships in the client, #52 names that route, and #52 carries an owner comment saying the result dialog must render `succeeded: false` rows distinctly with their `error` text (arriving in #186). This PR closes #52 with the sync-all surface absent and no issue filed — CLAUDE.md's "the gap becomes an issue, not a silent substitution". The per-group route is the right call for a group detail page; the bulk one still needs somewhere to live, even if that somewhere is an issue.

- **Minor** `Pages/UserDetail.razor:314`, `Pages/GroupDetail.razor:288` — the fake captures `QuotaUpdates` and `GroupUpdates` and nothing asserts on either, so no test covers "the budget editor writes a tier cap". I wrote the missing one and it passes — but D-013 is the load-bearing claim of this PR, and #174's review found a real bug of exactly this shape in the equivalent developer form.

- **Minor** `Tests/Web/Pages/UsersPageTests.cs:70` — `A_search_starts_again_from_page_one` never leaves page 1, so it passes whether or not the reset exists. Drive the grid to page 2 first and it becomes the test the plan file claims it is.

- **Minor** `Pages/Users.razor:190-191` (same at `Groups.razor:143`, `Requests.razor:230`) — `NavigateTo(Page.First)` reloads server data itself when `ServerData` is set, so the explicit `ReloadServerData()` after it fires a second identical request. Measured it: from page 1 a search is one fetch, from page 2 it's two. Cheap, but it's two responses racing to render the same grid.

- **Minor** `Shared/RequestReviewPanel.razor:180` — `ReviewedDate = DateTimeOffset.UtcNow` is the only naked `UtcNow` in the diff (CONVENTIONS: "No naked `UtcNow` outside the interceptor's `TimeProvider`"), and the panel then renders it as "Reviewed \<time\>" — a timestamp the browser invented. `/requests/{id}` re-reads immediately so it never shows there; the drawer keeps it until the next page fetch. Leaving `ReviewedDate` null falls through to the existing "Reviewed." wording for free.

- **Minor** `Shared/RequestReviewPanel.razor:157` — `api.md` says two reviewers deciding at once means one gets a `409`. That arrives as a snackbar and nothing re-reads, so the loser's drawer still says Pending with both buttons live and they can press again. Re-reading the request on a failed verdict shows them what actually happened.

- **Minor** `Pages/UsersSync.razor:29`, `Pages/GroupDetail.razor:95` — every destructive action in this PR is behind `ConfirmDialog` except the two syncs, and those are the ones that deactivate accounts, delete APIM subscriptions and drop group members. The prose above each button explains it well; a single unconfirmed click under that prose is still the odd one out.

- **Minor** `Pages/Users.razor:32`, `Groups.razor:24`, `Requests.razor:28` — `RowClick` is the only way into `/users/{id}`, `/groups/{id}` and the review drawer, and a `<tr>` isn't focusable and carries no role, so three admin grids are mouse-only. `GroupDetail`'s member grid already has the fix — an `OpenInNew` `MudIconButton` per row.

- **Minor** seven pages open with the same six-line `GetQuotaTiersAsync` + unwrap block (`Users:129`, `UserDetail:244`, `Groups:101`, `GroupNew:72`, `GroupDetail:211`, `Requests:148`, `RequestDetail:62`), refetching a static catalogue on every navigation between them. `ConfirmAsync` is duplicated character-for-character in `UserDetail:380` and `GroupDetail:399` and inlined a third time in `Foundry.DeleteAsync:197`, and the `ServerData` Forbidden/error preamble is copy-pasted three times, comment included. One scoped tier cache and two small helpers, and the nine pages get materially shorter.

- **Minor** #188, since you asked which should win: `TierDisplay`. It resolves a tier name from the live `GET /quota/tiers` catalogue, which is the D-013-correct answer; `QuotaDisplay`'s hard-coded switch is the half that breaks the first time a fork adds a tier product. Fold `GaugeColor`/`GaugeValue`/`TierDisplayName(productId)` into it, keep it in `Services/` (it's a helper, not a component), and name the two formats `FormatTokensCompact` and `FormatTokensExact` so nobody has to guess which one they got.

- **Nit** `Tests/Web/FoundryGateApiClientRouteTests.cs` — twelve tests covering the four corrected methods and a few neighbours; rotate, revoke, create-deployment, add/remove-member and both syncs are still unguarded, and every assertion is against `api.md`'s prose rather than the API. The test project already references `FoundryGate.Api` — asserting each client route resolves against the real `EndpointDataSource` would make this an actual tripwire instead of a second copy of the docs.

- **Nit** `Shared/QuotaTierPicker.razor:83-93` — toggling Unlimited on nulls the cap and toggling it back off never restores it, so an accidental on/off leaves the picker on "Inherited" with Save enabled. It's visible on screen, so nobody loses a budget silently, but the switch isn't a toggle today.

- **Nit** `Pages/Requests.razor:220` — the page reads `?status=` but changing the dropdown doesn't write it back, so the filter you're looking at isn't the one in your URL. One `NavigateTo(..., replace: true)` and the deep link works both directions.

- **Nit** `Pages/RequestDetail.razor:95` — `OnReviewedAsync` assigns the optimistic record, calls `StateHasChanged`, then immediately re-reads the whole request. Only the drawer needs the optimistic hop.

- **Nit** `plans/17-admin-management-ui.md:78` — "Live: the pages against a real API + Entra tenant (manual, with #132's checklist)" points at an issue about the APIM key lifecycle and Key Vault wrapping; nothing in it covers these five pages. Either the right issue or a new one, per CLAUDE.md.

- **Nit** `Shared/GroupSyncResultDialog.razor` — correct today, because the per-group route never returns `succeeded: false`. After #186 the record carries `Succeeded`/`Error` regardless; one line in the header comment saying "this is the single-group route, where a failure is the HTTP status" saves the next person from wiring up a branch that can't fire.

kolatts:
---

## Review fix pass

**Major 1 — review notes crossing requests.** Reproduced exactly as described. Notes
were keyed on "is this request pending" rather than "is this the request I was typed
against", and the drawer's scrim closed it without clearing `_selected`, so the panel
stayed mounted. Now: notes carry the request id they belong to and reset on any change
of it, every close path clears the selection, and the panel is `@key`'d on the id.
`Draft_review_notes_never_travel_from_one_request_to_the_next` is the reproduction,
scrim dismissal included.

**Major 2 — `/users/sync` copy.** #186 landed first and inverted the meaning: expansion
works, and a non-zero `skippedGroupAssignmentCount` is the groups whose expansion
*failed*. The old copy told an admin to restructure their tenant to fix a missing Graph
role. Rewritten in the page and in `web-ui.md` to name the real cause and the real fix,
pointing at #120 and #183. The test asserts the new wording and that the old advice is
gone.

**Major 3 — `POST /groups/sync-entra`.** `/groups` gains "Sync all Entra-linked groups"
behind a confirmation. `GroupSyncSummaryDialog` renders a row per group and separates
the two failure kinds #186 added, because they need different things from a person:
`GraphRead` says nothing was applied and re-running is safe; `PostCommit` says the
gateway moved and the database did not, and this group's reported state is not to be
trusted until someone looks. The per-group sync and the bulk user sync are confirmed
too — they were the only unconfirmed actions left, and the ones that delete
subscriptions.

**Minors.** Budget-editor writes are asserted on both pages (D-013 is this wave's
load-bearing claim and nothing covered it). The page-reset test now drives the grid to
page 2 first, and a sibling test pins the single fetch. `ReloadFromFirstPageAsync`
replaces `NavigateTo` + `ReloadServerData`, so a filter change no longer races two
identical responses. A 409 from a concurrent reviewer re-reads the request and replaces
the panel. Users and Groups rows carry a keyboard-reachable link. Seven copies of the
tier fetch collapse into a scoped `QuotaTierCatalog`; `ConfirmAsync` and the
`ServerData` preamble into two helpers in `AdminUiExtensions`.

On the naked `UtcNow`: I removed it rather than injecting `TimeProvider`. The value fed
"Reviewed {time}", so any client-side instant is a fabrication — a testable fabrication
is still one. Leaving `ReviewedDate` alone falls through to the plain "Reviewed."
wording until the next read supplies the server's. Happy to swap to `TimeProvider` if
you'd rather have the convention applied literally.

**Nits.** The route tripwire now resolves every client route against the API's own
routing table via `ApiTestFactory` rather than a second copy of `api.md`, and a
companion test fails if a mutating client method is added without a row. Unlimited
restores the cap it replaced. The status filter writes `?status=` back. `plans/17`
points at #192 (a new live-validation checklist for these pages) instead of #132.

## #188 — TierDisplay wins

Applied your ruling. `TierDisplay` moved to `Services/`, absorbed `GaugeColor`,
`GaugeValue`, `TierDisplayName` and `LevelSentence`, and the two formats are now
`FormatTokensCompact` ("5M tokens") and `FormatTokensExact` ("5,000,000").
`QuotaDisplay` is deleted and its tests moved and re-pointed. `TierDisplayName` keeps
the title-case fallback but prefers the catalogue's own `DisplayName` when it has it,
so a fork's own tier product renders its real name.

## #195 — the flaky test

Not shared state. Three tests asserted synchronously on state that only settles after
an asynchronous hop, so parallel classes saturating the thread pool changed the odds
rather than the bug.

The real defect was in `KeyDisplay.SetRevealedAsync`: it mutated `_revealedKey` and
then awaited `RevealedKeyChanged`, and `ComponentBase` renders only once the handler's
Task completes — so **"Hide" left the plaintext key on screen until the parent had
finished re-rendering**. Under load that gap is visible. It now calls
`StateHasChanged()` on its own change before the round trip. A hidden key that is still
readable for a frame is a worse bug than the flaky test that found it.

`Stops_refreshing_once_the_page_is_gone` sampled the call count once and again 200 ms
later, which raced any tick already in flight. It now waits for the count to stop
moving and then requires it to stay still — testing "the loop stopped" rather than "the
loop was slower than 200 ms".

Verified with 8 consecutive full-suite runs and 10 web-only runs, all green; the same
loop reproduced the failure twice in five before the fix.

## Rebase

Took main's harness wholesale — `WebTestContext`, `WebTestData` and
`FakeFoundryGateApiClient` are main's, with only the admin section re-applied (in
`FakeFoundryGateApiClient.Admin.cs`, plus the five signature corrections #169 forces on
the existing user methods). `web-ui.md` is main's corrected copy with my sections
appended; your line-7 fix survived, and "Pages still being built" is gone because they
aren't. `FoundryGateApiClient.cs` / `IFoundryGateApiClient.cs` were resolved
line-by-line and the route tripwire re-run — activate and deactivate are still POST.

## Gates

- `FoundryGate.Tests.Predeployment` — **1173 passed, 0 failed**
- `npm run build` in `docs-site/` — 15 pages, clean

Issue #38, opened 2026-06-04
Implement monthly quota reset as an Azure Function Timer Trigger


kolatts:
Part of #10.

Implement a `MonthlyResetBackgroundService : BackgroundService` that wakes at 00:01 UTC on the 1st of each month. For each active user it calls `QuotaResolutionService` to produce the new allocation and INSERTs a `QuotaAllocation` row for the new period (idempotent — skip if row already exists). Sets `IsHardStopped = false`. After processing all users, writes a single `quota.monthly-reset` audit log entry with the count. Use a distributed lock or idempotency check so simultaneous Container App replicas do not double-reset.

kolatts:
Claimed by an implementation agent on `claude/functions-38-39-84-119` (Functions wave: #38 + #39 + #84 + #119, with #165's ResetDayOfMonth gate).

Issue #182, opened 2026-09-02
Finish the CommitToken sweep: GroupService and EntraGroupSyncService still hand-roll the ternary


kolatts:
Follow-up from the #163/#158 sweep (keys/lifecycle/quota hardening wave).

`FoundryGate.Api/Services/CommitToken.For(reachedExternal, cancellationToken)` is now the one place the conditional commit point is expressed, and CONVENTIONS.md names it. Every caller in `Services/{Quota,Users,Lifecycle,Requests}` uses it.

Two do not, because they were being edited concurrently by another agent when the sweep landed and touching them would have caused a pointless conflict:

- `Services/Groups/GroupService.cs` — has its own `private static CancellationToken CommitToken(bool gatewayMoved, CancellationToken)` helper, which is the same function under a name that now shadows the shared type.
- `Services/Groups/EntraGroupSyncService.cs` — an inline `resolutions.Any(r => r.TierSyncRequested) ? CancellationToken.None : cancellationToken`.

Both are **correct today** (verified while sweeping; #152 fixed them). This is tidy-up, not a bug.

## Ask

- Replace both with `CommitToken.For(...)` and delete `GroupService`'s private helper — the local name shadowing the shared class is the part most likely to confuse the next reader.
- While there, check the read-backs: the sweep found two services (`QuotaAllocationService.GetMyAllocationAsync`, `QuotaRequestService.ApproveAsync`) that saved on `CancellationToken.None` and then re-read the committed row on the request token, so a disconnected client got an error for a change that had landed. Confirm neither group path does the same.

kolatts:
Update from the #184 review, which widens this slightly.

`CommitToken` has **moved out of the Api**: it is now `FoundryGate.Data/Concurrency/CommitToken.cs` (namespace `FoundryGate.Data.Concurrency`). The rule is about which token reaches `SaveChangesAsync`, and Data is what every host that saves already references — including `FoundryGate.Core`, which needs Data for the quota reset. So this is no longer just tidy-up in `Services/Groups`:

- **#187 must adopt it, not re-roll it.** `QuotaResetService` lands in `FoundryGate.Core`, which cannot reference the Api — that was the original reason the helper could not stay there. `using FoundryGate.Data.Concurrency;` and `CommitToken.For(reachedExternal, cancellationToken)`; do not reintroduce the ternary CONVENTIONS.md now forbids.
[...]

Issue #181, opened 2026-09-02
Make the /keys/me rate limits configurable rather than constants in the Api


kolatts:
Follow-up from the #136 fix (keys/lifecycle/quota hardening wave).

The limits landed as private constants in `FoundryGate.Api/Extensions/RateLimiterExtensions.cs`: 5 reveals and 3 rotations per minute per caller `oid`, in a one-minute fixed window. #136 itself said the numbers are a product call, and a fork whose developers work differently has to recompile to change them.

## Ask

- Move them to the options pattern (`AppSettings`, DataAnnotations, `ValidateRecursively` at startup, per CONVENTIONS.md) — permit counts and window, per policy.
- Keep the current values as the defaults, so a fork that configures nothing behaves exactly as it does today.
- The two constants that mirror them in `Api/Endpoints/KeysRateLimitEndpointTests` should then read the configured values instead of restating them.
- Worth doing together with, or after, the config work in #164/#123 so the keys settings land in the same shape as the rest.

kolatts:
One more reason to do this, from the #184 review.

`Api/Endpoints/KeysRateLimitEndpointTests` currently reads the **wall clock**. `System.Threading.RateLimiting`'s built-in limiters own their replenishment timer, and neither `FixedWindowRateLimiterOptions` nor the rate-limiting middleware accepts a `TimeProvider` — so `ApiTestFactory.TimeProvider` cannot move the limiter, and the tests can only stay inside a real 60-second window rather than step over it. Fine today (a handful of in-process requests against an in-memory host), latent flake later.

Making the window and permit counts configuration fixes both halves at once:

- the tests can configure a short window through `IWebHostBuilder.UseSetting` instead of racing a real one, and
- they can read the configured values instead of mirroring them in two `private const`s that have to be kept in step with `RateLimiterExtensions` by hand.

So this is worth a little more than "a fork might want different numbers".

kolatts:
Claiming this as part of a backend sweep (with #194, #153, #180, #130) on `claude/backend-sweep-b`.

Issue #180, opened 2026-09-02
Anomaly signal on repeated key reveals (N key.revealed rows in a window -> warning)


kolatts:
Follow-up from the #136 fix (keys/lifecycle/quota hardening wave).

#136 asked for two things. The limiter shipped: `POST /keys/me/reveal` is capped at 5 per minute per caller `oid` and `POST /keys/me/rotate` at 3, both returning a ProblemDetails `429` with `Retry-After`.

The second half did not:

> Consider an anomaly signal on top: N `key.revealed` rows for one user inside the window -> warning log / future alert (#84 dashboards).

A limiter stops a *fast* drain; it does nothing about a patient one. Four reveals a minute, every minute, is well inside the cap and is not something a human does.

## Ask

- Count `key.revealed` rows per user over a rolling window (the audit table already has the rows and the index) and log at Warning past a threshold, with the user id and the count.
- Decide where it runs: cheapest is inside `ApimKeyService.RevealAsync`, which already writes the row; a `FoundryGate.Functions` timer over the audit table is the version that also catches a slow drain across days.
- Feed it into #84's dashboards rather than inventing a second alerting path.

kolatts:
Claiming this as part of a backend sweep (with #194, #153, #181, #130) on `claude/backend-sweep-b`.

PR #184, opened 2026-09-02
fix(api): hold the commit point across every accepted external change, and close the key/first-login gaps

### LINKED ISSUE #136: Rate-limit POST /keys/me/reveal (and /rotate) per user — a leaked bearer token can pull the plaintext key indefinitely
Raised in the #133 review (keys wave, #36/#37/#95).

`POST /api/v1/keys/me/reveal` decrypts and returns the caller's full APIM key. It is audited (`key.revealed`) but not throttled: a leaked bearer token can pull the plaintext indefinitely and the only trace is a growing run of `key.revealed` rows. The same applies, less acutely, to `POST /keys/me/rotate` (each call mints a fresh key, so a token holder can churn credentials).

## Ask

- Add a per-user rate limit to the `/keys/me/reveal` (and `/keys/me/rotate`) actions with ASP.NET Core's built-in rate limiter (`AddRateLimiter` + `[EnableRateLimiting("...")]`), **partitioned on the caller's `oid`** (via `ICurrentUserAccessor.EntraObjectId` or the claim directly), not on IP — the UI sits behind a shared egress and admins share addresses.
- Something like a fixed window of a handful of reveals per 10 minutes per user; 429 with `Retry-After`. Numbers are a product call — the UI reveals once per page load at most.
- Consider an anomaly signal on top: N `key.revealed` rows for one user inside the window → warning log / future alert (#84 dashboards).
- Integration test through `ApiTestFactory` (the limiter partitions on the test `oid` header, so it is hermetic).

Not implemented in #133 to keep that PR to the lifecycle semantics; `KeysController.RevealMineAsync` references this issue.

### LINKED ISSUE #154: GET /users/me: absorb the first-login race instead of returning 409

### LINKED ISSUE #158: Sweep: audit + save after an external side effect must run on CancellationToken.None
CONVENTIONS.md, "External side effects have a commit point":

> once the external system has accepted the change, the audit row and `SaveChangesAsync` run on
> `CancellationToken.None` — a client that disconnects mid-call must not turn an accepted change into
> an unaudited one.

Several services still thread the request's `CancellationToken` all the way through the post-external
audit and save, so a client that hangs up between "Azure accepted it" and "we wrote it down" leaves the
database disagreeing with the world:

- `Services/Quota/QuotaAllocationService.ResetAsync` — `IQuotaResolutionService.ResolveManyAsync` may
  have moved subscriptions between tier products via `IGatewayTierSync` (live once #118 lands); the
  `audit.LogAsync` and both `SaveChangesAsync` calls after it use the request token.
- `Services/Keys/ApimKeyService` — `ProvisionAsync`, `RotateMineAsync`/`RotateForUserAsync` and
  `RevokeForUserAsync` all audit and save on the request token after APIM has already created,
  regenerated or deleted a subscription. Rotation is the worst case: the old key is dead the moment
  APIM returns, so an abandoned save leaves a user holding a key nobody can decrypt.
- `Services/Foundry/FoundryDeploymentService` already gets this right (`AuditAfterCommitAsync`) and is
  the shape to copy.

`Services/Requests/QuotaRequestService.ApproveAsync` was fixed in the #34/#35 PR (#148) — its audit row,
save and commit run on `CancellationToken.None` past the `ResolveAsync` that may move the gateway tier —
but doing the same sweep one service at a time is how this drifts, so it is worth one pass.

**Scope:** for each call site, identify the external commit point, switch the audit + `SaveChangesAsync`
(+ `CommitAsync` where a transaction is open) after it to `CancellationToken.None`, and log the residual
"save failed after the external system accepted" orphan at Error with the change's full identity, as
`FoundryDeploymentService` does. A test per service that cancels the token after the fake external
client returns and asserts the audit row still lands would pin it.

Found during review of #148.

### LINKED ISSUE #163: Sweep the commit-point rule across every quota-resolution caller (CancellationToken.None after a tier move)
Raised in the review of #152.

CONVENTIONS.md's commit-point rule is: once an external system has accepted a change, resolve-the-actor and every refusal must already have happened, and the audit row plus `SaveChangesAsync` run on `CancellationToken.None` — a client that disconnects mid-call must not turn an accepted change into an unaudited one. `FoundryDeploymentService.AuditAfterCommitAsync` is the reference implementation.

Quota resolution is an external side effect under that rule: `QuotaResolutionService.ResolveCoreAsync` calls `IGatewayTierSync.SyncAsync`, which (once #118's `ApimGatewayTierSync` lands) moves a developer's APIM subscription between tier products. Every caller of `ResolveAsync`/`ResolveManyAsync` therefore has a commit point, and not all of them observe it.

## Audited in #152

`GroupService` and `EntraGroupSyncService` were fixed in that PR: the actor is resolved and every refusal made before `ResolveManyAsync`, and the audit row and save run on `CancellationToken.None` when the resolution actually requested a tier sync (`QuotaResolution.TierSyncRequested`, which is the precise signal — an empty member list or an unchanged tier is not a commit point).

## Still to check

- **`QuotaAllocationService.ResetAsync`** — same shape: `ResolveManyAsync` over every active user, then `audit.LogAsync(...)` and `SaveChangesAsync` on the request token. A `POST /quota/reset` that the admin's browser abandons part-way could move a large number of APIM subscriptions with no audit row and no allocation rows.
- **`QuotaAllocationService.GetMyAllocationAsync`** — `ResolveAsync` then `SaveChangesAsync` on the request token, on the developer's own first call of the month.
- **#156's approval path** (`PUT /requests/{id}/approve`) once it lands, and anything else that grows a `ResolveManyAsync` call.

## Ask

Decide whether this belongs to `IQuotaResolutionService` itself rather than each caller — for example, having resolution report its commit point (it already returns `TierSyncRequested`) and having a small shared helper own the "audit + save on `None`" tail, so a new caller cannot forget. Then apply it to the sites above.

Latent until #118 replaces `NullGatewayTierSync`, so this is not urgent — but it should land in the same wave as the real tier sync rather than after it.

### LINKED ISSUE #167: Add User.LastLoginDate so "has this account been used?" has an honest column
Discovered in the #156 review.

`GET /users/me` used to stamp `User.LastSyncedDate` on every call, which meant two things went wrong at once: every profile load was a row `UPDATE` on `Users` (the table the Entra sync also writes to), and the column stopped meaning what `UserContracts` documents — *"When this user was last touched by an Entra sync"*.

The fix in #156 was to stop stamping it: `/users/me` now writes only when a token claim actually differs from the stored value, and `LastSyncedDate` is left to `POST /users/sync`. That restores the column's meaning but loses a genuinely useful signal — **"has this account been used?"** — which an admin wants for offboarding review, licence cleanup and spotting provisioned-but-never-active developers.

## Ask

- Add `User.LastLoginDate` (`DateTimeOffset?`, nullable — "never signed in" is a real and interesting state) plus the column in `FoundryGate.Database/dbo/Tables/Users.sql`; `SchemaParityTests` covers the rest.
- Stamp it in `UserService.GetMyProfileAsync` (and on first-login provisioning, where it equals `CreatedDate`).
- Decide the write policy before implementing: a naive stamp reintroduces the "every profile load is an UPDATE" problem this issue exists because of. Options: write only when the stored value is older than some granularity (a day is plenty for offboarding decisions), or write it from a background path rather than the request. Whichever is chosen, say so in the property's XML doc.
- Surface it on `UserResponse` / `UserDetailResponse` and in `reference/api.md`'s users table, and consider it for the admin list's default columns.
- Worth pairing with #122's `accountEnabled` decision: "disabled in Entra" and "never logged in" are the two signals an offboarding sweep actually wants.

### LINKED ISSUE #168: ApimKeyService.RotateAsync: a cancelled request after APIM regenerates leaves a dead key, uncompensated
Found while fixing the commit-point Major on #156.

`ApimKeyService.ProvisionAsync` and `RevokeCoreAsync` now run everything after APIM has accepted a change on `CancellationToken.None` (CONVENTIONS.md: *"once the external system has accepted the change, the audit row and `SaveChangesAsync` run on `CancellationToken.None`"*). `RotateAsync` was left alone because no lifecycle pipeline calls it — but it has the same gap, and a worse failure mode.

```csharp
await apim.RegeneratePrimaryKeyAsync(subscriptionName, cancellationToken);
await apim.RegenerateSecondaryKeyAsync(subscriptionName, cancellationToken);
...
keys = await apim.ListSecretsAsync(subscriptionName, cancellationToken);
await StoreKeyAsync(user, user.ApimSubscriptionId, keys.PrimaryKey, issuedDate, cancellationToken);
_ = await audit.LogAsync(AuditActions.KeyRotated, …, cancellationToken);
await dbContext.SaveChangesAsync(cancellationToken);
```

Once `RegeneratePrimaryKeyAsync` returns, the developer's old key is **already dead at the gateway**. If the caller disconnects in the window that follows, the new key is never stored: the row keeps ciphertext for a key APIM no longer recognises, so `POST /keys/me/reveal` hands back a dead credential and the developer is locked out with no signal.

Worse, the compensation that exists for exactly this — `RecordRotationFailureAsync`, which restores the previous values, drops the unsaved `key.rotated` row and writes `key.rotation-failed` — is guarded by:

```csharp
catch (Exception exception) when (exception is not OperationCanceledException)
```

so a cancellation is the one failure that skips it entirely. Nothing is logged, nothing is audited.

## Ask

- Run the post-regeneration section (`ListSecretsAsync`, `StoreKeyAsync`, the audit row, the save) on `CancellationToken.None`, matching `ProvisionAsync`/`RevokeCoreAsync`.
- Then re-examine the `is not OperationCanceledException` guard: with the token gone the case should not arise, but if it does, silently skipping the compensation is the wrong default — it should log at Error like the other arms.
- Pin it with the same probe shape #156 used: a fake APIM whose `AfterMutation` hook cancels the request token, then assert the stored ciphertext round-trips to the key APIM currently holds and a `key.rotated` row exists.
- Applies to `POST /keys/me/rotate` and `POST /keys/{userId}/rotate`.

kolatts:
Six issues from the keys / lifecycle / quota review backlog, all of them the same question from different angles: **what does the database owe the world once an external system has already said yes** — plus the two developer-facing gaps that fell out of reading the same code.

Merged `origin/main` (so #186 is in). Baseline on this branch's merge base is 1029 predeployment tests; this leaves **1033 green**, zero build warnings, `dotnet format` clean, docs build passing.

---

## `CommitToken.For` — one name for the conditional commit point (#163, #158)

Five services had five spellings of the same ternary deciding whether to finish on `CancellationToken.None`, and nothing stopping the sixth from forgetting it. `FoundryGate.Data/Concurrency/CommitToken.For(reachedExternal, cancellationToken)` is now the single expression of the rule, CONVENTIONS.md names it, and the predicate is stated explicitly: **"we reached the external system"** (`QuotaResolution.TierSyncRequested`), not "we called something that might have" — an empty member list or an unchanged tier is not a commit point.

It lives in **Data**, not the Api, so #187's `QuotaResetService` in `FoundryGate.Core` can use it instead of re-rolling the ternary CONVENTIONS now forbids. The rule is about which token reaches `SaveChangesAsync`, and every host that saves references Data.

Sweeping the call sites turned up one real bug, in two places. `QuotaAllocationService.GetMyAllocationAsync` and `QuotaRequestService.ApproveAsync` both save on `None` and then **re-read the committed row on the request's token** to build the response — so a client that disconnected got an error for a change that had actually landed. Both read-backs now use the commit token, and CONVENTIONS says read-backs count.

`QuotaRequestService.ApproveAsync` also finished unconditionally on `None`. That is a **real behaviour change from #148**, and the intended one: when resolution never reached the gateway there is nothing to protect, so an abandoned approval rolls back with its transaction instead of committing a decision nobody is waiting for.

`GroupService`/`EntraGroupSyncService` were verified correct but left alone — #187 and another wave were editing those files — tracked in **#182**.

## Key rotation past the commit point (#168)

The worst version of the gap. A developer's old key is dead the instant APIM returns from `RegeneratePrimaryKeyAsync`, but everything after it ran on the request's token — so a client that hung up in that window left the row holding ciphertext for a key the gateway no longer recognises. Reveal then handed back a dead credential and the developer was locked out with no signal.

And the compensation written for exactly that case (`RecordRotationFailureAsync`) was guarded by `when (exception is not OperationCanceledException)`, so a cancellation was the **one** failure that skipped it entirely.

- The commit point is the primary regeneration; everything after it runs on `CancellationToken.None`.
- The guard is gone, so compensation always runs.
- **Cancellation *during* the primary regeneration** (review catch) — the longest call in the method and the last on the request's token — now takes the same path, recorded as `regenerationConfirmed: false`, because "possibly stale" is what we actually know. There is nothing to restore, but it must not be silent.
- **The secondary regeneration moved after the store** (review catch). It used to sit between the commit point and the store, so an ARM blip on a key nobody has ever been issued restored the stale ciphertext and cost the developer their access. Now the new primary goes on the row first; a secondary failure is logged at Error and named on the `key.rotated` row (`keysRegenerated: ["primary"]`, `secondaryRotationError`) rather than failing the rotation.
- A subscription that vanishes right after the regeneration still gets the documented `409`, not a bare 500.

## First login: two tabs are not an error (#154)

Both requests find no `User` row, both provision, and the loser's insert hits the unique index on `EntraObjectId`. The pipeline rolled back cleanly, but the caller got a **409 on the very first request their UI ever makes**. `GET /users/me` now clears the change tracker, re-reads and returns the winner's profile with a `200`.

Only that race is absorbed. `UserLifecycleService` decides whether it lost one by asking the database who holds the oid — provider-agnostic, no error-number sniffing — so any other `DbUpdateException` keeps its own status, and the "first-login provisioning is not repeatable" conflict (no inner exception) is still a `409`.

**The lock window is unchanged**, and that is a decision worth a second look: the loser still waits out the winner's ARM round trip, because the provision pipeline holds its transaction across it. Shortening it means committing the `User` row before APIM is touched, which trades away the documented, test-asserted invariant *"a failed first login leaves no `User` row"*. Filed as **#179**, with the measurement to do first.

## `User.LastLoginDate` (#167)

Nullable, because "never signed in" is a real state for an offboarding sweep. Stamped by first-login provisioning (so it equals `CreatedDate` for a new row) and re-stamped by `GET /users/me` **only when it is already older than 15 minutes** — a naive stamp would make every profile load an `UPDATE` on the hottest table in the app, which is the exact problem that made `LastSyncedDate` dishonest and created this issue. Entity + `Users.sql` + `UserResponse` + docs.

## Rate limits on `/keys/me/reveal` and `/rotate` (#136)

Reveal returns the caller's plaintext gateway key. It was audited but not throttled, so a leaked bearer token could pull it indefinitely and the only trace was a growing run of `key.revealed` rows.

- Two fixed-window policies partitioned on the caller's `oid` — **not on IP**, per the issue: the UI sits behind a shared egress. 5 reveals and 3 rotations per minute.
- Rejection is a ProblemDetails `429` with `Retry-After`, and now logs at Information with the policy and partition key — the caller is told "tell an administrator"; that line is what tells the administrator, and it is what **#180** builds on.
- **Anonymous callers are not limited at all** (review catch). The global authorization is an MVC `AuthorizeFilter`, not endpoint metadata, so `UseRateLimiter` runs first: a single shared "anonymous" partition let one scanner turn everyone else's `401` into a `429`. They get `GetNoLimiter` — the request reaches MVC, is refused with a `401`, and touches no key material, database or gateway.
- The **admin** routes stay uncapped: an admin rotating a compromised team's keys is the traffic a cap would obstruct.
- `Program.cs` grows exactly two lines; everything else is `Api/Extensions/RateLimiterExtensions.cs` and `Domain/Constants/RateLimitPolicyNames.cs`.

---

## Tests

Probe-style, each reproducing its bug against the fake external system:

| What | Where |
|---|---|
| Rotate + disconnect the instant APIM regenerates → row holds the live key, `key.rotated` audited | `ApimKeyServiceTests` |
| Rotate cancelled *while* ARM regenerates the primary → `key.rotation-failed` marked unconfirmed, Error logged | `ApimKeyServiceTests` |
| Rotate whose failure *is* a cancellation → previous ciphertext restored, `key.rotation-failed` written | `ApimKeyServiceTests` |
| Secondary regeneration fails → developer keeps a working, revealable primary; stale secondary logged and audited | `ApimKeyServiceTests` |
| Subscription vanishes right after the regeneration → still a `409`, still compensated | `ApimKeyServiceTests` |
| Reset / `/me` allocation / approval + disconnect the instant a tier moves → rows and audit still commit | `QuotaAllocationServiceTests`, `QuotaRequestServiceTests` |
| An abandoned reset that never reached the gateway writes nothing | `QuotaAllocationServiceTests` |
| `PUT /users/{id}/quota` + disconnect as the gateway moves | `UserServiceTests` |
| First-login race → winner's profile, one row, complete quota | `UserServiceTests` |
| A first-login insert that fails on a *different* constraint is not absorbed | `UserServiceTests` |
| `LastLoginDate`: stamped, not rewritten inside the window, restamped past it, `null` for never | `UserServiceTests` |
| Rate limits: the caps, the `429` body and `Retry-After`, per-user isolation, anonymous callers keep their `401`, admin route uncapped | `KeysRateLimitEndpointTests` |
| Both halves of the commit-point rule, stated directly | `CommitTokenTests` |

The race is forced deterministically (an `ICurrentUserAccessor` that reports "no user" for the first two lookups while the winner's row is already committed), not by racing two real requests.

## Notes for the reviewer

- **`ApimKeyService.ProvisionAsync`'s orphan-reuse branch** still regenerates on the request token. Deliberate: nobody holds a key to an orphan, so a cancellation there costs an orphan the next provision adopts by name — the documented, self-healing residue — not a locked-out developer.
- The rate-limit tests read the wall clock. `System.Threading.RateLimiting`'s built-in limiters own their replenishment timer and expose no seam, so a test can only stay inside the window; the durable fix is **#181**'s configuration, and the file says so.
- `UserService.LastLoginGranularity` is `public` so the test asserts against the policy rather than restating `15`.

Closes #163
Closes #158
Closes #168
Closes #154
Closes #136
Closes #167


kolatts:
**APPROVE** — once the PR body is fixed, which is a hard blocker.

Careful, well-argued work. The commit-point sweep is correct at every call site, the rotation restructure fixes exactly the failure #168 described, and the first-login race absorption is tighter than the issue asked for. 560 of the 1040 added lines are tests, and they test the right things. The blocker is metadata: this PR is wearing **#187's description**.

---

**Issues**

- **Major** — PR body and `Closes` links belong to #187. It says `Closes #38 #39 #84 #119 #165`, none of which this branch implements. Merging as-is auto-closes five Functions issues that #187 actually does, and leaves #163/#158/#168/#154/#136/#167 open — the commits only say `Refs`, so nothing here closes them. Swap in the real body before this goes anywhere near the merge button.

- **Minor** `Services/Keys/ApimKeyService.cs:175-182` — the primary regeneration is still on the request token, and its only catch is `ApimSubscriptionNotFoundException`. Disconnect while that ARM call is in flight — the longest operation in the method, so the likeliest moment — and the `OperationCanceledException` walks straight out: no compensation, no `key.rotation-failed`, not even the Error log, while APIM may already have swapped the key. That's the #168 outcome one step earlier, and `IApimKeyService`'s new remark ("the compensation above cannot be skipped by a cancellation") currently overclaims. Nothing to restore at that point, but "we may have reached APIM and the stored key may be stale" is worth an Error log with the same remedy text.

- **Minor** `Extensions/RateLimiterExtensions.cs:~118` — `"anonymous"` is not unreachable. The global auth is an MVC `AuthorizeFilter` (`Program.cs:88`), not endpoint metadata, so `UseRateLimiter` runs before MVC ever evaluates it. Confirmed on the branch: five unauthenticated `POST /keys/me/reveal` return 401, the sixth onward returns **429**. Every anonymous caller on the host shares one 5/min bucket, so a scanner or an expired-token UI turns everyone else's 401 into a 429. `GetNoLimiter` when there's no `oid` costs nothing and makes the comment true.

- **Minor** `Services/Keys/ApimKeyService.cs:198-204` — the secondary regeneration sits between the commit point and the store. If it throws, compensation restores the stale ciphertext and the developer is locked out with a dead primary, even though the new primary was already live and one `ListSecretsAsync` away. Storing the primary first shrinks that failure mode from "locked out" to "secondary left unrotated", which is hygiene rather than an outage.

- **Minor** `Services/CommitToken.cs` — the type lives in the Api, but CONVENTIONS.md now says "never a hand-rolled ternary" for every service touching an external system. #187 puts `QuotaResetService` in `FoundryGate.Core`, which can't reference the Api, so it re-introduces the exact ternary the moment it lands. Worth moving somewhere both reach — Domain has zero deps and `CancellationToken` is BCL. (To be clear, `ApimKeyService` passing `None` inline is fine — an unconditional commit point doesn't need the helper, and the docs say so.)

- **Minor** `docs-site/src/content/docs/architecture/overview.mdx:225-241` — CLAUDE.md docs invariant 2 wants every behavioral change reflected at all levels it touches. `reference/api.md` got a thorough update; the moderate level didn't. That section describes both endpoints you just capped and doesn't mention it, and eleven lines later tells the reader `429 + Retry-After` is the gateway's per-minute cap — so someone who now gets a 429 from the control plane is pointed at the wrong subsystem. Two sentences in the Rotate/reveal paragraph.

- **Nit** `Extensions/RateLimiterExtensions.cs:78-98` — `OnRejected` logs nothing. The body tells the caller "tell an administrator", but nothing tells the administrator: the strongest signal that a drain is in progress is the one event with no `oid` and no user id attached. A `LogWarning` naming the policy and partition key is one line, and it gives #180 something to build on.

- **Nit** `Tests.Predeployment/Api/Endpoints/KeysRateLimitEndpointTests.cs` — the limiter reads wall-clock, so `factory.TimeProvider` can't move it and these tests depend on six round trips landing inside a real 60-second window. Fine today, latent flake later. Worth folding a window/permit seam into #181 so the tests read the configured values instead of mirroring them in two constants.

- **Nit** `Tests.Predeployment/Api/Services/Users/UserServiceTests.cs` — `A_first_login_whose_save_fails_for_any_other_reason_is_not_absorbed` fails the audit service with an `InvalidOperationException`, which never reaches the `catch (DbUpdateException)` at all. The implementation is right — the `AnyAsync` winner check is what makes it right — but the test proves a weaker claim than its name. A too-long string would exercise the actual branch.

---

**Verified, no action needed** — recording these so nobody re-derives them:

- `TierSyncRequested` is a precise commit-point signal: set only after `SyncAsync` returns, and the caller already guards on "has a subscription" and "tier actually changed", so `ApimGatewayTierSync`'s no-subscription early return can't produce a false positive.
- No call site flips the token to `None` before the external call. Every `CommitToken.For(...)` is computed after the call returns, and the transactions still *begin* on the request token.
- `UseRateLimiter()` is correctly after `UseAuthentication()`, so the `oid` partition key is populated for real callers. Partition memory is bounded — the partitioned limiter's heartbeat disposes idle partitions, so it's concurrent oids, not lifetime-distinct ones.
- The race absorption only swallows the real race: `CreateFromFirstLoginAsync` re-queries for the winner before converting to `ConflictException` — provider-agnostic, no error-number sniffing — and the "not repeatable" conflict carries no inner exception, so the filter leaves its 409 alone. Nice.
- `StampLogin` is a genuine one-column update at ≤4/hour/user, and the `AnyAsync` on the race path is an indexed seek on an exceptional branch. No N+1 anywhere in the diff.
- `ApproveAsync` narrowing from unconditional `None` to `CommitToken.For(TierSyncRequested, …)` is a real behavior change from #148, and the right one — when nothing reached the gateway the whole thing should roll back. Worth a line in the body once you write the right one.
- 429 body is `application/json`, which is what the exception handler already emits for a 404. Consistent; not this PR's problem.
- Schema parity, `dotnet format`, build, and 900/900 tests clean on the branch.

---

**Merge order: #184 → #186 → #187.**

#184 and #187 overlap on eight files, including all four services swept here, and #187 deletes the `ResetAsync` block #184 edits. #187 is the one doing `git mv` into a new project, so it goes last — rebasing file moves onto content edits is far easier than reapplying content edits into code that has moved and been rewritten. #186 sits in the middle: its only overlap with #184 is a doc comment on `EntraSyncResult` in a different region of `UserContracts.cs`, plus `CONVENTIONS.md` and `RecordingGatewayTierSync.cs` — all trivial. When #187 rebases it should pick up `CommitToken` rather than re-rolling the ternary, which is the Minor above and a good excuse to widen #182.

Filing #179/#180/#181/#182 for the parts you deliberately left is exactly the right trail.

kolatts:
Thanks — the two rotation catches are both real and both worse than they look on paper. Every item addressed; point by point.

Also merged `origin/main`, so #186 is in (one trivial conflict in `RecordingGatewayTierSync` — your `ThrowFor` and my `AfterSync` are now both there — plus a `LastLoginDate` for the new `WebTestData` fixture, deliberately *not* equal to its `LastSyncedDate` so a page that binds the wrong one is visibly wrong).

---

### Major — PR body belonged to #187. **Fixed.**

You were right and I could not reproduce how it happened: I wrote the correct body when I opened the PR and edited it twice afterwards, and at some point it was replaced wholesale with #187's. Rewritten for this branch, with `Closes #163 #158 #168 #154 #136 #167`, and verified via `gh pr view 184 --json body` that none of #38/#39/#84/#119/#165 survives. Good catch — merging that would have closed five issues nobody had implemented and left six open that this PR fixes.

### Minor — cancellation *during* the primary regeneration. **Fixed**, and the remark was overclaiming.

`ApimKeyService.RotateAsync` now catches `OperationCanceledException` around that call and runs the same compensation path: Error log, `key.rotation-failed` row, remedy text. There is nothing to restore, exactly as you say — so the row is recorded as **`regenerationConfirmed: false`**, and the log line says "*may* now be STALE" rather than "is". That distinction is the whole point for whoever reconciles it: "definitely dead, replace it" and "might be dead, check it" are different jobs, and it would have been dishonest to flatten them into the confirmed message.

`IApimKeyService`'s remark no longer claims cancellation cannot skip compensation; it now names this case as the one the compensation cannot fully cover and says why.

Probe: `Rotate_cancelled_while_ARM_is_still_regenerating_the_primary_leaves_a_trail_rather_than_nothing` — a new `ThrowOnRegeneratePrimaryKey` seam on the fake, asserting the row is untouched, the audit row says unconfirmed, and the Error log landed.

### Minor — anonymous rate-limit bucket. **Fixed.**

You are right about the mechanism and I had the pipeline model wrong: `AuthorizeFilter` is MVC, so `UseRateLimiter` runs long before anything has rejected an anonymous request, and "unreachable" was simply false. `GetNoLimiter` when there is no `oid`, and the comment now explains why that is safe rather than asserting something untrue — the request reaches MVC, gets its `401`, and touches no key material, no database and no gateway, so there is nothing on that path a limiter protects.

Your probe is now a test (`Anonymous_callers_do_not_share_one_bucket_and_keep_getting_their_401`): fifteen unauthenticated reveals, all `401`, and an authenticated caller's own budget untouched afterwards.

### Minor — secondary regeneration between the commit point and the store. **Fixed.**

Reordered: `ListSecrets` → **store the new primary** → regenerate the secondary. A secondary failure no longer throws out of the try at all — it is caught, logged at Error with its own remedy, and recorded on the `key.rotated` row as `keysRegenerated: ["primary"]` plus `secondaryRotationError`, so the rotation still succeeds and the trail is honest about what did and did not happen.

I kept it inside the existing single save rather than committing the primary and then rotating the secondary in a second unit of work. The failure you identified is fully closed either way, and one save keeps the row and the audit row that describes it atomic — a `key.rotated` row written before the secondary attempt could not have told you the outcome, which is the fact worth recording.

Probe: `A_secondary_regeneration_that_fails_costs_the_developer_nothing` — asserts the developer's new primary round-trips from the row, the old secondary is still live at the gateway, and there is no `key.rotation-failed`.

### Minor — `CommitToken` in the Api. **Fixed** — moved to Data, not Domain.

Agreed on the problem: #187 puts `QuotaResetService` in Core, which cannot reference the Api, so it would have re-rolled the exact ternary CONVENTIONS now forbids. I put it in **`FoundryGate.Data/Concurrency/CommitToken.cs`** rather than Domain: the rule is about which token reaches `SaveChangesAsync`, and Data is what every host that saves already references — including Core, which needs Data for the reset. Domain would have worked too; Data puts it next to the thing it is a rule about. CONVENTIONS names the new location, and #182 is widened to say #187 should pick it up rather than re-roll it.

### Minor — `architecture/overview.mdx`. **Fixed.**

Two additions at the moderate register. The Rotate/reveal paragraph now says both routes are capped per signed-in person (5 reveals, 3 rotations a minute), why, and that admin routes are not. And the "Quota exhausted / burst" paragraph no longer implies every `429` is the gateway's: it now says two subsystems can answer `429` and the URL tells them apart — the gateway origin means the per-minute *token* budget, `/api/v1/keys/me/*` means too many key operations, and only the first has anything to do with quota.

### Nit — `OnRejected` logs nothing. **Fixed.**

One line at Information (a rejection is expected traffic shaping, not a fault), carrying the policy name, method, path and partition key. That last field is the one that matters — a run of these for one `oid` is the signal, and it is what #180 will build on.

### Nit — wall-clock in the rate-limit tests. **Documented, not fixed.**

There is no seam: `System.Threading.RateLimiting`'s built-in limiters own their replenishment timer and neither `FixedWindowRateLimiterOptions` nor the middleware takes a `TimeProvider`, so a test can only stay inside the window rather than move it. The class remarks now say that, and point at **#181** — once the window and permits are configuration, these tests can shorten the window instead of racing it, and read the configured values instead of mirroring them in two constants. Added that to #181's body.

### Nit — the "not absorbed" test proved a weaker claim. **Fixed.**

Correct, and it was the one test in the set whose name outran what it did. It now creates a unique index on `Email` in its own database, seeds a user holding that address, and has the first-login caller present the same one — a genuine `DbUpdateException` with nothing to do with the oid, so the `AnyAsync` winner check finds nobody and it surfaces. Renamed to `A_first_login_whose_insert_fails_for_a_reason_other_than_the_oid_race_is_not_absorbed`. The audit-failure case is kept as its own test for what it does prove: the pipeline is one transaction, so a later failure still leaves no row.

---

### On your verified list

Two worth acknowledging rather than re-deriving:

- **`ApproveAsync` narrowing from unconditional `None`** — agreed it is a real behaviour change from #148 and the right one. It is now called out in the body: when resolution never reached the gateway there is nothing to protect, so an abandoned approval rolls back with its transaction instead of committing a decision nobody is waiting for.
- **Merge order #184 → #186 → #187** — #186 is merged into this branch already. Happy to go first, and #182 now tells #187 to adopt `CommitToken` from Data on the way through.

Gates on the merged head: build clean with warnings-as-errors, `dotnet format --verify-no-changes` clean, docs build passing, **1033/1033** predeployment tests green.

Issue #185, opened 2026-09-02
Prune SystemConfigurationKeys.Retired once every fork has re-seeded


kolatts:
> Parent: #224 — Engineering debt

Bookkeeping left by #164/#123.

Retiring a `SystemConfiguration` key is a two-part move: drop it from
`SystemConfigurationKeys.All` and `SystemConfiguration.GetSeedData()` so it stops being seeded, and
name it in `SystemConfigurationKeys.Retired` so `ReferenceDataSeeder`'s `deleteFilter` still covers
it and the row is deleted from databases that already have it. Without the second half the row would
be stranded in every deployed fork forever, because the filter only ever deletes keys FoundryGate
itself names.

`Retired` therefore has to outlive the removal — but only until every fork has run one
`db seed-reference`. After that the entries are pure history and the list grows monotonically with
every future retirement.

## Work

Decide the retention rule and apply it: either drop `ApimGatewayUrl` / `ApimProductId` /
`EntraTenantId` from `Retired` after a stated release boundary (with the "upgrade through vX first"
note in the docs that implies), or keep the list forever as documentation and say so in its XML doc,
in which case this issue closes as "won't do" with that decision recorded.

Nothing is broken either way; a stale entry costs one `Contains` per seeded row.

Issue #183, opened 2026-09-02
Graph: verify $select/$top on /transitiveMembers without ConsistencyLevel: eventual


kolatts:
> Parent: #220 — Live validation of the dev environment

Discovered while implementing #121.

`GraphEntraDirectoryClient.ListGroupMemberIdsAsync` calls
`GET /groups/{id}/transitiveMembers?$select=id&$top=999` with **no** `ConsistencyLevel: eventual` and
no `$count`. That call is now on two paths, not one: group sync (#41) and — as of #121 — expanding
every group-principal app-role assignment during `POST /users/sync`.

The v1.0 reference for [List group transitive members](https://learn.microsoft.com/en-us/graph/api/group-list-transitivemembers?view=graph-rest-1.0)
says, under Optional query parameters:

> The use of query parameters with this API, except for `$expand`, is supported only with advanced
> query parameters.

Read strictly, that makes `$select` and `$top` advanced-query parameters, which require
`ConsistencyLevel: eventual` (+ `$count=true`) and read from an eventually-consistent index. The
plain `/members` collection carries no such sentence, which is why the existing code was written
this way and deliberately avoided the eventual index — a reconciliation sync that can deactivate
people should not read a stale view.

**Unknown**: whether real Graph actually rejects (or silently ignores) `$select`/`$top` on
`transitiveMembers` without the header. Every test covering it is stubbed HTTP, so the suite proves
the client sends what we intend, not that Graph accepts it.

## Work

1. Against a real tenant, call `GET /groups/{id}/transitiveMembers?$select=id&$top=999` with an
   application token holding only `GroupMember.ReadBasic.All` and record the response — 200 with the
   ids, 200 ignoring `$select`, or `400 Request_UnsupportedQuery`.
2. If it is rejected or `$select` is ignored, decide between dropping `$select`/`$top` for the
   transitive call (paying full directoryObject payloads at Graph's default page size of 100) and
   sending `ConsistencyLevel: eventual` + `$count=true` (accepting the eventual index, and revisiting
   what that means for departure detection in `EntraUserSyncService`).
3. Whatever the answer, pin it in `GraphEntraDirectoryClient`'s remarks — the current comment states
   the avoidance as settled, and it is not.

Related: #120's live-validation checklist has the same question for plain `/members`; this is the
transitive half, which #121 made load-bearing for user sync too.

PR #186, opened 2026-09-02
feat(api): expand Entra group assignments, isolate group sync failures, retire dead config keys

### LINKED ISSUE #121: Entra user sync: expand group-based app role assignments (principalType=Group) to their user members

### LINKED ISSUE #123: SystemConfiguration EntraTenantId is unused by identity-based Graph sync — remove or repurpose

### LINKED ISSUE #145: Retire RequestDtoBindingController actions as production controllers bind the request records

### LINKED ISSUE #149: Entra group sync: hoist the user map out of the per-group loop and decide SyncAllAsync's failure semantics

### LINKED ISSUE #164: Retire the two dead SystemConfiguration keys (ApimProductId, EntraTenantId)

### LINKED ISSUE #189: Retire the dead ApimGatewayUrl SystemConfiguration key
Split out of #164 so the removal is covered by an issue of its own, per the "everything is a GitHub issue" rule — #164 only ever named `ApimProductId` and `EntraTenantId` ("Two of the eight seeded keys...").

The finding is the same and surfaced in #156's review: `SystemConfigurationKeys.ApimGatewayUrl` is seeded, listed by `GET /config`, has a per-key validation rule (`NormalizeHttpsUrl`), and is **read by nothing**. The gateway origin `GET /users/me` hands developers as `cliConfig.gatewayBaseUrl` comes from `GatewayOptions.ApimGatewayUrl`, which infra sets on the container from the APIM module's own output (`Gateway__ApimGatewayUrl`) — so the address can never drift from the gateway that was actually deployed. Its own XML doc already said "Legacy... Nothing reads this row".

Leaving it seeded is worse than removing it: an operator editing it gets a `200` and a stored value that changes nothing about where their developers' CLIs point.

## Work

Same treatment as #164's two keys, and in the same pass: drop it from `SystemConfigurationKeys.All`, `SystemConfiguration.GetSeedData()` and `SystemConfigValidator.Normalize` (taking `NormalizeHttpsUrl` with it, its only caller), name it in `SystemConfigurationKeys.Retired` so the reference-data seeder's delete filter still covers it and existing rows are removed on the next `db seed-reference`, and update `reference/configuration.mdx` and `reference/api.md`.

kolatts:
Five issues from the hardening wave: Entra group expansion, group sync-all semantics, dead configuration keys, and the last test-only controller.

## `POST /users/sync` expands group-based app-role assignments

`GraphEntraDirectoryClient.ListAssignedUsersAsync` now flattens every `principalType = Group` assignment to its **transitive** user members (`GET /groups/{id}/transitiveMembers?$select=id&$top=999`, next-link paged) and merges those ids into the same set as the direct assignees before hydration. Assigning `SG_AI_Developers` to the FoundryGate enterprise application — the common enterprise pattern — is a fully supported configuration now: adds, updates and departure detection all work for its members, de-duplicated against anyone assigned both ways.

`EntraAssignedUsers.SkippedGroupAssignments` changes meaning rather than disappearing: it carries only the groups whose expansion **failed** (Graph refused, or the group is gone). That is exactly the case where the run has a partial view of the population, so `EntraUserSyncService` keeps suspending departure detection on a non-empty list and the protection survives. On a healthy tenant `skippedGroupAssignmentCount` is `0`.

**A group is "unreadable" for transport faults too** (review follow-up). The expansion loop caught only `ODataError`, which models Graph's structured error body; once Kiota's retry handler gives up on DNS, a reset connection or a timeout the caller sees `HttpRequestException` or a `TaskCanceledException` instead, and those propagated — turning "Graph was briefly unreachable while reading one group" into a 500 for the whole run. All three now degrade that one group. The caller's own cancellation is never swallowed.

**The Graph role set is unchanged.** `GroupMember.ReadBasic.All` is the least-privileged application permission the [Graph v1.0 reference](https://learn.microsoft.com/en-us/graph/api/group-list-transitivemembers?view=graph-rest-1.0) lists for transitive members, and only `id` is selected, so `GroupMember.Read.All` stays unnecessary. #120 has a comment with the revised live-validation checklist.

## `POST /groups/sync-entra` reads the user table once and survives one group failing

The whole `Users` table was read once per linked group; it is now read once for the run and shared across the loop (nothing in this path inserts a user, so one snapshot cannot go stale). A failing group no longer ends the run: it is left exactly as it was, its staged-but-unsaved changes are detached so they cannot ride along on the next group's `SaveChangesAsync`, and the loop continues. `GroupSyncResult` gains `bool Succeeded = true` / `string? Error = null` — defaulted, so every existing construction and the Web API client compile unchanged.

`FeatureNotConfiguredException` and cancellation stay whole-run failures on purpose: "Entra is off on this host" is not a property of any one group, and a cancelled request has no caller left to read a summary.

**Not every failure is the same failure** (review follow-up). The commit is its own `CommitAsync` step now, because `commitToken` already knows when `tierSync.SyncAsync` has accepted a tier move. Below that commit point a failure is CONVENTIONS.md's "the external system accepted the change and the write did not land" case: it gets the Error log with the group's full identity that `FoundryDeploymentService.AuditAfterCommitAsync` established, then **one retry on `CancellationToken.None` with the pending rows still tracked** — those rows are the only record of what the gateway did, so discarding before retrying would guarantee the divergence. Only if the retry fails too is the group marked `GroupSyncErrorType.PostCommit` (with a second Error line summarizing the run), and `DiscardPendingChanges()` runs only after the retry has had its chance. Out of single-group `SyncAsync` it still throws.

`GroupSyncResult` therefore also gains `GroupSyncErrorType ErrorType = None`. `"GraphRead"` means nothing was applied anywhere and re-running is sufficient; `"PostCommit"` means the gateway changed and the database did not. A UI must render those differently — "try again" is right advice for one and actively misleading for the other.

## Three dead `SystemConfiguration` keys are gone

`ApimGatewayUrl` (#189, split out of #164 after review — #164 only ever named the other two), `ApimProductId` and `EntraTenantId` were seeded, listed, and read by nothing. #161 made editing them a `409 read-only`; this removes the rows and the read-only branch with them (a retired key is now simply unknown, which `PUT /config/{key}` already answers with `404`).

The delete needed care. `ReferenceDataSeeder`'s `deleteFilter` deliberately restricts orphan deletion to keys FoundryGate names, so dropping a key from `All` alone would have **stranded** its row in every deployed database rather than cleaning it up. Hence `SystemConfigurationKeys.Retired`: named there, never seeded, still inside the delete filter — so the next `db seed-reference` deletes them (the CLI already prints the count) while a fork operator's own rows stay untouched. A convention test pins that the two lists never overlap.

## The test-only binding controller is retired

Every one of the seven request records `RequestDtoBindingController` stood in for now has a production endpoint, so the controller and its `AddApplicationPart` registration are deleted and `RequestDtoBindingTests` posts each invalid body to the real route instead. Better guard: the binder under test is the production one, and a route that stops binding its record fails here rather than passing against an echo. `DomainArchitectureTests`'s reflection guard is untouched.

## Verification

- `FoundryGate.Tests.Predeployment` — **887 passed, 0 failed** (881 on main)
- `npm run build` in `docs-site/` — clean

## Docs

`reference/api.md` (departure detection re-enabled; sync-all failure semantics; the config 409 story removed), `reference/configuration.mdx` (a "Retired keys" table with what to read instead, plus a caution that an edited value is lost on the next seed), `reference/infrastructure.md` (why `GroupMember.ReadBasic.All` now covers two call sites), `architecture/overview.mdx` and `getting-started/fork-guide.mdx` (group assignment is supported; a non-zero skipped count now means a permission problem), and CONVENTIONS.md's §API service note.

## Decisions worth a reviewer's eye

1. **Failed expansion keeps the suspension rather than failing the run.** A group the API cannot read is a permission problem the operator fixes, not a reason to refuse the adds and updates that are correct — but it is also not a reason to deactivate everyone that group covers. Reusing `SkippedGroupAssignments` for it keeps the existing protection and the existing audit field.
2. **`GroupSyncResult`'s new members are defaulted.** Keeps it a positional response record per CONVENTIONS.md while leaving `new GroupSyncResult(id, a, r, s)` and the Blazor client compiling untouched.
3. **`DiscardPendingChanges` detaches Added/Modified/Deleted entries after a failed group.** Necessary because the groups share one request `AppDbContext` and `IQuotaResolutionService` does not save — without it, a group that threw after staging memberships would have them committed by whichever group saved next. A test raises the failure at exactly that point. After review it runs *after* the post-commit retry, never before it.

3b. **The post-commit retry is one attempt, not a loop.** A second failure of the same write is far more likely to be a durable problem than a blip, and the group's summary plus two Error log lines put it in front of a person; re-running the endpoint is the documented recovery and is idempotent. #163 tracks centralizing this whole pattern.
4. **Retiring a key needs `Retired`, not just removal from `All`.** Removal alone strands the row forever. #185 tracks when those three entries may be pruned.
5. **`FeatureNotConfiguredException` is not isolated per group** — every group would carry the same 503 message, and the host is what needs fixing.

## Deferred

- #183 — verify `$select`/`$top` on `/transitiveMembers` without `ConsistencyLevel: eventual`; the Graph reference implies they are advanced-query parameters, and every test of that call is stubbed HTTP. Load-bearing for both #41 and #121 in production.
- #185 — decide when `SystemConfigurationKeys.Retired` entries may be pruned.
- #52 has a comment: the admin group pages' Entra sync dialog must render `succeeded: false` rows distinctly now, not just sum the counts.
- #122 (`accountEnabled = false` as departure) is untouched — an owner decision.

Closes #121
Closes #149
Closes #164
Closes #123
Closes #145
Closes #189


kolatts:
**REQUEST CHANGES**

Good PR — the group expansion is the right shape, the delete-filter retirement is careful, the DTO-binding cleanup is a strict upgrade, and the test suite actually proves the hard cases (paging, de-dup, unreadable groups, cross-group isolation). One real gap in the failure-isolation work: it treats a Graph-read failure and a "gateway already moved, DB write then failed" failure identically, which is exactly the distinction CONVENTIONS.md's commit-point rule exists to preserve.

---

**Issues**

- **Major** `src/FoundryGate.Api/Services/Groups/EntraGroupSyncService.cs:72-94` — `SyncAllAsync`'s per-group `catch` treats every exception the same way (Warning log, `Succeeded: false`, `DiscardPendingChanges()`), but it has the information to know better: `commitToken` at line ~144 already flips to `CancellationToken.None` exactly when `tierSync.SyncAsync` has accepted a move. If `SaveChangesAsync`/`audit.LogAsync` then throws — a concurrency conflict, a dropped SQL connection — this is the "Azure accepted the change but the write failed, reconcile manually" case `FoundryDeploymentService.AuditAfterCommitAsync` logs at Error and rethrows for. Here it's folded into the same per-item Warning as an ordinary pre-mutation Graph fault, and `DiscardPendingChanges()` throws away the pending `QuotaAllocation` row that recorded what should have happened. Worse, before this PR that failure was at least a loud 500; now it's a quiet `Error` string in one row of a 200 array. #163 already tracks centralizing this pattern — but this PR is the one making the failure mode swallowable, so it's worth at minimum a distinct Error-level log at that specific call site rather than lumping it in.

- **Minor** `src/FoundryGate.Api/Services/Entra/GraphEntraDirectoryClient.cs:165` — the group-expansion loop only catches `ODataError`, which models Graph's structured API error body. A transport-level failure after the SDK's retry handler gives up (DNS, connection reset, timeout) surfaces as `HttpRequestException`/`TaskCanceledException`, not `ODataError`, and would propagate uncaught through `ListAssignedUsersAsync` — turning "Graph is briefly unreachable" into a 500 for the whole `POST /users/sync` run instead of the per-group suspension this feature exists to provide.

- **Nit** `src/FoundryGate.Domain/Constants/SystemConfigurationKeys.cs` / PR description — `ApimGatewayUrl` gets retired alongside `ApimProductId`/`EntraTenantId`, but #164 only ever named the latter two ("Two of the eight seeded keys..."). The dead-config finding for `ApimGatewayUrl` was real (surfaced in #156's review) and the removal is correct, it just isn't `Closes`'d by anything — worth a line in the PR body or its own issue number so "everything is a GitHub issue" holds.

---

Everything else checks out: `SkippedGroupAssignments` correctly narrows to "expansion failed" and the departure-detection suspension still covers every partial-view case I could construct; the transitive-member paging and de-dup are both tested against a real next-link and a real duplicate; #183's `$select`/`$top`-without-`ConsistencyLevel` question is honestly left open rather than quietly assumed away; the retired-key delete filter is proven to strand nothing and delete nothing it shouldn't (`SeedAsync_leaves_a_row_a_fork_operator_added_alone`); `[DoNotUpdate]` is untouched; `GroupSyncResult`'s new members are correctly defaulted so the Blazor client and every existing constructor call still compile; and `RequestDtoBindingTests` now hits real routes with real 400s, including the missing-id-but-still-400-before-404 case for the two group/quota mutation endpoints. Build clean, 880/880 predeployment green, `dotnet format` clean.

Expect a trivial merge conflict with #187 on `EntraGroupSyncService.cs`'s `using FoundryGate.Api.Services.Quota;` → `FoundryGate.Core.Quota` line (that PR moves the whole Quota folder) — textual only, not semantic. No real overlap with #184 (touches `Lifecycle`/`Keys`/`Program.cs`, and its one shared file — `UserContracts.cs` — edits a different record than this PR does).

kolatts:
All three addressed. Point by point:

---

**Major — `SyncAllAsync` treats a post-commit-point failure like a Graph read** (`d00eb22`)

You're right, and the framing "this PR is the one making the failure mode swallowable" is the part that decided it: before #149 that failure was a 500, so folding it into the per-item Warning was a regression I introduced, not a pre-existing gap I merely failed to fix.

The commit is its own step now (`CommitAsync`), split on the `gatewayMoved` flag that `commitToken` was already derived from:

- **Above the commit point** — nothing changed. The exception propagates and `SyncAllAsync` handles it as the ordinary per-group fault it is.
- **Below it** — Error log with the group's full identity and the reresolved user ids, using your precedent's wording ("the gateway accepted the tier move ... reconcile manually"), then **one retry of the save on `CancellationToken.None` with the pending rows still tracked**. Your point about `DiscardPendingChanges()` throwing away the `QuotaAllocation` row is exactly why the retry has to come first: those rows are the only record of what the gateway was told, so discarding before retrying would guarantee the divergence rather than close it. The discard now runs only after the retry has had its chance.
- **If the retry fails too** — `GroupSyncPostCommitException`, a second Error log, and the group is marked `GroupSyncErrorType.PostCommit`. Out of single-group `SyncAsync` it propagates and the request fails loudly, matching `AuditAfterCommitAsync`. A run with any such group also gets an Error-level summary line, because "3 groups reconciled" at Information with the divergence buried mid-array is how this gets missed.

`GroupSyncResult` gains `GroupSyncErrorType ErrorType = None` alongside `Error` (defaulted, so the Blazor client and every constructor call still compile). The category is the load-bearing part: `"GraphRead"` means nothing was applied anywhere and re-running is sufficient; `"PostCommit"` means the gateway changed and the database did not. "Try again" is right advice for the first and actively misleading for the second, so a UI must not have to parse a message to tell them apart. Documented as a table in `reference/api.md`, and the note on #52 for the sync dialog is updated.

Four tests, needing one new test-support hook (`RecordingCommandInterceptor.FailWhen`, to break a save at a chosen statement) plus the existing `CapturingLoggerProvider` wired into the suite:

- audit write fails after a tier move → Error logged, **retry succeeds**, membership and allocation land, result is a plain success (this is the one that pins "don't discard before retrying")
- membership insert fails on every attempt → `PostCommit` marker, both Error lines, and `_tierSync.Calls` proves the gateway really was told
- a Graph fault → `GraphRead`, no tier call, and explicitly *no* "reconcile manually" line
- single-group `SyncAsync` under the same failure still throws

Kept as one retry rather than a loop: a second failure of the same write is far more likely durable than transient, and re-running the endpoint is idempotent and documented. #163 remains the right home for centralizing the pattern.

---

**Minor — the expansion loop only catches `ODataError`** (`a5edf2d`)

Fixed, and thank you — this was a real hole rather than a stylistic one. `IsRecoverableGroupReadFault` now covers `ODataError`, `HttpRequestException` and `TaskCanceledException`, with the caller's own cancellation explicitly excluded: if `cancellationToken.IsCancellationRequested`, everything propagates, because a client that walked away is not a directory fault.

Tests: a theory over both transport shapes (`HttpRequestException`, and a `TaskCanceledException` wrapping a `TimeoutException` the way an `HttpClient` timeout arrives) proving the unreachable group is reported as unexpanded while a readable one still contributes its members — plus a companion that cancels the token mid-call and asserts the `TaskCanceledException` still escapes. `StubHttpMessageHandler.OnTransportFailure` is new; `OnJson` cannot express "no response at all".

`reference/api.md` now lists "Graph is briefly unreachable and the SDK's retries are exhausted" alongside the permission and deleted-group causes of a non-zero `skippedGroupAssignmentCount`.

---

**Nit — `ApimGatewayUrl` isn't `Closes`'d by anything**

Filed as #189 with the finding written out (seeded, listed, validated by `NormalizeHttpsUrl`, read by nothing; the origin on `/me` comes from `Gateway:ApimGatewayUrl`, which infra sets from the APIM module's output). `Closes #189` added to the PR body, and the body no longer implies #164 covered all three.

---

**Gates after the fixes:** build 0 warnings / 0 errors, **887/887** predeployment green, `dotnet format --verify-no-changes` clean, `docs-site` build clean. `git merge origin/main` is already up to date — holding for #187 to land before the `FoundryGate.Core.Quota` rebase, and not merging.

Issue #178, opened 2026-09-02
Live-validate the reconciliation KQL and both Function timers against a deployed gateway


kolatts:
> Parent: #220 — Live validation of the dev environment

Follow-up to #39/#84/#38, filed by the Functions implementation wave and revised after the #187 review. Everything in that PR is unit-tested against fakes — Azure Monitor Logs has no emulator, and a test that needs a live workspace is a test nobody runs. These are the parts only a deployed environment can answer. Companion to #105 (control-plane deploy) and #88 (Claude end-to-end).

## 1. The KQL returns what the job expects

Open the dev workspace (`log-foundrygate-dev`) > Logs, set the time range to the current month, and run the checked-in query verbatim — `src/FoundryGate.Functions/Kql/UsageBySubscription.kql`:

```kusto
ApiManagementGatewayLlmLog
| where isnotempty(CorrelationId)
| summarize
    PromptTokens = max(PromptTokens),
    CompletionTokens = max(CompletionTokens),
    TotalTokens = max(TotalTokens)
  by CorrelationId
| join kind=inner (
    ApiManagementGatewayLogs
    | where isnotempty(ApimSubscriptionId)
    | distinct CorrelationId, ApimSubscriptionId
  ) on CorrelationId
| summarize
    PromptTokens = sum(PromptTokens),
    CompletionTokens = sum(CompletionTokens),
    TotalTokens = sum(TotalTokens),
    RequestCount = count()
  by ApimSubscriptionId
```

- [ ] It parses (column names on both tables are as documented — `PromptTokens`, `CompletionTokens`, `TotalTokens`, `CorrelationId`, `ApimSubscriptionId`)
- [ ] The join actually matches: `RequestCount` is non-zero after driving traffic through the gateway
- [ ] `ApimSubscriptionId` renders as the subscription **name** (`foundrygate-{UserId}`), which is what `ApimSubscriptionNames.TryGetUserId` parses — if it renders as an ARM id or a display name instead, `UsageSyncJob` will count every row as an unknown subscription and the mapping needs changing

### 1a. The de-duplication assumption (the one that decides whether these numbers are billable)

The query collapses the LLM log to one row per `CorrelationId` with `max()` before summing, because the docs describe the token columns per *request* and emit several entries per exchange. **Verify that directly** — this is the assumption the whole rollup rests on, and getting it wrong multiplies or under-counts every developer's usage:

```kusto
// How many entries per request, and do their token counts agree?
[...]

Issue #177, opened 2026-09-02
Per-developer cost attribution: rate card in SystemConfiguration + cost on the dashboard


kolatts:
Split out of #84 (item 3) while implementing the usage-sync Function. Items 1, 2 and 4 of #84 landed in the Functions wave (`Kql/UsageBySubscription.kql`, `UsageSyncJob`, the streaming-undercount caveat in `architecture/overview`); cost attribution did not, because it is a separate surface (config keys + a dashboard field), not part of the reconciliation job.

## Why it cannot come from Azure Cost Management

Claude bills as a single aggregate Marketplace CCU meter — Cost Management cannot break Claude spend down per deployment, per subscription or per user (research.md §8, plan 24 §Metrics). OpenAI models additionally support the `deployment` billing tag, which is useful as a cross-check but does not attribute per developer either.

So per-developer cost has to be computed: `tokens × rate card`.

## Ask

- **Rate card storage.** New `SystemConfiguration` key(s) holding per-model input/output rates (or a small JSON blob under one key — decide, and mind `ValidationConstants.ConfigValueMaxLength`). It must go through the #161 config-validation path so a malformed rate card is a 400, not a wrong invoice.
- **Per-model token split.** `QuotaAllocation` records only a total. Cost needs prompt vs completion tokens, and per model — `Kql/UsageBySubscription.kql` already selects `PromptTokens`/`CompletionTokens` and the LLM log carries `ModelName`, so the query is a `by ApimSubscriptionId, ModelName` away. Where the split is stored (a new table? extra columns?) is the real design question here.
- **Dashboard.** Estimated cost per developer for the current period, alongside the token figures the dashboard already shows (#166 landed the dashboard service).
- **Honesty about precision.** Whatever is shown must be labelled an estimate: the token totals are a floor (interrupted streams undercount — #84), and cache-read/creation token weighting at the gateway is still unverified (#88).

## Not in scope

Changing enforcement. Cost is a reporting figure; the gateway enforces tokens, not dollars.

kolatts:
Claiming this for the final sweep C branch (claude/final-sweep-c).

PR #187, opened 2026-09-02
feat(functions): monthly quota reset and usage reconciliation, on a shared FoundryGate.Core

### LINKED ISSUE #38: Implement monthly quota reset as an Azure Function Timer Trigger

### LINKED ISSUE #39: Implement token usage sync as an Azure Function Timer Trigger
Part of #10.

`POST /internal/sync-usage` is authenticated by a shared header key (environment variable, not Entra) so a Container App cron job can call it. Query Azure Monitor Logs (Log Analytics workspace) using the `Azure.Monitor.Query` SDK for the APIM `llm-emit-token-metric` events in the current period, grouped by APIM subscription ID. Map subscription IDs to `QuotaAllocation` rows and update `TokensUsed`. This endpoint is idempotent — subsequent calls in the same period overwrite, not accumulate.

### LINKED ISSUE #84: Emit per-developer token metrics and reconcile usage from ApiManagementGatewayLlmLog

### LINKED ISSUE #119: Monthly reset Function cannot reference Api's IQuotaResolutionService — decide HTTP call vs moving resolution to Data

### LINKED ISSUE #165: SystemConfiguration.ResetDayOfMonth is stored but nothing reads it

kolatts:
The Functions host stops being a scaffold. Both scheduled jobs land, and the structural
prerequisite they needed — a home for services more than one host uses — lands with them.

Closes #38
Closes #39
Closes #84
Closes #119
Closes #165

---

### `src/FoundryGate.Core` — services more than one host needs (#119)

A class library referencing Data + Domain with **no ASP.NET Core dependency**, which is
precisely what makes it usable from an isolated worker. Moved in (`git mv`, history intact):
`GatewayOptions`/`GatewayTier` → `FoundryGate.Core.Configuration`;
`IQuotaResolutionService`/`QuotaResolutionService`, `QuotaResolution`/`QuotaPreview`,
`GatewayTierMapper`/`GatewayTierAssignment`, `IGatewayTierSync`/`NullGatewayTierSync` →
`FoundryGate.Core.Quota`, plus a new `AddQuotaCore()`.

New in Core: **`IQuotaResetService`**, which owns the reset both hosts run. The Api's
`QuotaAllocationService.ResetAsync` now resolves the acting admin and delegates; the Function
passes `QuotaResetTrigger.Scheduled()`. One implementation, two audit actions —
`quota.reset` naming the admin, `quota.monthly-reset` with no actor.

`ApimGatewayTierSync` stays in the Api (it composes the APIM key service). The Functions host
registers `NullGatewayTierSync`, which is safe because a reset re-resolves inputs nobody
changed and so cannot produce a tier change — asserted, not assumed
(`QuotaResetServiceTests.A_scheduled_reset_never_moves_a_subscription_between_tier_products`).

CONVENTIONS §Solution structure amended per the orchestrator's ruling; D-014 in
`fable-refactor-log.md` records why, with the options that lost.

### `MonthlyQuotaResetFunction` (#38) + `ResetDayOfMonth` (#165)

`TimerTrigger("0 1 0 * * *")` — **daily** at 00:01 UTC, not monthly. `MonthlyResetJob` reads
`SystemConfiguration[ResetDayOfMonth]` and stops on any other day, which is how #165 gets
honoured at all: a cron expression is fixed at deployment, so deriving it from the key would
mean a redeploy to change a setting — the same dead key with extra steps. Idempotency is what
makes the daily tick safe.

Then a blob lease on the Functions storage account (`IResetLock`/`BlobResetLock`,
identity-based; `NullResetLock` where no storage resolves), then Core's reset: upsert every
active user's allocation, **preserve `TokensUsed`**, clear `IsHardStopped`, stamp `ResetDate`,
one audit row, single save. **No APIM call** — the gateway's `llm-token-limit` monthly window
is a UTC calendar month that resets itself.

### `UsageSyncFunction` (#39) + the reconciliation feed (#84)

`TimerTrigger("0 */15 * * * *")`. `IUsageQueryClient` runs the checked-in
`src/FoundryGate.Functions/Kql/UsageBySubscription.kql` — `ApiManagementGatewayLlmLog` joined
to `ApiManagementGatewayLogs` on `CorrelationId`, summarised per `ApimSubscriptionId` —
through `LogsQueryClient` against the workspace GUID in `Gateway:LogAnalyticsWorkspaceId`.

`UsageSyncJob` maps subscriptions back to users via `ApimSubscriptionNames.TryGetUserId`,
**overwrites** `TokensUsed` with period totals, counts unknown subscriptions, and records
over-budget usage as **drift** (a Warning with the delta plus a count in the audit row). It
never sets `IsHardStopped`: enforcement is the gateway's 403, and that flag means offboarding.

### Functions host

`Program.cs`: `FunctionsApplication.CreateBuilder`, environment parsed to the lowercase
`local`/`qa`/`prod` enum, `AppTokenCredential`, `@KeyVault()` resolution,
`Configuration/AppSettings.cs` + `ValidateRecursively()`, `AddFoundryGateData`,
`AddFoundryGateFunctionsServices` (→ `AddQuotaCore()`), and worker OpenTelemetry → Azure
Monitor behind `OpenTelemetry:Enabled`.

---

## Design decisions worth a reviewer's attention

**1. `ValidateRecursively()` stops at assembly boundaries — this nearly shipped as a silent
regression.** `Imagile.Framework.Configuration` recurses only into property types declared in
the *root object's own assembly*. Moving `GatewayOptions` to Core therefore removed it from
startup validation entirely: a fork could have started with no `Gateway:Tiers` at all and only
found out at the first quota resolution. Caught by nine existing `AppSettingsValidationTests`
going red. Fix: each host's `AppSettings` is an `IValidatableObject` yielding
`CoreOptionsValidation.ValidateGateway(Gateway, nameof(Gateway))`, reproducing the same errors
and the same `Gateway.Member` message paths. Written into CONVENTIONS so the next Core-owned
options class does not rediscover it.

**2. Daily timer + a gate, rather than a monthly cron** — see #165 above and D-015.

**3. An idle reconciliation pass writes no audit row (D-016).** The brief said one
`usage.synced` row per run; at 96 ticks a day that is ~35k rows a year of "nothing happened"
in the table the admin audit viewer pages through, burying every real admin action. So the row
is written when the pass saw traffic or changed an allocation, and skipped (Debug log only)
when it did neither. The run itself is visible in Application Insights either way. Happy to
flip this if you'd rather have the heartbeat.

**4. The scheduled job audits as `quota.monthly-reset`, not `quota.reset`.** The brief said
`quota.reset`; `AuditActions.QuotaMonthlyReset` already existed for exactly this job (spec §6
fixes the string), and keeping them distinct is what lets an admin tell "the timer did this"
from "someone pressed the button" in the audit trail.

**5. The reset lock needed no new infra setting.** `Storage:AccountName` is optional and
normally empty: the host already gets `AzureWebJobsStorage__accountName` from
`function-app.bicep`, and the Functions identity already holds Storage Blob Data Owner. A cloud
*connection string* is refused at startup — the account has `allowSharedKeyAccess: false`, so
one could not work anyway.

**6. `UsageSyncJob` never mints allocation rows.** A developer who spends tokens before their
first `/me` of the month has no row to write to; the pass logs it and moves on, and the next
pass fills in the total, because the query is cumulative. Counted separately in the audit row
(`missingAllocations`) rather than folded into `unknownSubscriptions`, which would have been
misleading.

**7. Both hosts ship the same tier table.** `FoundryGate.Functions/appsettings.json` carries
`Gateway:Tiers` like the Api's; a predeployment test asserts the two files are identical, so
they cannot drift into resolving the same developer to different budgets.

---

## Verification

- `dotnet build FoundryGate.sln -c Release` — **0 warnings, 0 errors** (the `WorkerExtensions`
  inner build is fine at this path).
- `dotnet test src/FoundryGate.Tests.Predeployment -c Release` — **920 passed, 0 failed**
  (881 on main after #166; 39 new).
- `npm run build` in `docs-site/` — 14 pages, clean.
- `dotnet publish src/FoundryGate.Functions` — matches `_deploy-functions.yml`'s assertions
  (`host.json`, `.azurefunctions/`), and `functions.metadata` shows both triggers with the
  right schedules and `runOnStartup: false`.

New tests: `Core/Quota/QuotaResetServiceTests` (system vs admin actor, `TokensUsed` preserved,
`IsHardStopped` cleared, idempotency, inactive users skipped, no tier sync),
`Functions/Services/Quota/MonthlyResetJobTests` (day gate on/off, a changed `ResetDayOfMonth`,
unusable and missing values, lock honoured and released, lock refused writes nothing),
`Functions/Services/Usage/UsageSyncJobTests` (overwrite, convergence on re-run, drift without
hard-stop, unlimited cannot drift, unknown subscriptions, missing allocation rows, audit shape,
silent no-op, only the current period), `UsageQueriesTests` (the KQL loads from the embedded
file and names both tables), `Functions/Configuration/FunctionsAppSettingsTests` (fail-fast,
ARM-id-vs-GUID, storage connection-string rule, tier parity with the Api).

**No live Azure was touched.** Log Analytics is behind `IUsageQueryClient` with a fake and the
lease behind `IResetLock` with a fake. The KQL is checked in so the live pass is a copy-paste.

---

## What I bound from #108

`GatewayOptions` gained `LogAnalyticsWorkspaceId` and `LogAnalyticsWorkspaceResourceId` —
infra has been setting both since #43/#44 and nothing was reading them. The GUID one is
validated as a GUID at startup: handing the query API an ARM resource id is the single most
likely way to wire this wrong, and it would otherwise fail fifteen minutes later inside a timer
rather than at boot. No new environment variable was needed anywhere.

## Deferred, with issues

- **#177** — per-developer cost attribution (#84 item 3): rate card in `SystemConfiguration`,
  the per-model token split the KQL would need, and the dashboard figure. Out of scope here
  because it is a config + dashboard surface, not part of the reconciliation job.
- **#178** — live validation against a deployed gateway: the KQL run by hand (query included in
  the issue), both timers firing, the blob lease against the real storage account, and the
  `ApimSubscriptionId` rendering that the user mapping depends on.
- **#163** — commented, not closed: `QuotaAllocationService.ResetAsync`'s commit point is now
  handled inside `QuotaResetService` (one place, derived from `TierSyncRequested`), and the Api
  resolves the actor before the handoff. `ApimKeyService` and the `GetMyAllocationAsync` re-check
  remain.
- **#151** — commented with the pattern the scheduled Entra sync should follow (thin function →
  testable job, system-actor trigger, `IResetLock` reuse, the #165 cadence shape).

## Docs

`architecture/overview` (both pipelines, in the moderate register, plus the streaming-undercount
caveat as an aside and `FoundryGate.Core` in the projects table), `reference/configuration`
(`ResetDayOfMonth` is now read by something and says so; a Function App section for `Storage__*`
and the workspace-GUID requirement), `reference/infrastructure` (the lock's use of the storage
role), `why-foundrygate` and `SystemDiagram` ("on the 1st" → "on the day you set" — verified the
diagram's other labels are still true). Plan 10 rewritten: it still described suspend/re-enable
enforcement and App Insights `customMetrics`, both superseded by the #10 direction update.


kolatts:
**REQUEST CHANGES**

Best-argued PR in the repo — the Core extraction is the right call, the `ValidateRecursively`-stops-at-assembly-boundaries catch was genuinely excellent, and the tests are the kind you actually want. But three of the design decisions the PR asks me to bless don't survive contact with the code, and the reconciliation arithmetic has two separate ways to be wrong.

---

**Issues**

- **Major** `src/FoundryGate.Functions/Services/Usage/UsageSyncJob.cs:86` — D-016's guard is `usage.Count == 0 && updated == 0`, but the query returns month-to-date *totals*, so after the first request of the month `usage.Count` is never zero and every tick writes a `usage.synced` row regardless. I ran five consecutive ticks with four no-ops and got five audit rows — the ~35k/year the decision exists to prevent. Gate on "nothing changed", not "nothing was reported".

- **Major** `src/FoundryGate.Functions/Kql/UsageBySubscription.kql:19` — `ApiManagementGatewayLlmLog` emits **multiple entries per `CorrelationId`** ([docs](https://learn.microsoft.com/azure/api-management/api-management-howto-llm-logs): "requests and responses, including chunked messages for large requests and responses, appear in separate log entries that you can correlate by using the `CorrelationId` field" — hence the `SequenceNumber` column), so `sum(TotalTokens)` over an un-deduped join can multiply every developer's usage, and the overview's new "these totals are a floor, not a ceiling" aside would be exactly backwards. Collapse to one row per `CorrelationId` before joining and `distinct` the right side — two lines, and the question stops mattering.

- **Major** `src/FoundryGate.Functions/Services/FunctionsServiceCollectionExtensions.cs:47` — "a reset re-resolves inputs nobody changed and so cannot produce a tier change" holds only for *per-user* inputs. `PUT /config` on `DefaultMonthlyTokenQuota` re-resolves nothing, so the next scheduled reset moves every default-tier user's `TierProductId`/`AllocatedTokens` in SQL while `NullGatewayTierSync` records it at **Debug** — under a "No gateway is configured" message that is false in this host — and APIM keeps enforcing the old product. The dashboard says 20M, the gateway 403s at 5M, and nothing above Debug says so. (Same hole, smaller blast radius, for a user with no prior-period row: `previousTier` is null, so the sync fires — your own `MonthlyResetJobTests` comment concedes this.)

- **Major** `src/FoundryGate.Functions/Services/Usage/UsageSyncJob.cs:25` — `BillingPeriod.Current` means that from 00:00 UTC on the 1st nothing ever revisits the closed month, so the final pass's window *plus* Log Analytics ingestion lag is permanently missing from it. For a figure the KQL header calls "billing-grade", that's a gap; re-running the previous period for the first day or so of a new month closes it.

- **Major** `src/FoundryGate.Functions/Services/Quota/MonthlyResetJob.cs:52` — the gate is `today != configuredDay` with no record of whether this period was already reset, so one transient storage blip or exception at 00:01 loses the month's reset entirely — while `BlobResetLock.cs:67`, `IResetLock.cs:18` and `MonthlyQuotaResetFunction.cs:25` all promise "the next tick retries". `today >= configuredDay && not yet reset this period` makes all three comments true and costs one more query.

- **Minor** `src/FoundryGate.Functions/Services/Quota/BlobResetLock.cs:63` — only `RequestFailedException` is caught, so an `AuthenticationFailedException` (expired federated token, role assignment not yet propagated) escapes and fails the whole reset, which is exactly what `IResetLock`'s "it does not throw" contract exists to prevent.

- **Minor** `src/FoundryGate.Functions/Services/Quota/BlobResetLock.cs:53` — Azure caps a fixed-duration lease at 60s and nothing renews it, so on a large fork the mutual exclusion lapses mid-reset and `DisposeAsync`'s release then fails too. "The duration is a ceiling, not a deadline" is true for crash recovery, not for the guarantee the type exists to provide — worth either an infinite lease with explicit release, or a renewal timer, or a comment that says what actually happens.

- **Minor** `src/FoundryGate.Functions/Extensions/OpenTelemetryExtensions.cs:54` — CONVENTIONS §Configuration & auth asks for "HttpClient + EF Core instrumentation, `RecordException`"; the worker gets the three exporters and neither instrumentation, so two jobs whose entire workload is SQL plus one HTTP call emit spans for none of it. ASP.NET Core is fairly N/A here; the other two aren't.

- **Minor** `src/FoundryGate.Functions/Services/Usage/LogAnalyticsUsageQueryClient.cs:43` — every one of the 96 daily passes scans and joins the whole month to date, so a pass on the 28th does ~28× the work of one on the 2nd to answer the same question. Also `| order by TotalTokens desc` sorts a result set the caller dictionary-izes on arrival.

- **Minor** `docs-site/src/content/docs/reference/configuration.mdx:21` — #186 leaves this same table row saying `ResetDayOfMonth` is "Always `"1"` in v1 — changing this is unsupported and reserved for a future release." Whoever merges second owns reconciling that, and the answer should be this PR's.

- **Nit** `src/FoundryGate.Functions/Services/Usage/LogAnalyticsUsageQueryClient.cs:79` — `LogsTableRow.GetInt64(name)` throws when the column isn't in the returned table; only a *null value in a present column* returns null. So "an absent column is a schema problem the caller's counts will surface" is the one case the helper doesn't actually survive.

- **Nit** `docs-site/src/content/docs/reference/configuration.mdx:113` — "two settings of its own", followed by four rows.

- **Nit** — nothing asserts the Functions DI graph resolves. I built it with `ValidateOnBuild`/`ValidateScopes` and it's clean, and it picks `BlobResetLock` vs `NullResetLock` correctly off `AzureWebJobsStorage__accountName` — a 20-line test would keep it that way.

---

**Verified locally** on `89789a2`: `dotnet build -c Release` clean, 920/920 predeployment green, `dotnet format --verify-no-changes` clean, `docs-site` build clean, `functions.metadata` shows both triggers with the right 6-field NCRONTAB schedules and `runOnStartup: false`, and the Api image builds with the new Core copy stage. Namespace move is clean — no stale `using`s or dead crefs anywhere in `src/`, and `--follow` shows the history survived. `Gateway__*` / `ConnectionStrings__FoundryGate` / `AzureWebJobsStorage__accountName` all match what `control-plane.bicep` sets, `DOTNET_ENVIRONMENT` included, and the Log Analytics Reader + Storage Blob Data Owner assignments the new code needs are already in `control-plane-rbac.bicep`.

**Merge order**, for whoever is holding the queue: this one first — it creates Core and does the namespace move, and every conflict it causes elsewhere is a one-line `using`. Then #184 (its `CommitToken` work on `QuotaAllocationService.ResetAsync` now belongs on `QuotaResetService`, which already derives the commit point from `TierSyncRequested`). Then #186, fixing the `ResetDayOfMonth` docs row on the way in.

kolatts:
Thank you — every Major was right, and three of them were things I had asserted in prose and never actually checked. Point by point.

---

### Major 1 — the idle guard never fired (`UsageSyncJob.cs:86`) ✅ fixed

You're right, and the bug is worse than a missed optimisation: `usage.Count == 0` can only be true before the month's first request, so D-016 was decorative from the second tick of every month onward. I had written the guard against the shape of a *delta* query while writing a *totals* query three lines up.

Now: audit when a `TokensUsed` value actually moved, **or** when the problem counts — unknown subscriptions, drift, developers with no allocation row — differ from the last recorded pass, read back from that row's own details JSON.

Two things fell out of getting "new" right, both of which I'd have got wrong with a simpler gate:

- gating on the raw counts instead would write a row every 15 minutes for as long as one unknown subscription existed — the same flood, differently caused;
- gating on `updated` alone would silently swallow the first pass on which an admin's quota cut put someone over budget *without their usage changing*. That one is now its own test (`A_newly_over_budget_developer_is_recorded_even_though_their_usage_did_not_move`).

Your five-tick probe is pinned as `Five_consecutive_ticks_with_four_no_ops_write_one_audit_row`. D-016 amended with what was wrong and why.

### Major 2 — the KQL multiplied usage (`UsageBySubscription.kql:19`) ✅ fixed

Confirmed against both docs. The how-to says entries are separate and correlated by `CorrelationId`; the table reference settles the ambiguity that mattered — `PromptTokens: The number of prompt tokens used by the **request**`, not by the entry. So the same counts repeat across an exchange and `sum()` multiplied by however many entries a request produced. And you're right that the aside would then have been precisely backwards.

Fixed by collapsing to one row per `CorrelationId` before the join, and `distinct`-ing the gateway-log side so a duplicate there can't re-multiply.

I went with `max()` per column rather than `arg_max(SequenceNumber, *)`, and wrote the reasoning into the `.kql` header: `max` is right whether the counts repeat on every entry, sit only on the response entry, or only on the last one, whereas `arg_max` is right only in the last case and picks a token-less entry whenever the highest sequence number belongs to a chunked *request* rather than the response. The one shape `max` under-counts — genuinely partitioned per-chunk counts — is ruled out by the "used by the request" wording, and #178 now has a dedicated section (1a) that verifies it against a live gateway with an entries-per-CorrelationId probe before these numbers are trusted for billing. D-017 records it.

The overview aside is rewritten: it now explains the collapse and reserves "floor" for the case that survives it — token usage the gateway never logged at all, which the same MS doc calls out for broken streams.

### Major 3 — `NullGatewayTierSync` was the wrong seam (`FunctionsServiceCollectionExtensions.cs:47`) ✅ fixed

You're right and I'd argue it's the most serious of the five, because it's silent. My claim was true of *per-user* inputs and I generalised it without checking level 5. `PUT /config` on `DefaultMonthlyTokenQuota` re-resolves nobody, so the next scheduled reset moves every default-tier user in SQL, at Debug, under a "No gateway is configured" message that is false in that host.

- New `Core/Quota/WarningGatewayTierSync` — logs a Warning naming the user, the target tier and the subscription, and says explicitly that the API will reconcile on the user's next quota touch.
- The reset's audit row now carries `tierChangeCount`, so the trail records how many users are affected, not just the log stream.
- `MonthlyResetJobTests` now has both halves: `A_changed_DefaultMonthlyTokenQuota_moves_tiers_on_the_next_reset_and_is_reported` (the case I'd claimed impossible) and `A_reset_over_unchanged_inputs_moves_nobodys_tier` (the claim, scoped to what it is actually true of).

Filed both follow-ups: **#193** — `PUT /config` must re-resolve the users it affects, in the Api where `ApimGatewayTierSync` lives; **#194** — move the APIM management client into Core so this host can do the move itself (with a note that it widens the Functions identity to APIM Service Contributor, which is a real trade-off worth deciding rather than assuming). `WarningGatewayTierSync`'s XML doc points at both.

### Major 4 — the closed month was never revisited (`UsageSyncJob.cs:25`) ✅ fixed

Fair, and "billing-grade" was my word. For the first three days of a month (`PreviousPeriodGraceDays`, a documented constant) the pass reconciles the previous period as well, so the old month's final window plus ingestion lag lands. Tests at 00:30 on the 1st and just past the window. `UsageSyncOutcome.IncludedPreviousPeriod` makes it visible, and the audit row lists `periodsReconciled`.

### Major 5 — a missed day was a missed month (`MonthlyResetJob.cs:52`) ✅ fixed

Guilty on all three comments. Gate is now `today >= configuredDay && not yet reset this period`, where "already reset" is the existence of a `quota.monthly-reset` audit row inside the period's date window.

One deliberate deviation from the letter of the suggestion: the predicate is a date bound rather than a substring match on the details JSON. A run inside period P necessarily has an `OccurredDate` inside P — both come from the same `TimeProvider` — `OccurredDate` is indexed, and matching `"periodYear":2026` as text would break the day the serializer's formatting changed. `periodYear`/`periodMonth` stay in the details for whoever reads the trail, and the comment says which is load-bearing. A late run logs a Warning naming the day it should have happened. Four new tests: outage → next day resets, already reset → skip, last month's reset doesn't satisfy this month, admin's manual `quota.reset` doesn't suppress the scheduled run. D-018.

### Minor — `BlobResetLock` only caught `RequestFailedException` ✅ fixed

Right, and `AuthenticationFailedException` is the likely one on a fresh deploy. Now catches `Exception` (excluding `OperationCanceledException`, which is the host shutting down and must propagate) and logs at **Error** rather than Warning — unlike a 409 it needs an operator.

### Minor — the lease lapsed mid-reset ✅ fixed

Also right: 60s is Azure's cap for a fixed lease, and my "ceiling, not a deadline" defence covered crash recovery, not the guarantee the type exists for. The handle now renews at a third of the duration for as long as it's held; a failed renewal logs a Warning and stops, leaving the lease to expire, which is the same recoverable state as a crash. Documented on the class.

### Minor — missing OTel instrumentation ✅ fixed

Added HttpClient (with `RecordException`) and EF Core to the worker's tracing, HttpClient to metrics. Agreed ASP.NET Core is N/A here, and the comment says so.

### Minor — whole-month scan per pass ⚠️ partly

Verified it is *not* rescanning the workspace: the `QueryTimeRange` bounds `TimeGenerated` server-side, so the scan is the month, not retention. But you're right it's ~28× on the 28th, and that is inherent to period totals. I've left it and written the trade-off down at the call site: the alternative is deltas, and a delta the job fails to persist is lost forever where a total simply gets recomputed — idempotence is worth more here than a cheaper query. #178 gained a cost check to run late in a month rather than on the 2nd. `| order by` dropped; you were right that sorting a set the caller dictionary-izes is pure waste.

### Nit — `GetInt64` throws on an absent column ✅ fixed

Correct, and my comment claimed the opposite. The client now checks the returned columns once against the query's contract and throws one `UpstreamDependencyException` naming the missing column and what was returned, instead of an opaque per-row throw.

### Nit — `configuration.mdx` "two settings", four rows ✅ fixed

### Nit — nothing asserted the DI graph ✅ fixed

`FunctionsServiceCollectionExtensionsTests`: `ValidateOnBuild` + `ValidateScopes`, every registration resolved from a scope, and the `BlobResetLock`-vs-`NullResetLock` selection off `AzureWebJobsStorage__accountName`, the Azurite connection string, and neither.

### Minor — `ResetDayOfMonth` row vs #186

#186 merged first after all, so I've merged `origin/main` and written the row as current truth on top of their retired-keys rework — it now describes the on-or-after gate as well.

---

## Verification

- `dotnet test` — **1064 passed** (935 before merging main; 15 new tests in this pass)
- `docs-site` build — clean

## One thing you should know about, that isn't mine

While merging `origin/main` I hit a red suite and spent a while assuming it was me. It isn't: `Web/MePageTests.The_rotated_mask_survives_the_next_render`, landed by #174, fails **2 in 5 full-suite runs** on a clean `origin/main` and passes 24/24 when its class runs alone — an ordering/parallelism interaction, not an assertion bug. Filed as **#195** with the repro. Flagging it because it will get blamed on whichever PR is in flight when it goes red, and it nearly got blamed on this one.

## Merge order

Understood, and the queue changed under me: #186 is in and merged here; #184 now goes before me, so I'll rebase onto it and move `QuotaResetService` to its `CommitToken.For(...)` helper (the inline ternary I currently have is exactly what CONVENTIONS is trying to stamp out) before this merges.

kolatts:
### Rebased onto #184

One conflict, exactly where the two PRs were always going to meet: `QuotaAllocationService.ResetAsync`. #184 hardened the in-Api implementation's commit point; this branch had already moved that implementation into `FoundryGate.Core.Quota.QuotaResetService` and left the Api method as a five-line delegation. Resolved in favour of the delegation, with #184's semantics carried into Core rather than lost along with the code they were applied to:

- `QuotaResetService` uses **`CommitToken.For(tierSyncCount > 0, cancellationToken)`** — the inline ternary I had is exactly what CONVENTIONS (as amended by #184) now forbids, and `CommitToken` living in `FoundryGate.Data` rather than the Api is what makes it reachable from Core at all. Good call putting it there; this branch is the first consumer that proves the point.
- The token it returns still reaches the audit row, both saves, **and** the read-back inside `AdoptConcurrentlyCreatedRowsAsync` — the read-back half of the rule, which is the easiest part to drop during a move.
- `IQuotaResetService`'s token doc names `CommitToken.For` and states the negative case: a reset that moved nobody's tier never reached APIM and keeps the caller's own token.
- `UsageSyncJob` gained a short note on why it deliberately does **not** use `CommitToken` — reconciliation's only external call is a *read*, so there is no accepted change the database owes a record of, and an abandoned pass should stop. Recorded so the next reader doesn't "fix" it into `CancellationToken.None`.

Everything else merged clean: `GetMyAllocationAsync`'s read-back-on-the-commit-token fix came through as-is, and `RecordingGatewayTierSync` kept both sides (my `FoundryGate.Core.Quota` namespace plus your `ThrowFor` / `AfterSync`). `User.LastLoginDate` is nullable, so the Functions and Core test seeders needed nothing; `WebTestData` is Web-only and untouched here.

**Gates after the merge:** build 0 warnings / 0 errors, **1087/1087** tests (green on the first run — no #195 flake this time), `dotnet format --verify-no-changes` clean, docs build clean, all CI checks pass.

Issue #196, opened 2026-09-02
Enable Container Apps environment zone redundancy once private networking lands


kolatts:
> Parent: #222 — Owner setup and production readiness

Split out of #134.

`containerAppsZoneRedundant` is now plumbed through `infra/main.bicep` -> `modules/control-plane.bicep` -> `modules/container-app.bicep`, but `prod.bicepparam` sets it `false` and the parameter description says why: ARM only accepts `properties.zoneRedundant: true` on a Container Apps **managed environment that is VNet-integrated** (`properties.vnetConfiguration.infrastructureSubnetId`). FoundryGate's environment has no VNet today.

Zone redundancy is also immutable after creation, so this is not a flip-in-place on the existing prod environment — it is part of whatever change introduces private networking (spec §11) and recreates `cae-foundrygate-prod`.

**When private networking lands:**
- add the infrastructure subnet to `container-app.bicep` (`vnetConfiguration.infrastructureSubnetId`, /23 minimum for a workload-profiles environment),
- set `param containerAppsZoneRedundant = true` in `infra/parameters/prod.bicepparam`,
- update the prod row in `docs-site/src/content/docs/reference/cost-and-capacity.mdx` (zone redundancy itself is free on Container Apps; the VNet and any private endpoints are not),
- and note the recreate in the runbook — the Container App's ingress FQDN changes with a new environment.

Issue #138, opened 2026-09-01
UI PR previews: SWA staging hostnames need SPA redirect URIs for MSAL sign-in


kolatts:
> Parent: #222 — Owner setup and production readiness

`ui-deploy.yml` (#73) creates a Static Web Apps **staging environment** per pull request touching `src/FoundryGate.Web/**` (hostname `stapp-foundrygate-dev-<pr>.<region>.azurestaticapps.net`, posted on the PR by `Azure/static-web-apps-deploy`).

MSAL sign-in on those previews fails until the preview hostname is a registered SPA redirect URI on the FoundryGate.Web app registration — Entra does not allow wildcard redirect URIs for SPAs. Options:

1. Add a fixed set of preview redirect URIs (SWA names staging environments by PR number; Free tier allows 3 concurrent staging envs, so the numbers are not predictable).
2. Have the preview job register the hostname on the app registration (`az ad app update --spa-redirect-uris`) and remove it on close — needs `Application.ReadWrite.OwnedBy` on the CI principal.
3. Document previews as "unauthenticated UI shell only" and rely on the deployed dev site for auth flows.

Decide and implement; the workflow summary already points here.

kolatts:
## Unblocked by #198 (#155) — and the redirect URIs are only half of it

PR previews are back (`ui-deploy.yml`'s `preview` / `close-preview` jobs, under the `ui-preview` Environment), so this issue is live again rather than moot.

While wiring it up, a second reason a preview cannot exercise the signed-in app turned up, and it belongs on this issue because fixing only the redirect URIs would leave the preview just as unusable:

**CORS.** `infra/modules/control-plane.bicep` sets exactly one allowed origin on the API:

```bicep
{ name: 'Cors__AllowedOrigins__0', value: 'https://${staticWebApp.outputs.defaultHostname}' }
```

That is the dev site's own hostname. Every SWA preview gets a *different* hostname (`<name>-<slot>.<region>.azurestaticapps.net`), so even with a valid token every browser call from a preview is blocked before it reaches the API. Today a preview is honest for layout and for routes that never call the API, and nothing else.

**Both halves are needed, and they are asymmetric:**

[...]

Issue #109, opened 2026-09-01
Owner setup for the control plane: Entra app registrations, dedicated SQL admin group, CI principal group membership, GitHub variables


kolatts:
> Parent: #222 — Owner setup and production readiness

Manual (non-Bicep-able) steps the control-plane infra from #43/#44 depends on. **These are `needs-human` because of *privilege*, not tooling** — an agent has the Azure and GitHub CLIs and can run every `az ad app`, `az role assignment` and `gh variable` command below verbatim. What it does not have is Global/Privileged Role Administrator in the tenant, Owner on the subscription, or the standing to consent on the owner's behalf. Run them yourself, or grant the privilege deliberately and hand an agent the exact command list.

1. **FoundryGate.Api app registration** (spec §12): expose `api://{clientId}/access_as_user`, app role `FoundryGate.Admin`. Put its client id in the GitHub Environment variable `FG_ENTRA_API_CLIENT_ID` (dev + production) — `infra/parameters/{dev,prod}.bicepparam` read it with `readEnvironmentVariable('FG_ENTRA_API_CLIENT_ID', '<zero guid>')`; the zero GUID lets a bootstrap deploy succeed but rejects every token until the real id is set.
2. **Blazor SPA app registration** for `FoundryGate.Web` with the SWA hostname (`stapp-foundrygate-{env}.<n>.azurestaticapps.net`, output `staticWebAppHostname`) as redirect URI; update `src/FoundryGate.Web/wwwroot/appsettings.json` placeholders.
3. **SQL Entra admin group**: the param files currently point at `SG_IMAGILE_SQL_ADMINS` (`2ed4d6b7-575c-4046-aeb0-eb51bc254ef5`). Create a dedicated `SG_FOUNDRYGATE_SQL_ADMINS` (and ideally a separate prod group), update `sqlAdminGroupObjectId`/`sqlAdminGroupName` in `infra/parameters/*.bicepparam`.
4. **Add the CI OIDC app registration** (the one `azure/login@v2` uses) to that group — otherwise `_deploy-database.yml` cannot connect with `Authentication=Active Directory Default`. Azure SQL is Entra-only; there is no password fallback by design (CONVENTIONS.md).
[...]

PR #198, opened 2026-09-02
ci/infra: database OIDC via Environment vars, digest-pinned image, prod Bicep knobs, SWA previews, production approval ruling

### LINKED ISSUE #134: Plumb production-grade knobs through control-plane.bicep (zone redundancy, storage/ACR SKU, container CPU/memory)
Follow-up from the PR #111 review (Nit). `infra/parameters/prod.bicepparam` calls itself production-grade, but four leaf-module parameters are not exposed by `infra/modules/control-plane.bicep` / `infra/main.bicep`, so prod gets the leaf defaults: `zoneRedundant` (sql.bicep database, container-app.bicep environment — both `false`), `skuName` on storage-account.bicep (`Standard_LRS`), `sku` on container-registry.bicep (`Basic`), and `cpu`/`memory` on container-app.bicep (`0.25`/`0.5Gi`, sized to the published cost page). Deliberately v1-scoped and stated in prod.bicepparam.

Ask: add `sqlZoneRedundant`, `containerAppsZoneRedundant`, `functionsStorageSku`, `containerRegistrySku`, `containerAppCpu`/`containerAppMemory` to main.bicep (defaults = today's behaviour), pass them through the orchestrator, set prod values in prod.bicepparam, and update the prod cost row in `docs-site/src/content/docs/reference/cost-and-capacity.mdx` (ZR roughly doubles the SQL compute line). Note Container Apps env zone redundancy requires a VNet-integrated environment, so that one may stay out until private networking (spec §11) lands.

### LINKED ISSUE #137: _deploy-database.yml: read OIDC ids from Environment variables (vars.*) instead of secrets inputs
`.github/workflows/_deploy-database.yml` takes `AZURE_CLIENT_ID` / `AZURE_TENANT_ID` / `AZURE_SUBSCRIPTION_ID` as `secrets:` inputs. The rest of the #67 pipeline (`_deploy-infra.yml`, `_deploy-api.yml`, `_deploy-functions.yml`, `_deploy-ui.yml`, `_postdeployment-tests.yml`) reads them as **GitHub Environment variables** (`vars.*`, per #68 / plans/22 — they are identifiers, not secrets, and masking GUIDs makes resource ids unreadable in logs).

Because a `uses:` job cannot declare `environment:` it also cannot read Environment-scoped variables, so `_prepare-database.yml` currently bridges: it runs with `environment:` set, emits the three ids as job outputs, and the callers (`api-deploy.yml`, `deploy-all.yml`) pass them into `_deploy-database.yml`'s `secrets:` map.

**Ask:** make `_deploy-database.yml` read `vars.AZURE_CLIENT_ID` etc. itself (its `deploy` job already declares `environment: ${{ inputs.environment }}`), ideally via `.github/actions/azure-oidc-login`, and drop the `secrets:` block. Then remove the three `azure-*` outputs from `_prepare-database.yml` and the `secrets:` maps in both callers. Owned by whoever holds `_deploy-database.yml` (#96 / #106 agent) — the CI/CD PR for #68–#75 deliberately did not edit that file.

Also from that PR: `_deploy-database.yml` receives the **GitHub Environment name** (`dev` | `production`) as `environment` and passes it straight to `foundrygate ip setup --env`. The FoundryGate/Bicep environment name for production is `prod` (`rg-foundrygate-prod`, `sql-foundrygate-prod-<suffix>`), so `ip setup` must map `production` → `prod` (or the workflow must, before calling the CLI). Flagged on #96 as well.

### LINKED ISSUE #141: Production deploys need one Environment approval per stage — decide single-gate vs per-stage
GitHub approves **pending jobs**, not whole runs: every job that targets the `production` Environment waits for its own approval when it becomes pending. In `deploy-all.yml` for production that is one click per wave — Infra → Database prepare → Database → (API + Functions + UI together) → Postdeployment tests — five approvals for one full-stack deploy. `infra-deploy.yml` dispatch (dev-then-production) is one.

Acceptable for now (documented in docs `reference/ci-cd`), but worth revisiting once production exists:

- Option A: a single `approve-production` job with `environment: production`, then run the actual deploy jobs against an unprotected `production-deploy` Environment that carries the same variables and a federated credential of its own. One approval per run; the cost is a duplicated variable set.
- Option B: keep per-stage approvals (each is a real checkpoint: the infra what-if is visible before the DB/app stages run).

Decide when the first production deploy is scheduled.

### LINKED ISSUE #155: Restore Static Web Apps PR previews behind a safe identity
The UI PR preview jobs (`ui-deploy.yml` `preview` / `close-preview`, `_deploy-ui.yml` `action: close`) were removed in #144.

Why: a preview needs the SWA deployment token, which is read at run time with `az staticwebapp secrets list` under an identity that can write to the `dev` environment. A `pull_request` run executes the PR branch's own copy of the workflow files, so any PR could have added a step that printed that token or drove `az` against dev. As part of the same PR the `dev` GitHub Environment was restricted to protected branches, which blocks the preview jobs outright; PR-track Azure work is now limited to the read-only `dev-plan` Environment (Bicep what-if only).

To restore previews, the design needs an identity whose blast radius is one Static Web App's staging environments, reachable from an unprotected branch. Options:

- A `dev-preview` GitHub Environment with its own federated credential on an app registration whose only role assignment is on `stapp-foundrygate-dev` (Static Web Apps `Contributor` scoped to that one resource). Still leaks a token that can publish to that SWA's staging slots, which is a much smaller prize than Owner on the subscription.
- Keep the token out of the PR run entirely: build the artifact on `pull_request`, then publish it from a `workflow_run` job on the base branch's copy of the workflow.
- Drop previews and rely on `dotnet run` locally.

Blocks #138 (preview hostnames need SPA redirect URIs), which is moot until previews exist again.

### LINKED ISSUE #157: Pin the API image base tags by digest and tighten .dockerignore
From the #144 review, deferred out of that PR to keep it to the pipeline shape.

`src/FoundryGate.Api/Dockerfile` uses floating `mcr.microsoft.com/dotnet/sdk:10.0` and `aspnet:10.0` tags with no digest pin, so two builds of the same commit can produce different images. Pin `FROM ... @sha256:...` (with the tag kept in a comment for readability) and add a renovate-style note about how to refresh them.

`.dockerignore` also does not exclude `src/FoundryGate.Web`, `src/FoundryGate.Functions`, `src/FoundryGate.Cli`, `src/FoundryGate.Database` or the test projects, none of which the image `COPY`s. Excluding them shrinks the build context. Verify with `docker build` that nothing in the API build graph actually needs them first (`Directory.Build.props`, `Directory.Packages.props`, `global.json` and `NuGet.config` must stay).

Verification: `docker build -f src/FoundryGate.Api/Dockerfile .` still succeeds, the container still answers `/health` with 200 on 8080 as `uid=1654(app)`, and the reported build context is smaller.

kolatts:
Five infra/CI issues in one sweep. Nothing here deploys anything; the only live Azure call made while writing it was a read-only `what-if`.

Closes #137
Closes #157
Closes #134
Closes #155
Closes #141

---

## #137 — `_deploy-database.yml` reads `vars.*`, not `secrets:`

Its `deploy` job already declares `environment: ${{ inputs.environment }}`, so it can read Environment variables itself. It now does, through the shared `.github/actions/azure-oidc-login` guard, like every other `_deploy-*.yml`. Consequences:

- the `secrets:` block is gone, and with it the bespoke "Verify OIDC permissions" step (the composite's guard is strictly better — it also checks the variables and points at #109);
- `_prepare-database.yml` drops the `azure-client-id` / `azure-tenant-id` / `azure-subscription-id` bridge outputs;
- `deploy-all.yml` and `api-deploy.yml` drop their `secrets:` maps.

The `workflow_call` `inputs:` contract is otherwise untouched.

The issue's second point — `ip setup --env production` having to mean the Bicep environment `prod` — turned out to be **already handled**: `FoundryGateAzureResources.NormalizeEnvironment` maps `production` → `prod` (and `development` → `dev`). No workflow-side mapping needed; documented in `reference/ci-cd`.

## #157 — base images pinned by digest, `.dockerignore` tightened

`sdk:10.0` and `aspnet:10.0` are now `FROM ...@sha256:...` (multi-arch manifest **list** digests, so amd64 and arm64 both still resolve), with the tag kept in a comment beside each digest and a refresh recipe in the header. `.dockerignore` additionally excludes the six projects the image never `COPY`s — Web, Functions, Cli, Database, both test projects — plus `FoundryGate.sln` and `GitVersion.yml`. The five root files the build does need (`global.json`, `Directory.Build.props`, `Directory.Packages.props`, `NuGet.config`, `.editorconfig`) stay.

## #134 — production knobs plumbed through `control-plane.bicep`

`main.bicep` → `control-plane.bicep` → leaf modules, all defaulting to today's behaviour so dev/test are byte-identical:

| Parameter | default (dev) | prod |
|---|---|---|
| `sqlZoneRedundant` | `false` | `true` |
| `containerAppCpu` / `containerAppMemory` | `0.25` / `0.5Gi` | `0.5` / `1.0Gi` |
| `functionsStorageSku` | `Standard_LRS` | `Standard_ZRS` |
| `containerRegistrySku` | `Basic` | `Standard` |
| `containerAppsZoneRedundant` | `false` | `false` |

`containerAppsZoneRedundant` is deliberately plumbed-but-off: ARM only accepts `zoneRedundant: true` on a **VNet-integrated** Container Apps environment, and the property is immutable after creation, so enabling it belongs to the private-networking change (spec §11). `container-app.bicep` takes the parameter instead of hardcoding `false`, so it becomes a one-line flip — split out as **#196**.

Cost page numbers changed, so it was updated; prices re-verified against the Azure Retail Prices API (eastus2, 2026-09-02): GP Gen5 2 vCore `$0.304434/hr`, zone-redundancy surcharge `$0.18266/hr` (**+60%**, not the "roughly double" the issue guessed), ACR Basic `$0.1666/day` vs Standard `$0.6666/day`. The production table also still described GP *serverless*, which prod has not used since #111 — it now matches `prod.bicepparam` row for row. Total moves ~$450 → ~$660 (Basic v2) and ~$1,020 → ~$1,210 (Standard v2). (The Container Apps row was corrected in the review pass: three replicas at 0.5 vCPU / 1.0 GiB is ~$118/mo, not $70.)

## #155 — Static Web Apps PR previews, behind a one-resource identity

A `pull_request` run executes the PR branch's own copy of the workflow files, so the workflow can never be the boundary — only the identity can. Design:

- a new **`ui-preview`** GitHub Environment (already created via the API: no protection rules, any branch — a branch policy would stop previews the way one on `dev-plan` would stop what-ifs);
- its federated credential is meant to hold **exactly one** role assignment: the custom role **`FoundryGate SWA Preview Publisher`**, defined in `infra/modules/swa-preview-role.bicep` and assignable only on `stapp-foundrygate-dev`. There is no built-in role for this — no built-in grants *any* `Microsoft.Web/staticSites` action, so the built-in answer would have been plain Contributor. **Owner action, on #109 with the exact commands.**

Two invariants make that scoping sufficient, and they are in the file header for whoever reviews the next change to it:

1. **The preview jobs make no subscription-scope call.** No `az deployment sub show`. The SWA name, resource group, API base URL and MSAL ids are `ui-preview` Environment variables. An identity with a role on one resource could not read the deployment outputs anyway — requiring that it cannot is the point.
2. The only Azure call is `az staticwebapp secrets list` against that one app; the only Azure write is the deploy action's upload/close of the PR's own staging slot.

The previews live in `ui-deploy.yml`, **not** `_deploy-ui.yml` — that file is the production deploy path and runs under an identity with subscription access. Its header now says so.

Fork safety: missing variables and a missing OIDC token are both a soft skip with a `::notice::`, so a fork PR goes green. `close-preview` runs on `pull_request: closed` so slots do not accumulate. Sign-in on a preview hostname still fails until the SPA app registration gains the preview redirect URIs — **#138**, now unblocked.

## #141 — ruling: keep per-stage production approvals

**Six** approvals per full-stack production run (infra deploy · database prepare · database deploy · API · Functions+UI together · tests), and they stay. (The first draft said five, counted off the stage diagram; the real count is per *gated job* and was seven — `_deploy-api.yml` had two, now merged into one. See the review response.)

**Why not one gate.** Option A in the issue — one `approve-production` job, then the real work against an unprotected `production-deploy` Environment carrying the same variables and its own federated credential — does not collapse the gate, it *removes* it: any job in any workflow could then target `production-deploy` and mint a production-Owner token with no reviewer in the path. The variant that keeps the credential inside the approved job and passes it downstream is worse — an Azure access token in a job output is a bearer credential sitting in the API. And OIDC offers no third option: the token is minted per job and the Environment name is baked into its `subject`, so a job that does not declare `environment: production` cannot federate to production at all. "Approve once, then run on plain runners with the identity from an output" is not something that can be built, only something that can be faked by leaking a token.

**Why per-stage is right anyway.** Each stage is independently re-runnable, and stage 3 (dacpac + seed + the SQL firewall window) is the irreversible one; a human reading the infra result before approving the schema migration is the checkpoint that makes a separate `dev-then-production` chain unnecessary.

**What changed instead.** `deploy-all.yml` opens with an ungated, Azure-free `plan` job that prints the stage graph and — for production — a table of the five approvals and what each one buys, into the run summary *before* the first gate. Recorded as **D-019** in `fable-refactor-log.md` and documented in `reference/ci-cd`.

---

## Verification

| Gate | Result |
|---|---|
| `actionlint 1.7.7` (shellcheck 0.10.0 on `PATH`, so every `run:` block is shellchecked) | clean, 18 files |
| `shellcheck -x` on all five `.github/scripts/infra/*.sh` | clean |
| `bash .github/scripts/infra/resolve-api-image.test.sh` | 11 passed, 0 failed |
| `az bicep build --file infra/main.bicep` | clean (3 pre-existing `BCP081` warnings on `Microsoft.CognitiveServices@2026-07-01`) |
| `az bicep lint` on `main.bicep` + all 15 modules | clean |
| `az bicep build-params` for `test` / `dev` / `prod` | clean; prod renders all five new knobs, dev renders none (defaults hold) |
| `az deployment sub what-if` (**read-only**) with `prod.bicepparam` | `Succeeded`, 59 creates. ARM accepted every knob: database `zoneRedundant: true` + `GP_Gen5_2` + Geo, container `{cpu: 0.5, memory: '1.0Gi'}`, registry `Standard`, storage `Standard_ZRS`, managed environment `zoneRedundant: false` |
| `docker build -f src/FoundryGate.Api/Dockerfile .` | succeeds on the pinned digests |
| container smoke test | `/health` → `200 {"status":"Healthy","version":"0.0.1-local","environment":"local"}` on 8080, running as `uid=1654(app)` |
| `npm ci && npm run build` in `docs-site/` | 15 pages built |

**Expected on this PR's own checks:** the `dev-plan` what-if in `infra-deploy.yml` will skip or soft-fail — that Environment's Reader identity is not configured yet (#109), which is the designed loud-but-green path (`azure-oidc-login` soft-skips on every `pull_request` event). Same for the new `ui-preview` preview job: `ui-preview` exists but has no identity yet, so it emits a `::notice::` and goes green. `ci.yml`'s `build-test` is untouched.

## Decisions for the reviewer

1. **`ui-preview` is a third app registration, not a reuse of `dev-plan`.** `dev-plan` is subscription-Reader; adding a write role to it would give every PR what-if the ability to publish to the SWA. Separate identity, separate blast radius.
2. **Previews are in `ui-deploy.yml`, not parameterised into `_deploy-ui.yml`.** Adding an `action: close` branch to the production deploy path would have put PR-reachable code in the file that runs under subscription-scope credentials. One extra file's worth of duplication buys a reviewable boundary.
3. **`containerAppsZoneRedundant` ships as a parameter that is false everywhere.** The alternative was leaving it hardcoded until #196. Plumbing it now means the private-networking PR changes one line in `prod.bicepparam` instead of three files.
4. **The cost page's production table was rewritten, not patched.** It described a SQL SKU prod stopped using in #111; leaving that while adding a ZR line would have compounded the error.
5. **Digests are manifest-list, not per-platform.** A per-platform digest would pin CI to amd64 and break anyone building on arm64.

## Follow-ups filed

- **#196** — enable Container Apps environment zone redundancy once private networking lands (needs `infrastructureSubnetId`; the property is immutable, so it recreates `cae-foundrygate-prod` and changes the ingress FQDN).
- **#212** — automate the base-image digest refresh (Renovate/Dependabot pointed at MCR, resolving manifest-list digests).
- **#138** — gained the preview-CORS half: the API allows exactly one origin, so redirect URIs alone would not make a preview usable.

## Review pass (`34218f6`, `a254452`, merge `8c4ce9e`)

Four Majors, six Minors and four Nits from the review are addressed; the point-by-point response is in the comments. The two that changed behaviour rather than wording: `_deploy-api.yml`'s two gated jobs became one (the approval count was wrong in four files), and the API image was shipping `/app/appsettings.local.json` past a case-sensitive `.dockerignore` pattern — which also meant the `/health` smoke test only passed *because* of the leak. `build-api-image` now inspects the built image for developer configuration and supplies the real settings from the environment, as Container Apps does. `main` is merged in (23 commits, no conflicts).


kolatts:
**REQUEST CHANGES**

Genuinely strong sweep — the #137 bridge removal is clean, the digests really are manifest lists, and D-019's "OIDC bakes the Environment into the token subject, so an ungated `production-deploy` is not a gate" reasoning is correct and worth having written down. The problem is that the two things this PR exists to *produce* — the approval table and the `ui-preview` role assignment — are both wrong, and one of them is a role that does not exist.

I reproduced every gate in the Verification table on `0ab35c8` and they all pass: actionlint clean, `az bicep build`/`lint`/`build-params` clean for test/dev/prod, `docker build` + `/health` 200 as `uid=1654(app)`, docs 15 pages. Both base-image digests are genuine multi-arch manifest lists (amd64 + arm/v7 + arm64). A no-cache context probe confirms `.dockerignore` excludes nothing the Dockerfile `COPY`s. Both `_deploy-database.yml` callers still pass `environment:` and `id-token: write`, and no dangling `azure-*` output reference survives. No conflict with `main` or either other open PR.

---

**Issues**

- **Major** `.github/workflows/deploy-all.yml` lines 117-127 (and `reference/ci-cd.md` 179-192, `fable-refactor-log.md` D-019, `plans/22`) — A production run asks **seven** times, not five: `_deploy-api.yml` has *two* environment-gated jobs (`build-push` at :46 and `deploy` at :111), and `functions`/`ui` both `needs: [database, api]`, so row 4's "three jobs, one visit — they go pending together" cannot happen. The single deliverable of #141 is a reviewer knowing how many approvals are coming, and the table that prints it is wrong in four files.

- **Major** `.github/workflows/ui-deploy.yml` line 21, `reference/ci-cd.md` line 166, and the #109 comment — **"Static Web App Contributor" is not an Azure built-in role.** `az role definition list --name "Static Web App Contributor"` returns `[]`, and no built-in role grants any `Microsoft.Web/staticSites` action at all (the operation `Microsoft.Web/staticSites/listsecrets/action` is real; nothing built-in holds it outside `*`). The `az role assignment create` command in #109 will fail, and the thing anyone reaches for at that point is plain Contributor. A custom role with `read` + `listsecrets/action`, or Contributor scoped to that one `staticSites` resource, is what the design actually needs.

- **Major** `.github/workflows/ui-deploy.yml` lines 21-32 — The SWA deployment token is **app**-scoped, not slot-scoped: `swa deploy --deployment-token` publishes to the app's production environment from anywhere. So "the worst a malicious PR can then do is publish to that Static Web App's staging slots" isn't the worst case — it's overwriting `stapp-foundrygate-dev`'s live site, which is a registered MSAL redirect origin for the dev SPA (and #138 is about adding more of them). Probably still an acceptable blast radius, but this header is the control, so it should say the real thing.

- **Major** `docs-site/src/components/DeployPipeline.astro` line 74 and `docs-site/src/content/docs/architecture/overview.mdx` lines 290-302 — CONVENTIONS §"Architecture diagrams" requires the diagram components to track the deploy pipeline flow, and the component still says *"Nothing on the pull request track can change Azure."* with no preview stage and no `plan` stage; `overview.mdx` still says *"five GitHub Environments"* and *"nothing a pull request can run reaches [dev]"*. `reference/ci-cd.md` was carefully rewritten away from exactly that claim, which leaves these two as the only places still asserting the retired version.

- **Minor** `.dockerignore` line 44 — `**/appsettings.Local.json` does not match `src/FoundryGate.Api/appsettings.local.json` (dockerignore patterns are case-sensitive), so the image ships `/app/appsettings.local.json` containing `User Id=sa;Password=Temp1234!`. Never loaded at `ASPNETCORE_ENVIRONMENT=qa|prod`, but it's precisely what that line was written to stop and it will light up any image scanner.

- **Minor** `.github/workflows/ui-deploy.yml` lines 209-219 — `stapp-foundrygate-dev` is `Free`, which allows three staging environments. The fourth concurrent UI PR fails *after* the token fetch has already succeeded, so none of the soft-skip guards catch it and the check just goes red.

- **Minor** `.github/workflows/ui-deploy.yml` lines 39-41 and 241-242 — `control-plane.bicep:242` sets exactly one `Cors__AllowedOrigins__0`, the dev SWA's own hostname. Every browser call from a preview hostname is CORS-blocked, so "unauthenticated routes and layout are real" holds only for pages that never touch the API — and #138 alone won't change that.

- **Minor** `infra/main.bicep:180`, `infra/modules/control-plane.bicep:59`, `infra/parameters/prod.bicepparam:54` — All three still say zone redundancy "roughly doubles" / "costs roughly the compute line again" after the cost page was correctly rewritten to +60% (retail eastus2 confirms: $0.15222 + $0.09133 per vCore-hr). Same drift in the cost table's Container Apps row — three replicas at 0.5 vCPU / 1.0 GiB is ~$118/mo, not $70.

- **Minor** `infra/parameters/prod.bicepparam:70`, `infra/main.bicep:209` — "Flipping this to true is then a one-line change" / "that change is one parameter". There's no `vnetConfiguration` or `infrastructureSubnetId` anywhere in `infra/`, and #196 itself says the flip recreates `cae-foundrygate-prod` and changes the ingress FQDN. The comment promises the opposite of what will happen.

- **Minor** `.github/workflows/deploy-all.yml` lines 86-92 — The comment says "no Azure access", but with no `permissions:` block the job inherits `id-token: write` and `pull-requests: write` from the workflow. `permissions: {}` would make the sentence true.

- **Minor** `.github/workflows/ui-deploy.yml` lines 108-133 vs 259-282 — The OIDC login, the target check and the token fetch are byte-identical across `preview` and `close-preview`. The identity-boundary argument justifies not calling `_deploy-ui.yml`; it doesn't justify duplicating inside the file. The publish/`appsettings.json` block has already drifted from its `_deploy-ui.yml` original in this same commit — one copy has `set -uo pipefail`, the other doesn't.

- **Nit** `infra/parameters/prod.bicepparam` lines 53-55 — `sqlZoneRedundant = true` with `sqlBackupStorageRedundancy = 'Geo'` is valid for General Purpose (I checked; the ZR-requires-ZRS-backup rule is Hyperscale/Managed Instance only), but GRS is LRS in the primary region — the data survives a zone loss and the primary backups don't. `GeoZone` is already in the `@allowed` list; worth a line either way.

- **Nit** `.github/workflows/ui-deploy.yml` lines 56 and 62 — `src/FoundryGate.Core/**` builds a preview for nothing (`FoundryGate.Web.csproj` references Domain only, per the CONVENTIONS hard boundary), and `_deploy-ui.yml` is in the list even though the preview jobs deliberately never call it.

- **Nit** `src/FoundryGate.Api/Dockerfile` line 24 — "Renovate/Dependabot need the `custom.regex`/docker datasource pointed at mcr.microsoft.com — not wired up yet" is unfiled follow-up work sitting in a comment. Everything else here is scrupulous about #196/#109/#138/#105, which is what makes it stand out.

- **Nit** `infra/modules/control-plane.bicep` lines 73-81 — The five new params carry defaults that `main.bicep` unconditionally overrides and the leaf modules already declare, unlike the pass-throughs right beside them (`containerAppMinReplicas`, `sqlDatabaseSku`, `staticWebAppSku`) which deliberately have none. Three copies of `'0.25'` is two too many.

---

<sub>Reviewed at `0ab35c8`.</sub>

kolatts:
Thanks — this was a good catch list, and two of them were load-bearing in ways I had not seen. Every point addressed; details below, in your order. Head is `8c4ce9e` (which also merges `main`, 23 commits, no conflicts — more on that at the end).

---

### Major 1 — the approval count is seven, not five ✅ fixed, and the number is now six

You are right, and I had counted stages rather than gated jobs, which is exactly the mistake the table exists to prevent someone else making. Recounted off the job graph: `_deploy-infra.yml` gates only `deploy` (`validate` has no `environment:`), `_prepare-database.yml` one, `_deploy-database.yml` one, `_deploy-api.yml` **two**, `functions`+`ui` pending together, tests one → **seven**.

Took option (a) for the API and rejected it for functions/ui:

- **Collapsed `_deploy-api.yml`'s two gated jobs into one.** They were strictly sequential (`deploy` did `needs: build-push`, then re-checked out, re-logged in via OIDC and re-ran `export-outputs.sh` against the same deployment) so the second gate bought a reviewer nothing. The bootstrap flow is unaffected — it lives entirely in the second half, and the only cross-job value was `needs.build-push.outputs.image`, now `$GITHUB_ENV`. One approval, one checkout, one login and one `az deployment sub show` cheaper.
- **Did not change `functions`/`ui` to `needs: [database]`.** That would batch their approval with the API's and get to five, but it reintroduces precisely the day-0 race that ordering exists to prevent: on a bootstrap run the API stage re-runs the whole subscription deployment to swap the placeholder image, and an infra re-run restarting the Function App mid-upload is a real flake. Since GitHub gates at job-pending, ordering *is* approval batching — there is no way to keep the dependency and share the gate. One extra click is the cheaper side of that trade, and I would rather not trade a correctness property for it.

So the true count is **six**, and every copy now names the *jobs* rather than the stages so the next drift is visible: the `plan` script, `reference/ci-cd.md`, D-019 and `plans/22`. `deploy-all.yml`'s header and the `plan` job both carry a "if you add a job that declares `environment:`, these four are wrong until you update them together" note. D-019 records the correction rather than quietly overwriting the old number.

### Major 2 — "Static Web App Contributor" does not exist ✅ fixed with a custom role in Bicep

Confirmed both halves of your check against the live tenant:

```
az role definition list --name "Static Web App Contributor"                                    -> []
az role definition list --query "[?contains(to_string(permissions[].actions[]),'staticSites')]" -> []
```

No built-in grants any `Microsoft.Web/staticSites` action, so the fallback really would have been plain Contributor — Write and Delete on the site included, i.e. the outcome the design exists to avoid.

Added `infra/modules/swa-preview-role.bicep`: a subscription-scope `Microsoft.Authorization/roleDefinitions` named `FoundryGate SWA Preview Publisher ({env})`, deterministic GUID, deployed from `main.bicep` behind `deployControlPlane && deploySwaPreviewRole`. Four actions, each verified against `az provider operation show --namespace Microsoft.Web` (note the provider's own casing — `listsecrets` is lowercase in the manifest):

| Action | Why |
|---|---|
| `Microsoft.Web/staticSites/Read` | `az staticwebapp secrets list` resolves the site first |
| `Microsoft.Web/staticSites/listsecrets/action` | the deployment token |
| `Microsoft.Web/staticSites/builds/Read` | counting preview environments for the Free-tier check |
| `Microsoft.Web/staticSites/builds/Delete` | tearing a preview down on PR close |

Deliberately excluded and noted in the module: `staticSites/Write`, `staticSites/Delete`, `resetapikey/Action` (a compromised run could otherwise lock the real pipeline out), the `listappsettings` pair, and everything under `customdomains`/`privateEndpoint*`.

`assignableScopes` is the single Static Web App resource id, threaded through as a new `control-plane.bicep` output, so the role cannot be assigned anywhere else even deliberately. On your "fail loudly" point: Bicep has no assert, so an empty id becomes a deliberately invalid scope string and ARM rejects the deployment naming this module, rather than silently producing a subscription-assignable role.

Verified with a read-only `az deployment sub what-if` on `prod.bicepparam` — Succeeded, 60 creates (was 59), and the definition renders as:

```
roleName: FoundryGate SWA Preview Publisher (prod)
actions : [staticSites/Read, staticSites/listsecrets/action, staticSites/builds/Read, staticSites/builds/Delete]
scopes  : [/subscriptions/.../resourceGroups/rg-foundrygate-prod/providers/Microsoft.Web/staticSites/stapp-foundrygate-prod]
```

#109's owner commands now assign *that* role, with `deploySwaPreviewRole=false` documented as the escape hatch for a fork whose principal lacks `roleDefinitions/write`.

### Major 3 — the deployment token is app-scoped ✅ fixed; stated, not softened

You are right and the old header was the worst kind of wrong — a security control that overstates itself. Rewritten to separate the two boundaries explicitly:

- **Through ARM**, the identity holds the one custom role above and nothing else.
- **Through the token**, it is app-scoped, Azure exposes no slot-scoped SWA credential, so a compromised PR-track run could publish to the dev site's **production** slot — a registered MSAL redirect origin for the dev SPA, which makes the realistic worst case serving attacker-controlled JavaScript from an origin dev Entra trusts.

Accepted explicitly as a **dev-only** risk, with the reasoning written down (production SWA is a different resource this identity has no role on; the dev site holds no data; recovery is one `UI Deploy` dispatch) rather than left implied. Same text in `reference/ci-cd.md`.

Added `deployment_environment: pr-<n>` on both the upload and the close, and said plainly what it is: the intended staging target made explicit and auditable, *not* an enforcement the token respects.

### Major 4 — docs ladder ✅ fixed

- `DeployPipeline.astro`: added the UI-preview stage to `prConditional` and the ungated `plan` stage to `main`; the code comment and the `fgd-deploy-desc` screen-reader text no longer say "Nothing on the pull request track can change Azure" — they say the preview is the one PR-track thing that writes to Azure, and that all it can write is a staging slot on the dev site because its identity holds one role on one resource. The description also now mentions the six approvals.
- `architecture/overview.mdx`: **six** GitHub Environments, `ui-preview` named, and "nothing a pull request can run reaches it" replaced with the accurate framing — a PR runs its own copy of the workflow files, so the identity is the boundary, and here is how small each PR-reachable identity is.

### Minor — `.dockerignore` case sensitivity ✅ fixed, and it was worse than cosmetic

Reproduced: the image shipped `/app/appsettings.local.json`. Both casings are now listed for all four local-config names.

But chasing it turned up the part that matters: **the `/health` smoke test only passed *because* of the leak.** With the file correctly excluded the container exits — `Program.cs` startup validation requires `AzureAd__TenantId/ClientId/Audience/Instance` and `ConnectionStrings__FoundryGate`, and then `KeyProtectorFactory` requires `KeyProtection__Provider=DataProtection` outside Key Vault. So the smoke test was quietly asserting "the bug is present". Fixed by supplying all six from the environment, exactly as `control-plane.bicep` supplies them on Container Apps — the test now exercises the real configuration contract.

And because a dockerignore pattern that matches nothing fails *silently*, I put the check on the artefact rather than the pattern list: `build-api-image` now inspects the built image and fails on any `appsettings.{local,development,secrets}.json` / `local.settings.json`, before the smoke test. Shared by the PR check and the real build-and-push, so this cannot regress unnoticed.

### Minor — Free-tier staging limit ✅ fixed

Added a budget check between the token fetch and the publish: `az staticwebapp environment list`, count everything that is not buildId `default`, and if this PR has no slot yet and three are in use, emit a `::notice::` and skip green — naming both remedies (close a stale UI PR, or move dev to Standard). Needs `builds/Read`, which is why that action is in the custom role. Documented in `reference/ci-cd.md`.

### Minor — preview CORS ✅ added to #138, not fixed here

Confirmed: `control-plane.bicep` sets exactly one `Cors__AllowedOrigins__0`. Posted the analysis on #138 as the second half of that issue, with the asymmetry called out — the redirect-URI half needs Graph write (which `ui-preview` must *not* have, so it needs a different actor), while the CORS half is an API-side wildcard-subdomain match for dev only and is worth doing even before sign-in works. `ui-deploy.yml`'s header, the preview job summary and `reference/ci-cd.md` all now say a preview is honest for layout and non-API routes and nothing more, so nobody reads a green preview as a working app.

### Minor — "roughly doubles" ✅ fixed (three comments)

All three now say ~60% with the rates that justify it (`$0.152217/vCore-hr` + `$0.09133/vCore-hr` ZR surcharge, eastus2 retail 2026-09-02). You were right about the Container Apps row too: three replicas at 0.5 vCPU / 1.0 GiB is ~$118/mo, so the row is `~$39–118` and the production totals move to ~$660 (Basic v2) / ~$1,210 (Standard v2), with the arithmetic shown.

### Minor — `containerAppsZoneRedundant` "one-line change" ✅ fixed

Both comments now say the true thing: there is no `vnetConfiguration`/`infrastructureSubnetId` anywhere in `infra/`, the property is **immutable**, and #196 recreates `cae-foundrygate-prod` and changes the Container App's ingress FQDN. The parameter exists so that work inherits the wiring, not because the switch is cheap.

### Minor — `plan` job permissions ✅ fixed

`permissions: {}`, with a comment saying why the empty block is load-bearing.

### Minor — duplication across `preview` / `close-preview` ✅ fixed

Extracted to `.github/actions/swa-preview-token`: OIDC login, target check and token fetch, once. Both jobs call it. Its header also carries the scope-discipline rule, so the "no subscription-scope call" invariant now lives next to the only code that could break it. You were right that the identity-boundary argument justifies not calling `_deploy-ui.yml` and does not justify copy-paste inside the file — and the `set -uo pipefail` drift you spotted was the proof.

### Nit — ZR + `Geo` backups ✅ noted

Added the asymmetry to `prod.bicepparam`: RA-GRS is LRS *within* the primary region, so a zone loss takes the primary backups while the zone-redundant database keeps serving; `GeoZone` is the pairing that makes both halves zone-tolerant, and it is not the default only because it costs more and the last-resort restore path is the geo-secondary either way. Flagged to revisit with #105 rather than changed silently.

### Nit — dead `paths:` entries ✅ fixed

Dropped `src/FoundryGate.Core/**` (confirmed: `FoundryGate.Web.csproj` references Domain only) and `_deploy-ui.yml`; added `.github/actions/swa-preview-token/**`, which the preview now genuinely depends on.

### Nit — unfiled Renovate follow-up ✅ filed as #212

Fair hit — it was the one place I let follow-up work sit in a comment. #212 covers pointing the docker datasource at MCR (or a regex manager keyed on the tag comments that are already above each `FROM` for exactly this reason) and the requirement that it resolve **manifest-list** digests, or CI pins itself to amd64. Dockerfile references it.

### Nit — redundant defaults in `control-plane.bicep` ✅ fixed

All five defaults dropped, plus a note on the module explaining why nothing there carries one: `main.bicep` declares every default and `@allowed` list once and passes all of them unconditionally, so a default here is a third copy — and the copy that silently wins when someone adds a parameter to `main.bicep` and forgets to thread it. Without one, that is a build error.

---

## Verification

| Gate | Result |
|---|---|
| `actionlint 1.7.7` + `shellcheck 0.10.0` on PATH | clean, 18 workflows (integration confirmed live against a probe file that still reports SC2129) |
| `shellcheck -x` on all five `.github/scripts/infra/*.sh` | clean |
| all four composite actions parse (as `actionlint.yml` does) | ok |
| `resolve-api-image.test.sh` | 11 passed, 0 failed |
| `az bicep build` / `lint` — `main.bicep` + 16 modules | clean (3 pre-existing `BCP081` on `Microsoft.CognitiveServices@2026-07-01`) |
| `az bicep build-params` test / dev / prod | clean |
| read-only `az deployment sub what-if`, `prod.bicepparam` | Succeeded, 60 creates; role definition renders four actions scoped to the single Static Web App |
| `docker build` | succeeds |
| image content check | `/app` contains `appsettings.json` only — no `appsettings.local.json` |
| `/health` smoke, config from env only | 200, `uid=1654(app)` |
| `npm run build` in `docs-site/` | 15 pages |
| PR checks on `8c4ce9e` | all green |

## One thing worth flagging

The `build-test` failure on my first push of this fix pass was **not** mine: `GroupsPageTests.Searching_re_queries_from_page_one` is a bUnit `WaitForAssertion` timeout in code that arrived on `main` via #176 — a file my branch did not contain at all. I have since merged `main` (23 commits, no conflicts, both this PR's Bicep params and the merged control-plane changes intact), re-run every gate above against the merged tree, and confirmed all 10 `GroupsPageTests` pass locally. CI is green on the merge commit. Flagging it because a `WaitForAssertion` that fails under runner load and passes locally will do it again to someone else.

Not merging — over to you.

PR #200, opened 2026-09-02
docs-site: light theme palette + toggle, reflow KeyLifecycle for mobile

### LINKED ISSUE #114: docs-site: light palette + theme toggle (site is dark-only; diagrams are already token-driven)
Found while verifying #89's diagrams "in light and dark": the docs site is dark-only by design — `docs-site/src/components/ThemeProvider.astro` forces `data-theme="dark"`, `ThemeSelect.astro` renders nothing, and `src/styles/tokens.css` defines a single (dark) `--fg-*` palette. There is no light rendering to verify or support today.

If a light theme is ever wanted:
- Add light values for the `--fg-*` tokens under `[data-theme='light']` (bg scale, borders, text scale; azure/ember can stay).
- Re-enable Starlight's ThemeSelect/ThemeProvider (drop the overrides) and remove the `html { color-scheme: dark }` force in `starlight-theme.css`.
- Verify the landing page (`src/pages/index.astro`, nebula + `mix-blend-mode: screen`) and the aurora orbs in `PageFrame.astro`, which assume a dark ground.
- The architecture diagrams (`SystemDiagram`, `DeployPipeline`, `QuotaChain`, `KeyLifecycle` + `src/styles/diagrams.css`) consume only `--fg-*` tokens and `color-mix()` on them, so they need only a visual pass, not a rewrite.

Not blocking anything; filing so the "light theme" question has a home.

### LINKED ISSUE #115: docs-site: reflow the KeyLifecycle state machine for narrow screens instead of horizontal scroll
`docs-site/src/components/KeyLifecycle.astro` (#89) is an inline SVG with a 640 px minimum width. On a 390 px viewport it stays legible because it scrolls inside its own `.fgd-svg-scroll` container (the page itself never scrolls horizontally, and a "scroll sideways" hint is shown), but a reflowed vertical layout would be nicer on phones.

Options:
- A second, mobile-only SVG variant (states stacked vertically, loops drawn to the side) toggled by a media query — simple but duplicates content.
- Rebuild the state machine in HTML/CSS on the shared `.fgd-*` primitives (`src/styles/diagrams.css`) so it reflows like `SystemDiagram`/`DeployPipeline`; return edges (Suspended → Active) are the hard part.

Keep `role="img"` + `aria-label` and the token-only colours either way.

kolatts:
## Summary

- **#114 — light palette + theme toggle.** `tokens.css` now defines the light
  palette as the `:root` default, with dark values under `[data-theme="dark"]`
  and a `prefers-color-scheme: dark` guard (`:root:not([data-theme="light"])`),
  matching Starlight's own toggle mechanism. Re-enabled Starlight's stock
  `ThemeSelect`/`ThemeProvider` instead of the custom overrides that forced
  `data-theme="dark"` and rendered no picker; removed the `html { color-scheme:
  dark }` force. The four diagram components (`SystemDiagram`, `DeployPipeline`,
  `QuotaChain`, `KeyLifecycle`) are already token-driven per CONVENTIONS.md and
  needed no rewrite — only `tokens.css` plus one component-level fix (see
  below). The landing page (`index.astro`) predates the Starlight layout and
  keeps its own copy of the palette; it now reads the same
  `starlight-theme` `localStorage` key (try/catch-wrapped) so a theme choice
  made on either side of the site carries over.
- **#115 — reflow KeyLifecycle for narrow screens.** `KeyLifecycle.astro` gets
  a second, vertically-stacked SVG variant (states top-to-bottom, the two
  self-transitions collapsed into one "stays Active" annotation box, the
  re-activation edge drawn as a tall curve along the left margin), swapped in
  at `<=640px` via `diagrams.css` in place of the horizontal-scroll container
  and its "scroll sideways" hint. `role="group"` and the screen-reader-only
  long-form description are shared by both variants — no duplicate
  accessible text.

### Contrast work (#114)

`--fg-azure-neon` / `--fg-ember-soft` / `--fg-ember-hot` / `--fg-success` /
`--fg-info` are used as text (headings, links, status badges, SVG diagram
labels) throughout the site. Their dark-tuned bright values read at roughly
1.7–2.6:1 against a light background, so light mode gets darker, WCAG
AA-safe (>=4.5:1) variants of the same hues. `--fg-azure`, `--fg-azure-dim`,
`--fg-azure-hover` (as a button background) and `--fg-ember` stay constant
across themes — their existing contrast already holds. One place pairs a
theme-varying color with a color that intentionally stays constant:
`QuotaChain`'s numbered badge circle (`--fg-azure-dim` background, always a
dark chip) now uses a fixed light text color instead of the now-theme-aware
`--fg-azure-neon`.

### Other light-mode fixes found while verifying #114

- The nebula background art and aurora orbs (`PageFrame.astro`,
  `index.astro`) were designed against a dark ground; new
  `--fg-ambient-nebula-opacity` / `--fg-ambient-orb-opacity` tokens dial them
  back in light mode instead of a redraw.
- `logo-cropped.png` bakes in a silver-gradient wordmark tuned for a
  near-black background — "foundry" was nearly invisible on white. A new
  `--fg-logo-filter` token darkens it in light mode as a stopgap; a proper
  light-mode logo asset is a design task, not a CSS one, so it's not attempted
  here.
- `icons.png`'s dark-stroke line art was unconditionally inverted to white
  (for a dark ground); a new `--fg-icon-invert` token skips the invert in
  light mode.
- Fixed a **pre-existing** mobile horizontal-scroll bug found during
  verification: `.dashboard-wrap`'s glow pseudo-elements bleed ~40px past
  their box by design; added `overflow-x: hidden` on `html, body` to clip it
  at the viewport edge instead of letting it force horizontal scroll on
  narrow screens. Unrelated to theming — the bug predates this PR — but it's
  a one-line fix directly serving this PR's own "no horizontal scroll" gate.

## Verification

- `cd docs-site && npm ci && npm run build` — clean, on this branch rebased
  onto latest `origin/main`.
- Visual pass with the `playwright-cli` skill at 1280px and 390px, in both
  themes, on `/`, `getting-started/why-foundrygate`,
  `architecture/overview`, `architecture/feasibility`, `reference/api`:
  - Legible text in both themes; body text contrast checked against the
    token values (light body text ≈8.5:1 on `--fg-bg-base`).
  - No horizontal page scroll — confirmed via `document.documentElement.
    scrollWidth` vs `window.innerWidth` on all five pages at 390px (all
    390–391px, i.e. no overflow).
  - All four diagram components (`SystemDiagram`, `QuotaChain`,
    `DeployPipeline`, `KeyLifecycle`) read cleanly in light mode at both
    widths.
  - `KeyLifecycle`'s mobile variant renders as a clean vertical stack at
    390px with no sideways scroll, in both themes; the desktop
    horizontal-scroll variant is unchanged above 640px.
  - Starlight's theme `<select>` switches themes, persists to
    `localStorage["starlight-theme"]` across reload, and the landing page
    picks up the same stored preference.
- Screenshots kept in the local scratchpad, not committed.

## Deferred

None — both issues are closed by this PR. The logo/icon CSS filters above are
called out as stopgaps in code comments in case a proper light-mode asset is
wanted later, but nothing is left as an inline TODO; if a light logo variant
is wanted, it belongs in its own design-focused issue.

Closes #114
Closes #115


kolatts:
**APPROVE**

Clean pass. Light palette actually hits the AA numbers it claims, both KeyLifecycle SVGs render correctly, the deleted theme overrides left nothing dangling, and the toggle + landing-page sync genuinely work when driven through the real `<select>` — not just when the tokens are poked directly. Two minor things below worth a look, neither blocks.

---

**Verification**

- `npm ci && npm run build` on `d36455b` — clean, 15 pages, Pagefind index built.
- Screenshotted `/`, `getting-started/why-foundrygate`, `architecture/overview`, `architecture/feasibility`, `reference/api` at 1280/390 in both themes via the actual Starlight `<select>` (not just `dataset.theme`, which desyncs the picker's own label from what it renders — that's a test artifact, not a bug, confirmed by re-testing through `select` + reload).
- Contrast, computed from the literal `tokens.css` hex values (WCAG relative luminance), not eyeballed:
  - `--fg-text-secondary` (body) vs `--fg-bg-base`: **9.49:1** — comfortably past 4.5:1.
  - `--fg-azure-hover` (actual link color via `.sl-markdown-content a`) vs base: **6.01:1**.
  - `--fg-azure-neon` (hero "Automatically." accent, `QuotaChain` badge text) vs base: **5.67:1**.
  - `--fg-ember-soft` / `--fg-ember-hot` / `--fg-success` / `--fg-info`: 4.79–5.12:1, all clear.
  - `--fg-azure` itself sits at **4.33:1** against `--fg-bg-base` (below 4.5) — but it's only ever used as a `background`/`border-color`, never `color`, so it's judged against the 3:1 UI-component bar, which it clears. Worth knowing if anyone reaches for it as text color later.
- No horizontal scroll at 390px on any of the five pages, both themes (`scrollWidth`/`innerWidth` within 1px, matches the PR's own note).
- All four diagrams (`SystemDiagram`, `QuotaChain`, `KeyLifecycle`, `DeployPipeline`) legible in light mode at 1280px; `KeyLifecycle`'s vertical mobile variant renders cleanly at 390px in both themes.
- Both `KeyLifecycle` SVGs are `aria-hidden="true"` unconditionally (not toggled with the CSS visibility swap) — so neither is ever in the a11y tree; the accessible description lives entirely in the shared `<figure role="group" aria-describedby>` text. That's the right call and matches the existing desktop SVG's pattern — nothing to fix.
- No dangling references to the deleted `ThemeSelect.astro`/`ThemeProvider.astro` anywhere (config comment is the only mention).
- Diff is CSS/Astro-component only — no prose/content `.md`/`.mdx` files touched, no CLAUDE.md docs-invariant violations.

---

**Issues**

- **Minor** `docs-site/src/pages/index.astro` line 133–140 (`html, body { height: 100%; ... overflow-x: hidden; }`) — this turns `<body>` into its own internal scroll container instead of the viewport scrolling natively. Confirmed empirically: after `window.scrollTo`/wheel-scroll, `window.scrollY` stays `0` and `document.documentElement.scrollHeight` clamps to the viewport height, while `document.body.scrollTop` is where the real position lives. It's the CSS overflow spec's "if one axis is non-visible, the other computes to `auto`" rule firing on both `html` and `body` at once. Nothing on the page relies on `window.scrollY`/`scrollTo` today and the sticky nav still sticks (verified), so it's invisible right now — but it'll silently break the first `window.scrollTo` back-to-top button, in-page `#anchor` CTA, or browser scroll-to-text/find-in-page restoration someone adds later. Scoping the clip to a wrapper div (or just `body`, not `html`) instead of the document root sidesteps it entirely.
- **Minor** `docs-site/src/pages/index.astro` lines 111–223 — the landing page's token block now duplicates `tokens.css`'s light values, dark values, *and* the `prefers-color-scheme` media query almost line-for-line (~90 lines). That duplication already existed pre-PR for the dark palette; this PR roughly doubles it. Astro allows a plain page to `import '../styles/tokens.css'` the same way Starlight's `customCss` does — worth a follow-up before a third palette tweak drifts the two copies out of sync (this PR didn't drift them, to be clear — just flagging the growing maintenance surface).

---

kolatts:
Thanks — both Minors fixed in 56fb31b.

**1. `overflow-x: hidden` on `html, body`** — moved off both entirely. Added a `.page-clip` wrapper scoped to `<main>` only (that's where the actual overflow originates — `.dashboard-wrap`'s glow pseudo-elements bleeding ~40px past their box), leaving `<nav>` and `<footer>` outside it. `<nav>` is `position: sticky`, and putting it inside *any* element with non-default overflow breaks its stickiness (confirmed empirically — nav stopped pinning at the top when the wrapper included it). Re-verified: `window.scrollY` updates correctly on scroll, `document.body.scrollTop` stays `0` (html is the scroll container again, not body), sticky nav still pins after scrolling 800px, and `scrollWidth === innerWidth` at 390px on `/` in both themes.

**2. Duplicated token block** — split `tokens.css` into `palette.css` (just the `--fg-*` variables: light default, `[data-theme='dark']`, and the `prefers-color-scheme` media query) and a slimmed-down `tokens.css` (the `@fontsource` `@import`s + `@import './palette.css'`, still what Starlight's `customCss` consumes). `index.astro` now does `@import '../styles/palette.css'` instead of copying ~90 lines.

Worth flagging: my first attempt imported `tokens.css` directly (fonts included), which built clean but broke at runtime — pulling the `@fontsource` `@import`s into a second Vite CSS entry point mangled the asset-URL rewriting for the font files (`url("./files/inter-...woff2")` instead of the hashed `_astro/inter-....woff2` path Vite actually emits), 404s across the board. Confirmed via console — 10 failed font requests on `/`. Splitting the palette out from the font imports fixed it: `grep`ed `dist/_astro/*.css` for `files/inter`/`files/monaspace` afterward, zero hits, zero console errors on `/` or on a docs page.

Local verification: `npm run build` clean; playwright pass on `/` at 1280px/390px in both themes (screenshots match pre-fix renders pixel-for-pixel where nothing changed); `architecture/overview` sanity-checked for console errors too. Pushed to `claude/docs-site-theme-114-115`; CI now green (`build-test`, `Build Astro site`, `docs-build` all pass).

Issue #197, opened 2026-09-02
HealthEndpointTests readiness test passes only when no SQL Server is reachable locally


kolatts:
`Api/Endpoints/HealthEndpointTests.HealthReady_returns_503_rather_than_throwing_when_the_database_is_unreachable` fails on `main` (562a759) on a machine where the local docker SQL container is running:

```
Expected: ServiceUnavailable
Actual:   OK
```

The test overrides the connection string to a closed port through `builder.ConfigureAppConfiguration(...)`. CONVENTIONS.md's integration-test section says exactly why that does not work: with minimal hosting, `ConfigureAppConfiguration`'s sources are appended only *after* `Program.cs` has already read configuration, so the override does not reach what the host actually bound — factory configuration has to go through `IWebHostBuilder.UseSetting`.

So the 503 the test asserts is not produced by the override at all; it is produced by whatever real connection string `appsettings.local.json` carries being unreachable. On a developer machine (or a CI runner) that happens to have SQL listening at that address, the readiness check reports Healthy and the test fails.

**Fix:** switch the override to `builder.UseSetting("ConnectionStrings:FoundryGate", "...")` and re-verify with a local SQL container running — the assertion is a good one, it just is not currently testing what it claims.

Found while running the Predeployment suite for the #182/#193/#147/#159/#170/#108 sweep; unrelated to those changes (reproduced on `main` with no local modifications).

kolatts:
Claiming this for the final sweep C branch (claude/final-sweep-c).

PR #202, opened 2026-09-02
feat(cli): implement foundrygate db compare (DacFx schema-compare, Windows-only)

### LINKED ISSUE #103: Implement foundrygate db compare (DacFx schema-compare, Windows-only) to regenerate dbo/Tables/*.sql from local docker SQL

kolatts:
## Summary

Implements `foundrygate db compare [--connection-string <cs>] [--apply] [--check]`, deferred from #100 into #103 (plans/23-database-tooling.md, original design).

- **What it does**: DacFx `SchemaComparison` between a live SQL Server database (default: the local docker instance `foundrygate local setup` keeps current via `EnsureCreated` against the EF model) and `FoundryGate.Database`'s checked-in `dbo/Tables/*.sql`, scoped to table objects only (`SchemaCompareOptions` excludes every other DacFx object type and ignores pure column reordering/whitespace/comments).
- With no flags, or `--check`: prints the differences and exits non-zero if any exist (usable as a local pre-commit gate). No differences → no files are ever touched.
- `--apply`: additionally regenerates the affected table files via DacFx's own `PublishChangesToProject` (the same engine SqlPackage/SSDT's schema-compare tooling uses, which is why the checked-in files already look like its output), then exits 0 on success.
- Windows-only: fails fast on macOS/Linux with a message pointing at `SchemaParityTests`, the cross-platform CI backstop this command complements and never replaces.

**Two real DacFx bugs found and fixed by the live proof** (not visible from reading the API docs):

1. `SchemaCompareProjectEndpoint`'s project reader doesn't understand `FoundryGate.Database.sqlproj`'s SDK-style implicit `dbo/**/*.sql` globbing (`Sdk="Microsoft.Build.Sql"`) — passed an empty script list it silently loads a **zero-object target model**, so every table compares as "Add" even with no real drift, with no error raised. Fixed by enumerating the project's `.sql` files explicitly and reading the project's `<DSP>` MSBuild property instead of hardcoding a schema name in its place.
2. When `PublishChangesToProject` rewrites a changed `CREATE TABLE` batch, DacFx declares the primary key inline and leaves this repo's original separate `ALTER TABLE ... ADD CONSTRAINT [PK_x] PRIMARY KEY` batch untouched below it — two declarations of the same constraint, which still parses (and still passes `SchemaParityTests`' first-match regex) but **fails at deploy time**. `SqlTableFileNormalizer` strips the redundant batch from every file `--apply` touches.

## Design decisions for the reviewer

- **No hand-rolled SQL renderer.** `--apply` regenerates files through DacFx's own `PublishChangesToProject` (matching imagile-app's `SchemaComparisonHelpers` pattern) rather than a from-scratch text renderer — the checked-in files already mimic that tool's output style, and reusing the real engine is both more faithful and far less code to get subtly wrong than reverse-engineering DacFx's column-alignment algorithm byte-for-byte. `SqlTableFileNormalizer` is the one piece of hand-written text-rewriting logic, and it exists only to fix the one real defect DacFx's own output has (see bug 2 above).
- **No `LibGit2Sharp`.** The original plan's ordering-noise-discard machinery (a post-hoc git-diff comparison) is unnecessary: `SchemaCompareOptions.Create()` sets DacFx's own `IgnoreColumnOrder`/`IgnoreWhitespace`/`IgnoreComments`, so pure reordering is never reported as a difference in the first place, and `--apply` never touches a file unless there's a real difference to publish.
- **Testability split** mirrors `DeployCommand`/`DacpacDeployOptions`/`GrantIdentitiesRunner`: `CompareRunner` (decision/formatting logic) is tested against a fake `ISchemaComparer`; `SchemaCompareOptions` (object-type scoping) and `SqlTableFileNormalizer` (the real text-rewriting logic) are tested directly with fixtures; `DacFxSchemaComparer` (the actual DacFx calls) has nothing to fake around a live SQL connection and `.sqlproj`, so it's covered by this PR's live proof instead.
- `--check` and bare `db compare` (no flags) behave identically (report-only, non-zero exit on drift) — `--check` exists as a discoverable, explicit flag for scripting callers.

## Verification

- `dotnet build FoundryGate.sln -c Release` — 0 warnings, 0 errors.
- `dotnet test src/FoundryGate.Tests.Predeployment -c Release` — **1268/1268 passing**.
  - Note: `HealthEndpointTests.HealthReady_returns_503_...` flakes when a SQL Server is reachable locally on the machine — pre-existing, unrelated to this PR (reproduces identically with this branch's changes stashed), already tracked in #197. The full suite is green with no local SQL Server reachable (i.e. after `docker compose down`, which is the state used for the numbers above).
- **Live proof** (docker SQL Server, `docker compose up -d`, `foundrygate local setup`):
  - (a) No drift on `main`'s schema: `db compare --check` → `Schema is up to date...`, exit 0, `git status` shows zero file changes.
  - (b) Deliberate drift (`User.DisplayName` `[StringLength(200)]` → `[StringLength(250)]`, then `local setup` to recreate the DB from the new model):
    - `db compare --check` → `Found 1 table difference(s): Change Table [dbo].[Users]`, exit 1.
    - `db compare --apply` → regenerates `Users.sql`, exit 0.
    - `dotnet test --filter SchemaParityTests` → passes against the regenerated file.
    - `dotnet build FoundryGate.sln` → dacpac builds clean (no duplicate-PK error).
    - `foundrygate db deploy <dacpac> <fresh-db-connection-string>` → deploys successfully; verified via `sqlcmd` that the deployed `Users.DisplayName` column is `NVARCHAR(250)` and there's exactly one `PK_Users` constraint.
    - Reverted the entity change and regenerated file; re-ran `local setup` + `db compare --check` → clean again, exit 0.
  - `docker compose down` — torn down after verification.

## Command surface

```
foundrygate db compare [--connection-string <cs>] [--apply] [--check]
```

- `--connection-string`: defaults to the local docker SQL Server (`Server=127.0.0.1,3433;...`, matching `local setup`).
- `--apply`: regenerate `FoundryGate.Database/dbo/Tables/*.sql` to match the source database.
- `--check`: print differences and exit non-zero if any exist, without changing files (same behavior as no flags).

## Deferred / discovered work

- The `HealthEndpointTests` local-SQL-Server-presence flake mentioned above is out of this PR's scope (Api is another agent's territory per the concurrency split) and is already tracked as #197.

Closes #103


kolatts:
**REQUEST CHANGES**

Good bones — the Windows-only guard, the `ISchemaComparer`/`CompareRunner` split, and catching two real DacFx quirks via live proof instead of trusting the docs are all the right instincts. But the fix for the "DacFx duplicates constraints on regen" bug only covers primary keys, not foreign keys, and I reproduced a case where that duplication fails the dacpac **build** outright — not a hypothetical, verified end-to-end.

---

**Issues**

- **Major** `src/FoundryGate.Cli/Commands/Db/Compare/SqlTableFileNormalizer.cs` — `RemoveDuplicatePrimaryKeyAlterStatement` only strips a duplicated **primary key**, but DacFx's `PublishChangesToProject` inlines foreign keys into the rewritten `CREATE TABLE` batch too, leaving the original `ALTER TABLE ... ADD CONSTRAINT [FK_x]` batches untouched below it. I reproduced this live: added a column to `GroupMember` (composite PK, 3 FKs), ran `local setup` + `db compare --apply`, and the regenerated `GroupMembers.sql` declared all three FKs twice. `dotnet build src/FoundryGate.Database` fails with `SQL71508: The model already has an element that has the same name dbo.FK_GroupMembers_*` for each one — this isn't a deploy-time-only failure like the PK bug, it doesn't even build. And `SchemaParityTests` passes it silently, same blind spot as the PK case before this PR's fix (the FK regex requires literal `ADD CONSTRAINT`, which only matches the surviving `ALTER TABLE` copy). The PR's own live proof only exercised `Users.sql` — no FKs, single-column PK — so this path never got exercised. Every table in the schema except `Users` has at least one FK; as written, `--apply` is landmined for most of the actual schema.
- **Minor** `src/FoundryGate.Cli/Commands/Db/Compare/SqlTableFileNormalizer.cs` — the `BalancedParens` regex and the PK-matching pattern are copy-pasted near-verbatim from `SchemaParityTests.cs` (same `.NET` balancing-group construct, same "how do I match a nested-paren constraint clause" problem solved twice). Worth sharing if the two need to keep parsing the same SQL shape in sync.
- **Minor** `src/FoundryGate.Cli/Commands/Db/Compare/SqlTableFileNormalizer.cs` — if the inline PK is found but `AlterTablePrimaryKeyBatchRegex` doesn't match the leftover `ALTER TABLE` batch (e.g. real DacFx whitespace drifts from the hand-crafted test fixtures), `Regex.Replace` quietly no-ops — no exception, no signal to the caller that the fixup didn't happen. Combined with the Major finding above, there's currently no safety net between "the normalizer silently didn't do its job" and a broken file landing in the repo.

---

**Everything else checked out**

- Build: `dotnet build FoundryGate.sln -c Release` — clean, 0 warnings/errors.
- Tests: `dotnet test src/FoundryGate.Tests.Predeployment -c Release --no-build` — 1267/1268; the one failure (`DashboardPageTests.Stops_refreshing_once_the_page_is_gone`) is an unrelated pre-existing timing flake in the Web project, confirmed by re-running it alone (12/12 green) — not touched by this diff.
- `dotnet format --verify-no-changes` — clean.
- Live proof against docker SQL Server: `db compare --check` on `main`'s schema (no drift) → exit 0, zero file changes. Drift detection also confirmed correct (`Found 1 table difference(s)` → exit 1) before hitting the FK bug above.
- `RepoLocator` correctly resolves the repo root and `.sqlproj` path from an arbitrary cwd (ran the built CLI from `/tmp`, well outside the repo) — matches the doc comment's claim.
- Connection-string default (`Server=127.0.0.1,3433;...`) matches `docker-compose.yml`'s port mapping exactly.
- Error handling is better than I expected: a live connection failure (docker down) surfaces as a clean `db compare failed: Schema comparison failed: ...` one-liner via `CliErrors`, not a raw `SqlException` stack trace — `DacFxSchemaComparer.Compare()` correctly funnels DacFx's error list through `InvalidOperationException`, which `CliErrors.IsExpected` already covers.
- `--apply` scope is correctly locked to `dbo/Tables/*.sql` via `SchemaCompareOptions.Create()` excluding every non-`Tables` `ObjectType` — verified the exclusion list and the `IgnoreColumnOrder`/`IgnoreWhitespace`/`IgnoreComments` settings.
- `ArgumentNullException`/`ArgumentException.ThrowIfNull*` conventions, primary constructors, and the `ISchemaComparer`/`CompareRunner` fake-testability split all match established precedent elsewhere in `FoundryGate.Cli` (`GrantIdentitiesRunner` et al.) — no CONVENTIONS.md deviations found.
- Docs changes (`infrastructure.md`, `overview.mdx`, `plans/23-database-tooling.md`) accurately describe what the command does.

The FK-duplication bug is the blocker — everything downstream of `--apply` for `Users.sql` was genuinely verified, but the majority of this schema (anything with a foreign key) will build-fail the moment DacFx has to rewrite a `CREATE TABLE` batch for it. Given `SchemaParityTests` won't catch it either, I'd want either the normalizer generalized to strip duplicate FK batches the same way it does PK batches, or `--apply` to detect and refuse to leave a file in a state that fails a local `dotnet build src/FoundryGate.Database` before declaring success.

kolatts:
One more pass turned up a Nit, for completeness (doesn't change the decision above):

- **Nit** `src/FoundryGate.Cli/Commands/Db/Compare/CompareCommand.cs` — the `--connection-string` option's `Description` interpolates `SetupCommand.LocalConnectionString` in full, so `foundrygate db compare --help` prints the literal dev SA password (`Temp1234!`) to the console. Not a new secret — it's the same docker-compose-only local password `SetupCommand` already echoes at runtime today — but worth a redacted description if you're touching this file again.

(Also chased down whether a raw `SqlException` could leak past `CliErrors.IsExpected` on a connection failure, since it's not in the allowed-exception list — confirmed live with docker stopped that it doesn't: `DacFxSchemaComparer.Compare()` funnels DacFx's own error list through `InvalidOperationException`, which is already covered. No action needed there.)

kolatts:
Fix pass for the review — point by point.

**Major** — `SqlTableFileNormalizer` only de-duplicated the primary key.

Fixed. `RemoveDuplicateInlineConstraints` (replacing `RemoveDuplicatePrimaryKeyAlterStatement`) now handles every constraint kind DacFx can inline: PK, FK, `UNIQUE`, `CHECK`, and column-level `DEFAULT`. I went with the inversion you suggested rather than generalizing the original "strip the ALTER TABLE batch" direction: this repo's separate `ALTER TABLE ... ADD CONSTRAINT` batches are now treated as canonical, and any DacFx-inlined duplicate is stripped out of the `CREATE TABLE` body instead. That's both simpler (one pass, no per-kind "reconstruct the ALTER TABLE regex" logic) and self-reinforcing — the file lands back in exactly the same shape every other hand-authored table already uses, rather than a hybrid of this repo's style and DacFx's.

Reproduced your exact repro (`GroupMembers`, composite PK + 3 FKs) as a fixture — `SqlTableFileNormalizerTests.Foreign_keys_Removes_every_inline_duplicate_reproducing_the_real_GroupMembers_failure` asserts the output is byte-for-byte the real checked-in `GroupMembers.sql`. Also added fixtures for `UNIQUE`, `CHECK`, and `DEFAULT` (all synthetic — none of those exist in this schema today, but the reviewed generalization needed to cover them), plus a case where a constraint has no matching `ALTER TABLE` batch at all (genuinely new, not a duplicate) and is correctly left untouched.

Live proof re-run drifted three tables at once specifically to exercise this: `GroupMember` gained a nullable `Note` column (forces every inline constraint — PK + 3 FKs — to get duplicated, your exact scenario), `QuotaIncreaseRequest.Justification` widened 2000→2500 (PK + 3 FKs), `Group.Description` widened 1000→1500 (single PK, no FKs, exercises the filtered index survives untouched). After `--apply`: `dotnet build FoundryGate.sln` (which builds `src/FoundryGate.Database`) succeeded — the exact `SQL71508` failure you hit is gone — `SchemaParityTests` passed, and I went one step further than "builds": `db deploy` of the rebuilt dacpac to a fresh database succeeded, and I confirmed via `sys.foreign_keys`/`sys.key_constraints` there are exactly 9 constraints across the three tables (1 PK + 3 FK each for `GroupMembers`/`QuotaIncreaseRequests`, 1 PK for `Groups`) — no duplicates, and the `Groups.IX_Groups_EntraGroupId` filtered index survived with its `WHERE ([EntraGroupId]<>'')` intact.

**Fail loudly instead of silently no-op'ing** — done, and it earned its keep during development. `RemoveDuplicateInlineConstraints` has a post-condition check: after removal, it re-scans for any constraint name that still has both an `ALTER TABLE` batch *and* an inline occurrence, and throws (naming the file) if it finds one, rather than shipping a file that looks fixed but isn't. While generalizing the removal logic to handle multiple constraints per table, I hit a real bug this exact check caught: two adjacent removed constraints were both trying to claim the same separating comma (one via "consume the following comma," the other via "consume the preceding comma"), which silently dropped one constraint's removal — the safety net threw instead of shipping the corrupted file, which is exactly how it's supposed to work. Fixed by switching to a single consistent rule (always consume the *preceding* comma, which is provably non-overlapping for any subset of targeted items — see the method's doc comment for why). There's also a guard against a `CREATE TABLE` body match running past its real closing paren and swallowing an `ALTER TABLE` batch into what it thinks is the column list (checks the captured body doesn't itself contain `ALTER TABLE`/`CREATE TABLE`); a unit test forces that path with a deliberately malformed fixture and asserts the throw names the file.

**Minor — share `BalancedParens`.** Done: it's now `FoundryGate.Cli.Commands.Db.Compare.SqlPatterns.BalancedParens`, a public constant, and `SchemaParityTests` references it directly (`private const string BalancedParens = SqlPatterns.BalancedParens;`) via the `ProjectReference` the test project already has on `FoundryGate.Cli` — same source, not a synced copy.

**Minor — redact the SA password from `--help`.** Done: `--connection-string`'s `Description` no longer interpolates `SetupCommand.LocalConnectionString`; it now says "the local docker SQL Server from docker-compose.yml, port 3433" and points at the file instead.

**Gates**: `dotnet build FoundryGate.sln -c Release` — 0 warnings/errors. `dotnet test src/FoundryGate.Tests.Predeployment -c Release` — 1274/1274 (docker down; the `HealthEndpointTests` flake you flagged only reproduces with a local SQL Server reachable, tracked separately in #197, not touched by this diff). `dotnet format --verify-no-changes` — clean.

**Live proof (full re-run)**: docker up → `local setup` → `db compare --check` on clean `main` (0 diffs, exit 0) → three-table deliberate drift (`GroupMember.Note`, `QuotaIncreaseRequest.Justification`, `Group.Description`) → `local setup` again → `db compare --check` (3 diffs, exit 1) → `db compare --apply` (regenerates all 3 files, no normalizer throw) → `dotnet build FoundryGate.sln` succeeds → `SchemaParityTests` passes → `db deploy` to a fresh database succeeds, constraint counts and the filtered index verified via `sqlcmd` → reverted all 6 files (3 entities + 3 regenerated `.sql`) → `local setup` + `db compare --check` clean again (exit 0) → `docker compose down`.

Pushed to `claude/cli-db-compare-103` (191af24). Not merged.

Issue #203, opened 2026-09-02
Flaky: MePageTests.Rotating_asks_for_confirmation_before_it_touches_the_api fails intermittently in a full-suite run


kolatts:
Sibling of #195, which the `KeyDisplay` render fix (395e8a0) closed for `The_rotated_mask_survives_the_next_render`. A different test in the same class still fails intermittently:

```
FoundryGate.Tests.Predeployment.Web.MePageTests.Rotating_asks_for_confirmation_before_it_touches_the_api [FAIL]
Failed!  - Failed: 1, Passed: 1285, Total: 1286
```

Reproduced once on `main` + the #182/#193/#147/#159/#170/#108 branch during a full-suite run; passes on an immediate re-run of the whole suite, and passes 24/24 when the class is run on its own (`--filter FullyQualifiedName~MePageTests`). So it is timing/ordering sensitive, not a real regression — same signature as #195 before it was diagnosed.

Worth looking at with the #195 fix in hand: that one turned out to be a render the assertion did not wait for, and a confirmation-dialog assertion is the same shape of race (assert on the dialog before bUnit has flushed the render that produces it). Likely wants the same `WaitForAssertion`/explicit-render treatment rather than a retry.

kolatts:
Another sighting (from PR #198's first CI run): `GroupsPageTests.Searching_re_queries_from_page_one` — a bUnit `WaitForAssertion` timeout under runner load; passes locally 10/10. Add to the list of tests to harden alongside `Rotating_asks_for_confirmation…`. — orchestrator

kolatts:
Claiming this for the final sweep C branch (claude/final-sweep-c).

kolatts:
Also seen once during PR #211's fix pass: `Web/DashboardPageTests.Stops_refreshing_once_the_page_is_gone` (timer/disposal family) — green on re-run. Folded here rather than a new issue; the sweep hardening #203 should cover it. — orchestrator

kolatts:
**A third sibling in the same family: `DashboardPageTests.Stops_refreshing_once_the_page_is_gone`.**

Seen on PR #216 (`claude/final-sweep-d`, which touches neither `FoundryGate.Web` nor `Tests.Predeployment/Web` — `git diff --stat origin/main...HEAD -- src/FoundryGate.Web src/FoundryGate.Tests.Predeployment/Web` is empty):

```
[...]

Issue #201, opened 2026-09-02
infra should emit Gateway__Tiers__* so a fork that overrides quotaTiers cannot drift from the control plane


kolatts:
Split out of #108, which verified the rest of the `Gateway__*` binding.

`infra/modules/control-plane.bicep` sets every other `Gateway__*` value on both hosts (`SubscriptionId`, `ResourceGroup`, `ApimName`, `ApimGatewayUrl`, `LogAnalyticsWorkspaceId`, `LogAnalyticsWorkspaceResourceId`, `KeyEncryptionKeyUri`, `FoundryAccountNames__{i}`). It does **not** set the tier table. Both hosts get `Gateway:Tiers` from their own shipped `appsettings.json`, and three predeployment tests keep those in step with each other and with `infra/main.bicep`'s `quotaTiers` **default**:

- `GatewayOptionsTiersTests.Shipped_appsettings_tiers_match_infra_main_bicep_quotaTiers_ids_and_caps_in_order`
- `FunctionsAppSettingsTests.The_shipped_Functions_tier_table_is_identical_to_the_Apis`
- `GatewayInfraBindingTests.Infra_does_not_emit_the_tier_table` (new, pins this issue's premise)

So the shipped configuration is consistent. The gap is a fork that **overrides `quotaTiers` at deploy time**: APIM then creates products and `llm-token-limit` caps the control plane has never heard of, while the control plane keeps validating quotas against the appsettings values. D-013 says a budget *is* a tier — so this is the exact "SQL says one thing, the gateway enforces another" class the `isGatewayCapped` flag exists to make visible, except here it is a deployment-time misconfiguration nothing detects.

**Ask:** have `control-plane.bicep` project `quotaTiers` into `sharedAppConfig` as
`Gateway__Tiers__{i}__ProductId` / `__DisplayName` / `__MonthlyTokenQuota`, so the deployed tier table always comes from the same parameter that created the products.

**Two things to get right when doing it:**

1. **The binder appends list items; it does not replace them.** `GatewayOptions.Tiers` deliberately has no C# defaults for exactly this reason (see the class remarks). With the env vars set *and* the shipped `appsettings.json` still carrying three tiers, every tier would be duplicated, and `GatewayOptions.Validate` would fail startup on "lists a product id more than once" — which is at least loud, but the fix is to empty `Gateway:Tiers` in both hosts' `appsettings.json` (or make it local-only) in the same PR.
[...]

Issue #199, opened 2026-09-02
PUT /config on DefaultMonthlyTokenQuota is synchronous: a large fork turns one edit into thousands of ARM calls


kolatts:
Follow-up from #193, which is now implemented synchronously.

`ConfigService.UpdateAsync` re-resolves every active default-tier developer inside the request when `DefaultMonthlyTokenQuota` changes, and each one whose tier actually moves is an `IGatewayTierSync` call — an APIM subscription re-scope. That is exactly right for correctness (the database never claims a default the gateway is not enforcing) and fine for the fork sizes we ship for: a few hundred developers, mostly already on the tier the default maps to, so the tier-change count is usually far smaller than the re-resolved count.

It does not scale to a tenant with thousands of default-tier developers. One `PUT /config` becomes thousands of sequential ARM calls on an HTTP request thread:

- the request will exceed any sane gateway/ingress timeout long before it finishes;
- a failure partway leaves the whole unit of work rolled back, so the admin has to retry the entire sweep;
- there is no progress signal, and the client that gave up has no way to learn whether it landed.

**Options, roughly in order of appetite:**

1. **Queue it.** Write the config row, enqueue a "re-resolve default-tier users" message, and let the Functions host drain it — which needs #180 (give the Functions host the ability to move tiers) first, or the queued work has the same blind spot the monthly reset has today.
2. **Bound it.** Re-resolve synchronously up to a threshold and switch to queued above it, so small forks keep the immediate, transactional behaviour and large ones degrade to eventually-consistent with an audit row saying which mode ran.
3. **Batch the gateway calls.** Most of the cost is per-subscription ARM round trips; if APIM ever exposes a bulk re-scope, or if #118 lets us skip developers already on the target product without inferring from history, the synchronous path may simply stop being a problem.

Worth deciding alongside #180 rather than on its own — the two share the "who is allowed to move a tier, and from where" question.

kolatts:
Naming the monthly reset here too, per the #211 review.

Since #194 the scheduled reset can move APIM subscriptions, and #211's fix pass makes each move its own committed unit of work — so a `DefaultMonthlyTokenQuota` change now produces **N serialized ARM round trips plus N saves** driven by the timer, on top of the Api-side fan-out this issue already covers. Same shape, same fix (batch or throttle the moves), just a second entry point.

Two things that are already true and bound it a little:

[...]

PR #204, opened 2026-09-02
feat: config concurrency + default-quota re-resolution, quota-request integrity, gateway config verification

### LINKED ISSUE #108: Bind the Gateway__* configuration section set by infra (APIM/Foundry/workspace addressing) in Api and Functions

### LINKED ISSUE #147: Guard the one-pending-quota-request-per-period rule with a filtered unique index

### LINKED ISSUE #159: Stale quota increase requests: approval ignores the request's own period and nothing expires them
`QuotaIncreaseRequest` records the billing period it was filed for (`PeriodYear`/`PeriodMonth`), but
nothing in the review path or anywhere else ever looks at it again:

- `QuotaRequestService.ApproveAsync` applies the approved tier and re-resolves
  `BillingPeriod.Current(timeProvider)` — today's period, not the request's. Approving a request filed in
  July in September raises the user's budget for September, while the response and the row still report
  `periodYear: 2026, periodMonth: 7`. The number in the UI and the period actually affected disagree.
- Nothing expires a pending request. A request nobody reviewed stays `Pending` and approvable forever,
  and it keeps blocking new submissions only for its own period — so a developer can accumulate one
  stale pending request per month, all of them still individually approvable.

The approval-time downgrade guard added in #148 limits the damage (a stale request whose tier the user
has since passed is now a `409`), but a stale request that is still technically an increase applies a
months-old justification to today's budget with no signal that it is old.

**Options:**

1. **Expire on read/review.** Treat a pending request whose period is older than the current one as
   stale: refuse approval with `409` naming the period it was filed for, and surface it in
   `GET /requests` (a computed `isStale`, or a status). Cheapest, and keeps the audit trail honest.
2. **Sweep.** Have the monthly reset (`POST /quota/reset`, and the Functions job in #38) close pending
   requests from earlier periods as `Rejected` with a system note, the way
   `CancelPendingForUserAsync` closes a departing user's. Gives admins an empty queue each month.
3. **Scope approval to the request's period.** Least useful — raising a budget for a month that has
   already been billed and reset does nothing for the developer asking.

Recommend (1) plus (2): refuse stale approvals immediately, and have the monthly reset clear them so the
review queue never accumulates.

Found during review of #148.

### LINKED ISSUE #170: PUT /config/{key} is last-write-wins: no concurrency token on SystemConfiguration
Raised in review of #166 (Minor, explicitly not blocking).

`ConfigService.UpdateAsync` reads the `SystemConfiguration` table, mutates the matched row and saves. `SystemConfiguration` carries no concurrency token, so two admins editing the same key from two open `/config` pages silently produce last-write-wins: no `DbUpdateConcurrencyException`, no signal to the loser, and the audit trail shows both writes as if they were sequential and intentional.

**Deferred rather than fixed in #166, deliberately.** The obvious cheap fix is an optional `ExpectedUpdatedDate` on `UpdateSystemConfigRequest` — 409 when it does not match the stored `UpdatedDate`. Adding it now would mean shipping a public request field with no caller: the admin config page (#55) does not exist yet, so nothing would send it, nothing would exercise the 409 outside a synthetic test, and the shape would be frozen before the UI that has to use it is designed. The field is optional and therefore additive, so deferring costs nothing in compatibility — which is precisely the argument for waiting until #55 is being built and can be designed against a real "Save all changes" flow (that flow PUTs several dirty rows in sequence, and the interesting question is what it does with a partial failure, not what a single PUT returns).

**When #55 lands, do this:**

1. Add `DateTimeOffset? ExpectedUpdatedDate` to `UpdateSystemConfigRequest`.
2. In `UpdateAsync`, 409 when it is supplied and differs from the row's `UpdatedDate`, with a message naming the admin who got there first (`UpdatedByUser.DisplayName`) — `GET /config` already returns `updatedByDisplayName`, so the UI can render "Ada Lovelace changed this while you were editing" without another round trip.
3. Decide the multi-row save semantics in the UI: stop at the first conflict and show a diff, rather than pushing the remaining rows over someone else's work.

A real EF concurrency token (`[Timestamp]`/`rowversion`) is the alternative, but it is a schema change on a reference-data table whose columns are all `[DoNotUpdate]`, and it would make the seeder's job more delicate for a nine-row table. The request-field approach keeps the guarantee where the contention actually is: between two humans with a form open.

### LINKED ISSUE #182: Finish the CommitToken sweep: GroupService and EntraGroupSyncService still hand-roll the ternary

### LINKED ISSUE #193: PUT /config on DefaultMonthlyTokenQuota must re-resolve the users it affects
Found in review of #187 (Functions wave).

`PUT /config` accepts a new `DefaultMonthlyTokenQuota` and writes the row — and re-resolves nobody. Quota resolution reads that key as level 5 of the precedence chain, so every user who falls through to it has a stale `QuotaAllocation` from the moment the key changes until something else touches them.

## Why that is worse than it sounds

The first thing to notice the change is usually the **scheduled monthly reset**, which runs in the Functions host. That host has no APIM management client, so it re-resolves the tier in SQL and cannot move the developer's APIM subscription to the new tier product. The result:

- `QuotaAllocation.AllocatedTokens` / `TierProductId` say Power (20M)
- the gateway's `token-quota` policy still enforces Standard (5M) and returns `403` at 5M
- the dashboard and the gateway disagree, for every default-tier developer at once

#187 makes this **visible** — `WarningGatewayTierSync` logs a Warning per affected user and the reset's audit row carries `tierChangeCount` — but visible is not fixed.

## Ask

`ConfigService` (or whatever owns `PUT /config`) should, when `DefaultMonthlyTokenQuota` changes:

1. Find the users whose resolution actually depends on it — no `MonthlyTokenQuota`, not `IsUnlimited`, and no group membership that would win at levels 3–4. (`IQuotaResolutionService.PreviewAsync` answers "would this user's budget change?" without writing, if a cheap bulk form is worth adding.)
2. Call `IQuotaResolutionService.ResolveManyAsync` for the current period, which fires `ApimGatewayTierSync` for each user whose tier moved — this is the Api, so the real sync is registered.
3. Observe the commit point: resolution may have re-scoped subscriptions at the gateway, so the audit row and the save go on `CancellationToken.None` (CONVENTIONS.md; the pattern is in `QuotaResetService`).
4. Audit it as one row with the affected count — an admin changing one config value and moving 200 developers between products deserves a trail entry that says so.

Worth deciding at the same time: whether a large fork wants this synchronous on the request, or queued. A tenant with thousands of default-tier developers would make one `PUT /config` into thousands of ARM calls.

## Related

- #180 — the other half: give the Functions host the ability to move tiers itself, so a reset is not a silent divergence.
- The `IsGatewayCapped` flag already covers the *shape* of this problem for legacy numeric quotas; this is the same class of "SQL says one thing, gateway enforces another" and should end the same way.

### LINKED ISSUE #195: Flaky: MePageTests.The_rotated_mask_survives_the_next_render fails ~2 in 5 full-suite runs

kolatts:
Six issues across config, quota requests and the gateway configuration contract. One commit per issue; each is independently reviewable.

## #182 — `CommitToken.For` instead of hand-rolled commit tokens

`GroupService` carried a private `CommitToken(bool, CancellationToken)` that *shadowed* the shared `FoundryGate.Data.Concurrency.CommitToken` type inside the very class that needed it, and `EntraGroupSyncService` still spelled the rule out as an inline ternary. Both call `CommitToken.For(...)` now. Behaviour identical — `For` is the same conditional.

## #170 — optional `ExpectedUpdatedDate` on `PUT /config/{key}`

`UpdateSystemConfigRequest` gains an optional `ExpectedUpdatedDate`. Supplied and mismatched → `409` naming the value the row holds now, when it changed, and (when it has one) the editor's display name, so a config page can say "Ada Lovelace changed this while you were editing" without another round trip. That name is a second query on the conflict path only.

The check runs **before** the per-key value validation: a caller whose view of the row is stale has to re-read it whatever they were writing, so a stale write with a bad value is the 409, not the 400. Comparison is by instant, so a UTC-normalizing client (and the SQLite harness, which always reads back `+00:00`) still matches. Omitting the field keeps last-write-wins, so the field is additive.

## #147 — filtered unique index behind one-pending-request-per-period

The `AnyAsync` pre-check had nothing behind it, so two concurrent submissions could both insert and show the same person twice in the reviewer queue. `QuotaIncreaseRequest` now declares `IX_QuotaIncreaseRequests_PendingPerUserPeriod` on `(UserId, PeriodYear, PeriodMonth)`, filtered to `[StatusType] = 0` in the co-located configuration — the half an attribute cannot express, and the half that keeps a *decided* request from locking a developer out for the rest of the month. Mirrored in `dbo/Tables/QuotaIncreaseRequests.sql`; `SchemaParityTests` already understood filtered indexes and confirms the two agree. The insert translates the violation into the same `ConflictException` the serial path gives, matching on the identifier each provider names (index name on SQL Server, columns on SQLite) — the `GroupService.SaveGroupAsync` precedent.

`DashboardServiceTests` seeded two pending rows for one user and period — a state the API never produced; it uses two requesters now.

## #159 — stale quota increase requests

Both halves of the issue's recommendation:

1. **Approval refuses a closed period** (`409` naming both periods). Approval re-resolves *today's* period, so approving a July request in September raised September's budget while the row still said July. It is the first refusal in `ApproveAsync`, before any read that can reach the gateway. Rejection stays unrestricted so an admin can always clear a queue by hand.
2. **The monthly reset sweeps the queue.** The rule is `FoundryGate.Core/Requests/QuotaRequestExpiry` — Core, because `IQuotaResetService` lives there and cannot reference the Api, and because a timer and a button must not disagree about what expiry means. Pending requests from an earlier period become `Rejected` with a system note and no reviewer (the `CancelPendingForUserAsync` shape; the enum has no "Expired" state and the null reviewer is what tells the two apart), plus one `quota.requests-expired` audit row via `IAuditWriter.AddSystem` carrying the count and ids. Nothing saves in Core: the rows join the reset's unit of work.

`IQuotaRequestService.ExpireStaleAsync(ct)` gives the Api the same sweep over its own save. No HTTP caller yet, by design.

## #193 — `PUT /config` on `DefaultMonthlyTokenQuota` re-resolves the users it affects

That key is level 5 of the precedence chain, so it changed and nobody moved: the allocation row said Power (20M) while the gateway still enforced Standard (5M), for every default-tier developer at once. `ConfigService.UpdateAsync` now finds every **active** user the default governs (no user override, no group winning at levels 3–4) union anyone whose current allocation already reads `SystemDefault`, re-resolves them via `ResolveManyAsync` in the same unit of work, and finishes on `CommitToken.For(any TierSyncRequested, ct)`. The audit row carries `reresolvedUserCount` and `tierChangeCount`.

**One prerequisite in Core**, and the crux of the change: `QuotaResolutionService.GetSystemDefaultAsync` read the key with `AsNoTracking()` and a projection, which cannot see a pending change — it would have faithfully written the OLD default back onto everyone. It checks the change tracker first now, the same "pending state first, database second" rule `FindExistingAsync` and `LoadGroupPoliciesAsync` already follow.

The affected set is a query, not a `PreviewAsync` per user: the chain is deterministic, so the level is a predicate, and a preview each would be one round trip per developer on the one path that may already have hundreds.

## #108 — `Gateway__*` binding verified, `configuration.mdx` made the truth table

**Verified.** The bicep's shared block sets eight `Gateway__*` values on the Container App *and* the Function App: `SubscriptionId`, `ResourceGroup`, `ApimName`, `ApimGatewayUrl`, `LogAnalyticsWorkspaceId`, `LogAnalyticsWorkspaceResourceId`, `KeyEncryptionKeyUri`, `FoundryAccountNames__{i}`. Every one binds to a `GatewayOptions` member today; no member is unfed except `Tiers`.

**`Tiers` is the one gap, and appsettings covers the shipped configuration.** Infra does not emit `Gateway__Tiers__*`; both hosts ship the table in their own `appsettings.json`, kept in step with each other and with `infra/main.bicep`'s `quotaTiers` **default** by existing tests. That holds until a fork *overrides* `quotaTiers` at deploy time, at which point APIM enforces caps the control plane has never heard of — filed #201.

**New drift alarm**, since "we checked once" is not a guarantee: `GatewayInfraBindingTests` reads `control-plane.bicep` as text (the `SchemaParityTests` precedent) and fails if a name drifts, if infra sets a variable nothing binds, if a new `Gateway` member arrives with no declared source, if infra starts emitting the tier table behind the appsettings guidance, or if the two host modules stop receiving the same shared block.

**Docs.** `reference/configuration.mdx` is one per-host truth table (shared / gateway addressing with a "read by" column / API-only / Functions-only) instead of an API table plus a Functions paragraph saying "the same, plus". `reference/infrastructure.md` keeps its table as the provenance view and points at configuration.mdx for meaning.

No `infra/` changes — that area belongs to another agent this wave, and nothing needed changing there for the binding itself.

## Verification

- `dotnet test FoundryGate.Tests.Predeployment` — **1286/1286 passing** (rebased on `origin/main` @ 4389c84). One full-suite run hit `MePageTests.Rotating_asks_for_confirmation_before_it_touches_the_api`, which passes on re-run and 24/24 in isolation — a sibling of the just-fixed #195 flake, filed as #203.
- `npm run build` in `docs-site/` — 15 pages, complete

## Discovered and filed

- **#197** — `HealthEndpointTests` readiness test passes only when no SQL Server is reachable locally. Reproduced on `main` with no local changes: it overrides the connection string via `ConfigureAppConfiguration`, which CONVENTIONS.md explicitly says does not reach bound configuration under minimal hosting, so the 503 comes from the real connection string being unreachable rather than from the override.
- **#199** — `PUT /config` on `DefaultMonthlyTokenQuota` is synchronous; a fork with thousands of default-tier developers turns one edit into thousands of ARM calls on a request thread. Worth deciding alongside #180.
- **#201** — infra should emit `Gateway__Tiers__*` (see #108 above). Note in the issue: the binder *appends* list items, so the shipped appsettings tiers must be emptied in the same change or every tier duplicates.
- **#203** — the `MePageTests` sibling flake.

## For the reviewer

- **#147, index placement.** The columns and uniqueness are an `[Index]` attribute; only `HasFilter` is in the configuration. That splits one index across two places, where the `Groups.EntraGroupId` precedent put the whole thing in the configuration — but it is what CONVENTIONS.md asks for ("annotations first; configuration only for what annotations can't express"), and the attribute carries a comment telling the reader to read it together with the filter, because a `IsUnique = true` that is not actually globally unique is the misreading worth preventing. Happy to move it wholesale to the configuration if you prefer the Group shape.
- **#159, no `Expired` status.** Expired requests are `Rejected` with a null `ReviewedByUserId` and a system note, reusing the `CancelPendingForUserAsync` shape. A fourth enum value would be more honest but every read path, filter and UI would have to learn it; the null reviewer already distinguishes them.
- **#193, the union.** The affected set includes users whose *current allocation* says `SystemDefault` even if the chain has since moved them off it. That means a config edit can correct one stale row belonging to a user the default no longer governs. It is a correction, it costs one row, and the audit count records it — but it does widen the blast radius of a config edit slightly.
- **#108, the reflection test.** `Every_GatewayOptions_member_is_either_set_by_infra_or_listed_as_deliberately_absent` will fail if someone adds a `GatewayOptions` property without deciding where it comes from — including the `ModelAliases` work in flight on another branch. That is the alarm working as intended (a new `Gateway` member with no source is exactly #108's bug), and the failure message says what to do, but it is a deliberate cross-branch tripwire rather than a silent addition.
- `RecordingCommandInterceptor` gains a `BeforeExecuting` hook. Where `FailWhen` breaks a statement, this changes the world at a chosen moment — the only way to open a read-then-write window from a test. Used once, by #147's race test.

Closes #182
Closes #193
Closes #147
Closes #159
Closes #170
Closes #108

### DIFF
--- a/docs-site/src/content/docs/architecture/overview.mdx
+++ b/docs-site/src/content/docs/architecture/overview.mdx
@@ -225,6 +225,10 @@ changing that setting takes effect the next day rather than at the next deployme
       detail: "Runs on or after SystemConfiguration[ResetDayOfMonth] (1–28) when this month has no reset in the audit log yet, so a tick lost to an outage is picked up by the next day rather than costing the month its reset. One replica then takes a short blob lease, renewed while held, so two instances waking together cannot both run — though the reset is idempotent either way.",
     },
     {
+      title: "Close quota requests the period left behind",
+      detail: "Any increase request still Pending from a month that has ended becomes Rejected, with a system note and no reviewer. Those requests can no longer be approved anyway — approval refuses a closed period — so leaving them would just mean the review queue starts every month with last month's dead entries in it. First, before anything can touch the gateway, so a failure here costs nothing.",
+    },
+    {
       title: "Resolve quota for the new period and upsert QuotaAllocation",
       detail: "For each active developer, run the five-level quota resolution for the new month and insert their row. A row that already exists is re-resolved but keeps its reconciled TokensUsed — zeroing it mid-period would only make the dashboard disagree with the gateway.",
     },
@@ -233,8 +237,8 @@ changing that setting takes effect the next day rather than at the next deployme
       detail: "Gateway token-quota counters reset themselves at the UTC month boundary — no APIM counter manipulation needed, and no subscription state is touched. The Function only clears the IsHardStopped mirror so dashboards start the month clean.",
     },
     {
-      title: "Write one audit log entry",
-      detail: "A single quota.monthly-reset row, committed in the same transaction as every allocation it describes. The same code runs behind an admin's manual POST /quota/reset, which audits as quota.reset and names the admin.",
+      title: "Write the audit log entries",
+      detail: "One quota.monthly-reset row carrying the counts, committed in the same transaction as every allocation it describes — plus a second quota.requests-expired row when the sweep above actually closed something (a sweep that finds nothing writes no row at all). The same code runs behind an admin's manual POST /quota/reset, which audits as quota.reset and names the admin.",
     },
   ]}
 />
--- a/docs-site/src/content/docs/reference/api.md
+++ b/docs-site/src/content/docs/reference/api.md
@@ -291,8 +291,12 @@ appears on their first `GET /quota/allocations/me` of the month, as it always di
 
 **One open request per user per period.** A second submission while one is still `Pending` in the
 same UTC calendar month is `409`; a decided (approved or rejected) request frees the slot at once, so
-a rejected developer can re-ask with a better justification. The check is not yet backed by a database
-constraint, so two simultaneous submissions can both land ([#147](https://github.com/kolatts/foundry-gate/issues/147)).
+a rejected developer can re-ask with a better justification. The rule is backed by a **filtered unique
+index** on `(userId, periodYear, periodMonth) WHERE status = Pending`
+([#147](https://github.com/kolatts/foundry-gate/issues/147)), so two simultaneous submissions — a
+double-clicked button, a retrying client — get the same `409` as two sequential ones instead of both
+landing and showing the same person twice in the reviewer queue. The filter is what keeps a *decided*
+request from blocking the rest of the month.
 
 Every submission writes `quota.requested`; approval writes `quota.approved` with the before/after
 quota, the quota and level resolution reported at review time, the tier product and whether the gateway
--- a/docs-site/src/content/docs/reference/api.md
+++ b/docs-site/src/content/docs/reference/api.md
@@ -250,8 +250,10 @@ exist are re-resolved (`allocatedTokens`, level, tier, capped flag) but **keep t
 `tokensUsed`** — the gateway's monthly window resets itself, so zeroing the mirror mid-month
 would only make dashboards lie. Every touched row gets `isHardStopped = false` and
 `resetDate = now`. Exactly one audit row (`quota.reset`, attributed to the calling admin, details
-`{ usersResetCount, periodYear, periodMonth, createdCount, tierSyncCount, expiredRequestCount }`)
-per run. Returns `{ usersResetCount, periodYear, periodMonth, resetDate }`. Running it twice in a
+`{ usersResetCount, periodYear, periodMonth, tierChangeCount, expiredRequestCount }`)
+per run. `tierChangeCount` is named for what it means to a human reading the trail — every one of those
+is a developer whose *enforced* budget moved this run — not for the seam it went through.
+Returns `{ usersResetCount, periodYear, periodMonth, resetDate, expiredRequestCount }`. Running it twice in a
 month produces the same row count.
 
 **The reset also sweeps the review queue.** In the same transaction, every quota increase request
@@ -273,6 +275,7 @@ nothing writes no row at all. The scheduled monthly job runs exactly the same co
 | `GET` | `/requests/{id}` | Owner or Admin | Request detail. `404` for anyone else — see below |
 | `POST` | `/requests/{id}/approve` | Admin | Approve. Applies the tier to the user and re-resolves the current period immediately |
 | `POST` | `/requests/{id}/reject` | Admin | Reject. Nothing about the user's quota changes |
+| `POST` | `/requests/expire-stale` | Admin | Close every request left `Pending` from a period that has ended. No body. Returns `{ expiredCount }` |
 
 **A request asks for a tier, not a number.** `requestedQuota` must be `null` (unlimited) or exactly
 one of the configured tier caps from `GET /quota/tiers` — anything else is `400` listing the allowed
@@ -446,10 +449,17 @@ transaction**, which moves their APIM subscription to the new tier product
 ([#193](https://github.com/kolatts/foundry-gate/issues/193)). Before this, the key changed and nobody
 moved: the allocation row said one thing, the gateway enforced another, and the first thing to notice
 was usually the scheduled monthly reset — which runs in the Functions host, has no APIM management
-client, and so could only log the divergence it was creating. A failed gateway move fails the whole
-edit, so the database never claims a default the gateway is not enforcing. Re-saving the key's existing
-value re-resolves nobody, and deactivated users are skipped (their key is revoked; there is nothing to
-move).
+client, and so could only log the divergence it was creating. Re-saving the key's existing value
+re-resolves nobody, and deactivated users are skipped (their key is revoked; there is nothing to move).
+
+A failed gateway move fails the whole edit, so **the database never claims a default the gateway is not
+enforcing**. The converse is not guaranteed, and on a large fork it is the likelier failure: the
+subscriptions are re-scoped one at a time, so if developer 200 of 400 throws, 199 have already been
+moved at the gateway and nothing commits — the gateway then enforces a default SQL never recorded. The
+next reset or config edit converges it, and the direction is the safe one (the record is behind the
+enforcement, not ahead of it), but it is a real partial state. It is the same shape the monthly reset
+has, and [#199](https://github.com/kolatts/foundry-gate/issues/199) — which is about not doing thousands
+of ARM calls inside one request at all — is where it gets solved.
 
 Every accepted edit stamps `updatedByUserId` (the calling admin) and `updatedDate`, and writes one
 `config.updated` audit row with `{ key, before, after, reresolvedUserCount, tierChangeCount }` — in the
--- a/docs-site/src/content/docs/reference/api.md
+++ b/docs-site/src/content/docs/reference/api.md
@@ -376,13 +376,13 @@ Rules the mutation paths enforce (CLAUDE.md "Anthropic deployments are create-on
 | Method | Path | Auth | Description |
 |---|---|---|---|
 | `GET` | `/config` | Admin | Every `SystemConfiguration` key, ordered by key: `{ key, value, updatedDate, updatedByUserId, updatedByDisplayName }`. The last two are `null` for a seeded key no admin has edited |
-| `PUT` | `/config/{key}` | Admin | Set one key's value. Returns the row as it now stands |
+| `PUT` | `/config/{key}` | Admin | Set one key's value. Returns the row as it now stands. Optional `expectedUpdatedDate` makes the write conditional (`409` if someone else got there first) |
 | `GET` | `/audit` | Admin | Audit log, paged. Filter: `?actor=&action=&from=&to=` |
 | `GET` | `/dashboard` | Admin | Summary stats for the dashboard. `?fresh=true` bypasses the 30-second cache |
 
 ### `PUT /config/{key}`
 
-Body: `{ "value": "..." }`. The key is matched case-insensitively; an unknown one is `404`. The
+Body: `{ "value": "...", "expectedUpdatedDate": "..." }` (the second field optional). The key is matched case-insensitively; an unknown one is `404`. The
 value is validated **per key** before it is stored — a configuration editor that accepts a value
 the system cannot use only moves the failure somewhere harder to find — and stored **normalized**
 (trimmed, booleans lower-cased, URLs and resource ids without a trailing slash), so two admins
@@ -406,6 +406,16 @@ Three keys that earlier versions seeded — `ApimGatewayUrl`, `ApimProductId`, `
 answered `409 read-only`; now the rows are gone entirely (the next `db seed-reference` deletes them),
 `GET /config` does not list them, and `PUT` on one is the ordinary `404`. What replaced each is in the
 [Configuration Reference](/foundry-gate/reference/configuration/#retired-keys).
+
+**Optimistic concurrency is opt-in per request.** Send back the `updatedDate` you read from
+`GET /config` as `expectedUpdatedDate` and the write is refused with `409` if anyone has changed the
+row since — the `detail` names the value it holds now, when it changed and, when the row has an
+editor, who they were, so a config page can say "Ada Lovelace changed this while you were editing"
+without another round trip. Omit the field and the write is last-write-wins, exactly as before
+([#170](https://github.com/kolatts/foundry-gate/issues/170)). The check runs *before* the per-key
+value rule, so a stale write with a bad value is the `409`, not the `400`: a caller whose view of the
+row is out of date has to re-read it whatever they were trying to write. The timestamp is compared as
+an instant, so a client that normalizes to UTC still matches.
 
 Every accepted edit stamps `updatedByUserId` (the calling admin) and `updatedDate`, and writes one
 `config.updated` audit row with `{ key, before, after }` — in the same transaction as the change, so
--- a/docs-site/src/content/docs/reference/api.md
+++ b/docs-site/src/content/docs/reference/api.md
@@ -250,9 +250,18 @@ exist are re-resolved (`allocatedTokens`, level, tier, capped flag) but **keep t
 `tokensUsed`** — the gateway's monthly window resets itself, so zeroing the mirror mid-month
 would only make dashboards lie. Every touched row gets `isHardStopped = false` and
 `resetDate = now`. Exactly one audit row (`quota.reset`, attributed to the calling admin, details
-`{ usersResetCount, periodYear, periodMonth, createdCount, tierSyncCount }`) per run. Returns
-`{ usersResetCount, periodYear, periodMonth, resetDate }`. Running it twice in a month produces
-the same row count.
+`{ usersResetCount, periodYear, periodMonth, createdCount, tierSyncCount, expiredRequestCount }`)
+per run. Returns `{ usersResetCount, periodYear, periodMonth, resetDate }`. Running it twice in a
+month produces the same row count.
+
+**The reset also sweeps the review queue.** In the same transaction, every quota increase request
+still `Pending` from a *closed* period is set to `Rejected` with a system note and no reviewer — the
+same shape a departing user's requests get ([#159](https://github.com/kolatts/foundry-gate/issues/159)).
+Those requests can no longer be approved anyway (see below), so leaving them Pending would only mean
+an admin opens the queue each month with last month's dead entries still in it. When the sweep closes
+anything it writes one additional `quota.requests-expired` audit row (no actor, no target, details
+`{ expiredCount, currentPeriodYear, currentPeriodMonth, expiredRequestIds }`); a sweep that finds
+nothing writes no row at all. The scheduled monthly job runs exactly the same code.
 
 ## Quota Increase Requests
 
@@ -297,6 +306,14 @@ index** on `(userId, periodYear, periodMonth) WHERE status = Pending`
 double-clicked button, a retrying client — get the same `409` as two sequential ones instead of both
 landing and showing the same person twice in the reviewer queue. The filter is what keeps a *decided*
 request from blocking the rest of the month.
+
+**A request only applies to the period it was filed for.** Approval re-resolves the *current* period,
+so approving a July request in September would raise September's budget while the row and the response
+still said `periodYear: 2026, periodMonth: 7` — the number in the UI and the month actually affected
+disagreeing. Approving a request whose period has closed is therefore `409` naming both periods
+([#159](https://github.com/kolatts/foundry-gate/issues/159)). **Rejection is always allowed**, so an
+admin can clear an old queue by hand; and the monthly reset closes stale requests itself, so in
+practice this `409` only covers the window between a period ending and the next reset running.
 
 Every submission writes `quota.requested`; approval writes `quota.approved` with the before/after
 quota, the quota and level resolution reported at review time, the tier product and whether the gateway
--- a/docs-site/src/content/docs/reference/api.md
+++ b/docs-site/src/content/docs/reference/api.md
@@ -438,9 +438,23 @@ value rule, so a stale write with a bad value is the `409`, not the `400`: a cal
 row is out of date has to re-read it whatever they were trying to write. The timestamp is compared as
 an instant, so a client that normalizes to UTC still matches.
 
+**Changing `DefaultMonthlyTokenQuota` re-resolves the developers it governs.** That key is level 5 of
+the quota precedence chain, so the moment it changes, every **active** user who falls through to it —
+no `isUnlimited`, no user quota, no group that would win at levels 3–4 — plus anyone whose current
+allocation already reads `SystemDefault`, is re-resolved for the current period **in the same
+transaction**, which moves their APIM subscription to the new tier product
+([#193](https://github.com/kolatts/foundry-gate/issues/193)). Before this, the key changed and nobody
+moved: the allocation row said one thing, the gateway enforced another, and the first thing to notice
+was usually the scheduled monthly reset — which runs in the Functions host, has no APIM management
+client, and so could only log the divergence it was creating. A failed gateway move fails the whole
+edit, so the database never claims a default the gateway is not enforcing. Re-saving the key's existing
+value re-resolves nobody, and deactivated users are skipped (their key is revoked; there is nothing to
+move).
+
 Every accepted edit stamps `updatedByUserId` (the calling admin) and `updatedDate`, and writes one
-`config.updated` audit row with `{ key, before, after }` — in the same transaction as the change, so
-a value can never move without a trail. The calling admin must already have a FoundryGate user row
+`config.updated` audit row with `{ key, before, after, reresolvedUserCount, tierChangeCount }` — in the
+same transaction as the change, so a value can never move without a trail, and an admin who changed one
+value and moved 200 developers between products has an entry that says so. The calling admin must already have a FoundryGate user row
 (`GET /users/me` provisions one) or the write is `403`; reading `/config` needs no such row.
 
 **A deploy never reverts an edit.** `SystemConfiguration`'s value, timestamp and editor columns are
--- a/docs-site/src/content/docs/reference/configuration.mdx
+++ b/docs-site/src/content/docs/reference/configuration.mdx
@@ -65,73 +65,90 @@ optional — the Api refuses to start without at least one tier.
 Validated at startup (fail-fast): at least one tier, product ids ⊆ `GatewayTiers.All`, no
 duplicates, every cap in range, exactly one unlimited tier, and at least one finite tier.
 
-**Adding or changing a tier** means changing two things together: `quotaTiers` in
+**Adding or changing a tier** means changing three things together: `quotaTiers` in
 `infra/main.bicep` (creates the APIM product and renders its `llm-token-limit` policy) and
-`Gateway:Tiers` here (what the Api validates quotas against and serves from `GET /quota/tiers`).
-A new product id must also be added to `GatewayTiers.All` in Domain. Until #108 has infra emit
-`Gateway__Tiers__*`, a fork that overrides `quotaTiers` at deploy time updates this section by
-hand. Existing quotas that no longer match any tier are not rejected on read — they are enforced
-at the next tier up and flagged `isGatewayCapped` until an admin corrects them.
+`Gateway:Tiers` in **both** hosts' `appsettings.json` (what the Api validates quotas against and
+serves from `GET /quota/tiers`, and what quota resolution in the Functions host maps onto). A new
+product id must also be added to `GatewayTiers.All` in Domain. Three predeployment tests keep the
+four in step, so getting this wrong fails the build rather than a deployment.
+
+`Gateway__Tiers__*` is the one part of the `Gateway` section infra does **not** set — every other
+value is an environment variable on both hosts (see below). A fork that overrides the `quotaTiers`
+parameter at deploy time therefore has to update both `appsettings.json` files by hand, or the gateway
+will enforce caps the control plane has never heard of;
+[#201](https://github.com/kolatts/foundry-gate/issues/201) tracks having infra emit them. Existing
+quotas that no longer match any tier are not rejected on read — they are enforced at the next tier up
+and flagged `isGatewayCapped` until an admin corrects them.
 
 ---
 
-## API environment variables / appsettings
+## Environment variables / appsettings
 
-Set as Container App environment variables by `infra/modules/control-plane.bicep` — the
-Bicep is the source of truth for names and values, and the
-[Infrastructure Reference](/foundry-gate/reference/infrastructure/) lists them per host.
-None of them is a secret: SQL uses Entra authentication, so even the connection string is
-plain configuration. Secrets, when one is ever needed, are set out of band in Key Vault
-and referenced from appsettings as `@KeyVault(SecretName)`.
+One table for both control-plane hosts. `infra/modules/control-plane.bicep` sets these as Container
+App environment variables and Function App application settings; the Bicep is the source of truth for
+the names, and the [Infrastructure Reference](/foundry-gate/reference/infrastructure/) covers the
+resources behind the values. None of them is a secret: SQL uses Entra authentication, so even the
+connection string is plain configuration. Secrets, when one is ever needed, are set out of band in Key
+Vault and referenced from appsettings as `@KeyVault(SecretName)`.
 
-| Key | Source | Description |
+The **Host** column is what each host *binds and can use* — ✅ set by infra, — not applicable to that
+host. Anything marked ✅ for both comes from the bicep's shared block, so the two can never be told
+about different gateways ([#108](https://github.com/kolatts/foundry-gate/issues/108); a predeployment
+test, `GatewayInfraBindingTests`, reads the bicep and fails if a `Gateway__*` name drifts from
+`GatewayOptions`, if infra sets one nothing binds, or if a new `Gateway` member arrives with no source).
+
+### Shared by both hosts
+
+| Key | API | Functions | Description |
+|---|:--:|:--:|---|
+| `AZURE_CLIENT_ID` | ✅ | ✅ | Client id of that host's user-assigned managed identity — selects it in the token credential chain |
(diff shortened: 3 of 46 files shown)


kolatts:
## Review — backend sweep A

**APPROVE**

Six issues, six commits, 1286 green, format clean, and a "For the reviewer" section that raised the three things I was going to ask about before I asked them. The `GetSystemDefaultAsync` change-tracker read is the crux and it's right — without it #193 would have faithfully written the old default back onto everyone. Two things to sort before this actually merges: the #108 tripwire will red-light sweep B for doing exactly what the tripwire's failure message tells it to do, and somebody should run one query against the deployed database before the new unique index meets it.

---

### Issues

- **Minor** `Tests.Predeployment/Core/Configuration/GatewayInfraBindingTests.cs:37` — the scrape regex understands `Gateway__X` and `Gateway__X__${i}`, but not `Gateway__X__${i}__Field`. That's the form #211 uses for `Gateway__ModelAliases__${i}__Tier`, and the form `Gateway__Tiers__{i}__ProductId` takes when #201 lands. I ran it through .NET's regex engine: those lines match nothing, so `ModelAliases` never reaches `fromInfra` and `Every_GatewayOptions_member_is_either_set_by_infra_or_listed_as_deliberately_absent` fails — with a message telling B to add the bicep lines it already added. Worth widening here rather than making B debug it.

- **Minor** `GatewayInfraBindingTests.cs:53` — `Every_Gateway_variable_infra_sets_binds_to_a_GatewayOptions_member` doesn't test its own name. The eight assertions are hardcoded property names and the binder ignores unknown keys, so adding `Gateway__Whatever` to the bicep with nothing bound to it passes green. The PR body and `configuration.mdx` both sell "fails if infra sets one nothing binds." (`Both_hosts_get_the_same_gateway_configuration` is the same shape — it greps two identifiers out of a text slice, so a comment mentioning them satisfies it.)

- **Minor** `FoundryGate.Database/dbo/Tables/QuotaIncreaseRequests.sql:47` — the new unique index is created unconditionally and there's no pre-deploy script anywhere in `FoundryGate.Database`. A fork that already has two Pending rows for one `(UserId, PeriodYear, PeriodMonth)` — the exact duplicate this PR exists to stop, which the old read-then-write check could produce — fails index creation and takes the whole dacpac step down with it, which on merge means the auto-deploy chain. Has anyone run that count against qa/prod? If it's non-zero this wants a pre-deploy dedupe or a manual cleanup first.

- **Minor** `Api/Services/Requests/QuotaRequestService.cs:457` — `Mentions` is byte-identical to `GroupService.Mentions` (line 532), doc comment included. Commit 1 of this PR deletes a duplicated `CommitToken` helper because the shared one exists; commit 3 adds a duplicated exception matcher. It wants to live next to `CommitToken` in `Data/Concurrency`.

- **Minor** `Api/Services/Config/ConfigService.cs:83` — `ExpectedUpdatedDate` is a read-then-compare on a tracked entity, with no conditional `UPDATE … WHERE UpdatedDate = @expected` behind it, so two admins who genuinely race still both pass and the second still overwrites the first. That's fine for what #170 actually asked for (two humans with a form open) — but `ClaimAsync` twelve hundred lines down in this same PR uses `ExecuteUpdateAsync` for exactly this reason, and #147's pre-check carries a comment spelling out "this is the fast path, not the guard." Same sentence would serve here; "optimistic concurrency" currently promises more than it does.

- **Minor** `ConfigService.cs:183` — `SystemDefaultUserIdsAsync` re-expresses levels 1–4 of the precedence chain as a SQL predicate while `QuotaResolutionService.ResolveLevelAsync` owns the same rule in C#. They agree today and nothing keeps them agreeing; add a level or change "any unlimited group wins" and this quietly picks the wrong user set, because both halves would still return a plausible list. Would a test asserting the two select the same users be cheap enough to bother?

- **Minor** `Core/Quota/QuotaResetService.cs:52` — the queue sweep's query runs *after* `CommitToken.For`, so a transient DB failure there discards a reset whose APIM tier moves have already been accepted, which is the gap `CommitToken` exists to close. It touches nothing external and its rows still commit with the reset if you run it above line 44.

- **Minor** `Api/Services/Requests/IQuotaRequestService.cs:159` — `ExpireStaleAsync` has no caller and its own doc says so, with no issue number attached (every other deferral in this PR cites one). It also pulls `IQuotaRequestExpiry` into `QuotaRequestService`'s constructor for every submit/approve/reject/list. Either file the issue for the operator route or drop the Api half — `QuotaResetService` already lives in Core and calls the rule directly.

- **Minor** `docs-site/…/reference/api.md:253` — `quota.reset` details are listed as `{ …, createdCount, tierSyncCount, expiredRequestCount }`. The code writes `tierChangeCount` and no `createdCount` at all, and carries a comment explaining that rename deliberately. The PR edited this exact line and carried both stale names through.

- **Minor** `docs-site/…/architecture/overview.mdx:236` — the `MonthlyResetFunction` pipeline itemises the reset's steps and still ends at "write one audit log entry." The reset now also sweeps the review queue and can write a second row. CLAUDE.md asks for behavioural changes at every level they touch, and this is the one page that makes a claim the PR falsified — the two new 409s are fine as reference-only.

- **Minor** `docs-site/…/reference/configuration.mdx:87,94` — "One table for both control-plane hosts" (there are four) and "The **Host** column is what each host binds and can use" (there is no Host column, and the `—` legend never appears in a row). Leftovers from an earlier single-table draft.

- **Nit** `Api/Controllers/ConfigController.cs:33` — `PUT /config/{key}` gained a 409 but still declares only 400/403/404. `GroupsController` declares `Status409Conflict` six times, so the house standard is clear.

- **Nit** `Core/Quota/QuotaResolutionService.cs:351` — the `Local` lookup matches the key `Ordinal` while the DB fallback is case-insensitive under SQL Server's collation, and `ConfigService.UpdateAsync:76` picks `OrdinalIgnoreCase` deliberately for exactly this reason. Contrived today, but the failure mode is the quiet one: miss the tracker hit, read the pre-edit value, write the old default onto everybody while still saving the new one.

- **Nit** `Data/Entities/QuotaIncreaseRequest.cs:101` — "written identically here and in `dbo/Tables/QuotaIncreaseRequests.sql`" isn't literally true: `[StatusType] = 0` vs `([StatusType] = 0)`. Parity passes because `NormalizeFilter` strips balanced outer parens. The `Groups.EntraGroupId` precedent has the same discrepancy and the same sentence, for what it's worth.

- **Nit** `Core/Quota/QuotaResetService.cs:107` — `expiredRequestCount` reaches the audit row and the log line but not `QuotaResetOutcome`, so an admin who hits `POST /quota/reset` and cleared six stale requests gets a response that doesn't mention it.

---

### On the three you flagged

- **#147, index placement.** The split is right, and it's the `Groups.EntraGroupId` precedent that's out of step with CONVENTIONS.md, not this. Columns, uniqueness and name are expressible as an attribute; the `WHERE` isn't. Leave it.
- **#159, no `Expired` status.** Agreed. `Rejected` + null reviewer + the system note is the shape `CancelPendingForUserAsync` already established, and a fourth enum value costs every read path a new case for a distinction the null reviewer already draws. Revisit only if a UI needs to *filter* on it.
- **#193, the union.** Fine — one row, a genuine SQL-vs-gateway disagreement corrected, and `reresolvedUserCount` records the blast radius. One asymmetry worth fixing in the prose: the docs say "a failed gateway move fails the whole edit, so the database never claims a default the gateway is not enforcing." True, and the converse now also happens — if user 200 of 400 throws, 199 subscriptions have already moved and nothing commits, so the gateway enforces a default SQL never recorded. Same shape the reset has and #199 is the right home for it, but the sentence currently only tells the reassuring half.

---

### Merge order

**This one first, then rebase #211 (sweep B), then #210 (frontend follow-ups).**

1. **#211 needs the regex fix from here.** Left as-is, B has to widen `GatewaySettingRegex` itself so `Gateway__ModelAliases__${i}__Tier` reduces to `ModelAliases`. Fix it in this PR and B just adds its property and its bicep lines, and the alarm stays quiet — which is what a good tripwire does.
2. **#211 asserts something this PR makes false.** "changing the system-wide `DefaultMonthlyTokenQuota` re-resolves nobody" appears in B's new "Why the Functions identity can write to APIM" section in `reference/infrastructure.md`, in `FunctionsServiceCollectionExtensions`'s comment, and in a `MonthlyResetJobTests` comment. The other half of B's argument (a developer with no earlier allocation has no known previous tier) survives, so the case for the RBAC grant still stands — the leading justification just needs rewriting on rebase.
3. Textual conflicts with #211 are all trivial: `AuditActions.cs` (different regions), `FunctionsServiceCollectionExtensions.cs` (adjacent lines), and four test files that only gain constructor arguments.

**#210 is the sharper collision.** It rewrites `ConfigService.ListAsync`, adds a sixth positional member to `SystemConfigEntryResponse`, and inserts `SystemConfigValidator.EnsureEditable(entry.Key)` at the *exact* line this PR puts the `ExpectedUpdatedDate` check. Whoever rebases has to pick which 409 wins for a stale write to a system-managed key — I'd put `EnsureEditable` first: "this key is never editable" is a stronger statement than "your copy is stale," and it needs no second query.

kolatts:
## Fix pass — point by point

Two commits: `871cdcd` (code) and `9eca976` (verification, tests, docs). 1298 green, 0 warnings, format clean, docs build clean, CI green.

Two of these turned out to be real bugs rather than tidy-ups — the tripwire one you predicted, and one the #193 tests caught on the way through. Both called out below.

---

### `GatewayInfraBindingTests` — the regex does not understand `Gateway__X__${i}__Field`

Fixed, and widened past what #211 needs. The scrape now models three shapes — `Gateway__Member`, `Gateway__Member__${i}`, `Gateway__Member__${i}__Field` — and reduces all three to a member name, so `Gateway__ModelAliases__${i}__Tier` lands on `ModelAliases` and `Gateway__Tiers__${i}__ProductId` on `Tiers`.

Two things beyond the ask:

- `A_nested_field_setting_binds_to_the_list_items_property` runs the full parse-and-bind path over a synthetic `Gateway__Tiers__${i}__{ProductId,DisplayName,MonthlyTokenQuota}` block and asserts the values arrive on `GatewayTier`. `Tiers` exists today, so it is a real end-to-end proof that a nested-field key reaches its property, not just that the regex matches the string.
- A `[Theory]` pins each shape separately, so a regex that stops matching one fails with "the scrape missed this" rather than a downstream "member unaccounted for" a long way from the cause. `ModelAliases` is in that theory now, with no property behind it — parsing has to work before B lands.

A field with no `${i}` between it and the member (an object-valued setting) is refused with a message rather than silently mis-modelled.

### The other tripwire doesn't test its own name

You were right, and it was worse than the eight hardcoded assertions: the binder ignores keys it cannot place, so `Gateway__Whatever` bound to nothing and passed.

It now derives everything. Scrape the bicep → build a configuration from the scraped keys with a distinct generated value each (type-aware, so a `long` cap gets a number) → bind a real `GatewayOptions` → check **by reflection** that every scraped member came back carrying exactly what went in, including per-index and per-field for collections. Distinct values matter: a member bound from the wrong key is now a value mismatch, not an accidental pass.

And the alarm's own claim is tested. `The_binding_check_fails_when_infra_sets_something_nothing_binds` is a theory over three ways to break it — an unknown member, an indexed key against a scalar property, an unknown field on a list item — asserting each produces a failure naming the cause.

`Both_hosts_get_the_same_gateway_configuration` was the same shape of hole; it strips `//` comments before the identifier check now, so prose mentioning `sharedAppConfig` no longer satisfies it.

### Pre-deploy dedupe before the unique index

Nobody had run that count, so I built the failure and fixed it rather than guessing.

`FoundryGate.Database/Scripts/PreDeployment.sql`, declared `<PreDeploy>` in the `.sqlproj` — and removed from the SDK's `**/*.sql` `Build` glob, which otherwise tries to parse DML as schema. It keeps the **newest** pending request per `(UserId, PeriodYear, PeriodMonth)` (most recent justification; `QuotaIncreaseRequestId` breaks ties, which the race that produced them makes likely) and closes the rest as `Rejected` with `ReviewNotes = 'Superseded by a later request (pre-deploy dedupe)'` and no reviewer — the #159 lapsed-request shape.

**Verified against SQL Server 2022 in docker, not just built.** Deployed clean, dropped the index, seeded two pending rows for one (user, period) plus a control row in another period, redeployed:

```
Pre-deploy: closed 1 duplicate pending quota increase request(s) so IX_QuotaIncreaseRequests_PendingPerUserPeriod can be created (#147).
Creating Index [dbo].[QuotaIncreaseRequests].[IX_QuotaIncreaseRequests_PendingPerUserPeriod]...
Deployed FoundryGatePreDeployProbe successfully.
```

Newest row still `Pending`, older `Rejected` with the note and no reviewer, other period untouched. A third deploy is silent — idempotent. Also confirmed the script is embedded in the dacpac as `predeploy.sql`, so `db deploy`'s `DacServices` path picks it up with nothing extra wired.

One thing that fell out of testing and is now documented: a filtered index makes `SET QUOTED_IDENTIFIER ON` mandatory for DML on that table. EF Core, SqlClient and DacFx all set it, so nothing in the app or the pipeline is affected — but a hand-run `sqlcmd -Q` against `QuotaIncreaseRequests` fails with `Msg 1934` without it. Worth knowing before someone debugs it at 2am.

### `Mentions` duplicated

Gone. `FoundryGate.Data.Concurrency.UniqueIndexViolation.Mentions`, next to `CommitToken` as you suggested, with the two-markers-per-index convention (SQL Server names the index, SQLite names the columns) documented once instead of twice.

### `ExpectedUpdatedDate` is a read-then-compare

Agreed, and it promised more than it delivered, so I made it deliver rather than downgrading the prose.

`ConfigService.ClaimAsync` writes the row with one conditional `UPDATE … WHERE [Key] = @key AND [UpdatedDate] = @expected`, inside a transaction opened before anything can reach the gateway — `ApproveAsync`'s shape exactly, for the same reason (a failed tier move rolls the claim back). Zero rows re-reads to distinguish "someone else wrote it" (409, naming the current value and editor) from "it was deleted" (404). The pre-check above it keeps its 409 and now carries the "this is the fast path, not the guard" sentence, pointing at #147's.

`A_concurrent_write_between_the_check_and_the_claim_loses_with_a_409` drives a competing write in through the interceptor hook between the pre-check and the statement, which is the only way to reach the guard.

**This turned up a real bug.** With `ExecuteUpdate` the row is no longer written through the change tracker — but the reference-data seeder leaves those rows *tracked*, and `GetSystemDefaultAsync` reads the tracker first. So the re-resolution read the stale pre-edit copy and wrote the **old** default onto every default-tier developer while still committing the new one — precisely the #193 bug, reintroduced through the back door. `ClaimAsync` now detaches the tracked rows for that key, exactly as `QuotaRequestService.ClaimAsync` does; there it is tidiness, here it is load-bearing. The #193 tests caught it immediately, which was reassuring.

### `SystemDefaultUserIdsAsync` re-expresses levels 1–4

Cheap enough, so yes. `The_affected_user_query_selects_exactly_the_users_the_resolution_chain_puts_on_the_system_default` seeds one user per level (plus a deactivated one), asks `PreviewAsync` — the service that owns the rule — which of them land on `SystemDefault`, then runs the edit and asserts the users that got a re-resolved allocation are exactly that set. Add a level or change "any unlimited group wins" and the two disagree loudly.

### Queue sweep runs after the commit point

Moved above `ResolveManyAsync`. Its rows still commit with the reset, but a transient failure in its query now costs nothing instead of discarding a reset APIM has already accepted, and it runs on the caller's own token. Comment says why it is where it is.

### `ExpireStaleAsync` has no caller

**Wired it** rather than filing an issue, and rather than dropping the Api half.

`POST /api/v1/requests/expire-stale`, admin-only, returns `{ expiredCount }`. The case: between a period ending and the next reset, those requests are already unapprovable but still clutter the queue, and the only way to clear them today is `POST /quota/reset` — which also re-resolves every allocation and can move tiers at the gateway. A sweep-only action is strictly cheaper. It also turns a seam that was only ever exercised by the reset into tested behaviour, which is most of why the "no caller" smell was worth taking seriously. Auth-matrix test plus a behaviour test; the interface doc no longer says "nothing calls this".

### `api.md` audit details, `overview.mdx` reset pipeline, `configuration.mdx` leftovers, `ConfigController` 409

All fixed.

- `quota.reset` details are `{ usersResetCount, periodYear, periodMonth, tierChangeCount, expiredRequestCount }`, with a line explaining the deliberate `tierChangeCount` naming so the next person does not "correct" it back. The response gains `expiredRequestCount`, and `POST /requests/expire-stale` is in the endpoint table.
- The `MonthlyResetFunction` pipeline gains a "Close quota requests the period left behind" step (first, and it says why it is first) and its last step is now "Write the audit log entries", plural, covering the conditional second row.
- `configuration.mdx` loses "One table" and the Host-column legend. The shared table's API/Functions columns were `✅ ✅` on every row under a heading that already said "Shared by both hosts", so they carried no information and are gone.
- `ConfigController` declares `Status409Conflict` with a `<response code="409">`.

### `QuotaResetOutcome` missing `expiredRequestCount`

Added, and carried through `QuotaResetResult` so `POST /quota/reset` actually reports it. Asserted on the outcome in the reset test.

### `QuotaResolutionService` `Local` lookup is `Ordinal`

Fixed to `OrdinalIgnoreCase`, matching `ConfigService`. You are right that the failure mode is the quiet one. Its comment also no longer claims `PUT /config` depends on that read — after the conditional UPDATE it does not. The read stays, because resolution reading this key differently from how it reads allocations and group policies is what made the bug possible in the first place, and #210 rewrites this service.

### `written identically` vs the outer parens

Left as-is, deliberately. It matches the `Groups.EntraGroupId` precedent, `NormalizeFilter` is what makes it true in the sense that matters, and rewording one of the two without the other would be worse. Happy to fix both in a follow-up if you want the sentence literally true.

---

### On the merge order

Agreed, and #211's rebase should be easier now: the regex handles `Gateway__ModelAliases__${i}__*`, so B adds its property and its bicep lines and the alarm stays quiet. Worth flagging to B that the tripwire is stricter than it was — it now *binds* rather than pattern-matching, so a `ModelAliases` property whose element type does not carry a settable `Tier`/`Alias`/`DeploymentName`/`Provider` will fail with a message naming the field.

The three statements #211 has to rewrite on rebase ("changing `DefaultMonthlyTokenQuota` re-resolves nobody") are all false now — `reference/infrastructure.md`, `FunctionsServiceCollectionExtensions`, `MonthlyResetJobTests`. The second half of B's argument still stands, as you said.

For #210: `EnsureEditable` first is right. A stale write to a never-editable key should hear "this key is never editable" — and it needs no query. It slots in above the `ExpectedUpdatedDate` pre-check, before `ClaimAsync` and well before anything can reach the gateway.

Issue #208, opened 2026-09-02
The dashboard can count hard-stopped and over-budget users but nothing can list them


kolatts:
## Problem

#190 added `hardStoppedUserCount` and `overBudgetUserCount` to `GET /dashboard`, and `/dashboard`
now shows both. Neither is reachable as a *list*: there is no way to answer "who are those two
people?" from the portal.

- `GET /users` filters on `search` and `isActive` only (`UserListQuery`).
- `GET /quota/allocations` is paged over the current period but has no filter for `isHardStopped` or
  for usage against the cap, and there is no Web page over it at all.

The dashboard's hard-stopped card therefore links to `/audit?action=key.revoked` — the trail that
*produces* the state, which is genuinely useful but is not the same question. The over-budget count
does not link anywhere.

## What's needed

Pick one:

1. **Filters on `GET /quota/allocations`** — `?isHardStopped=true`, `?overBudget=true` — plus a
   `/quota` admin page (a grid of this period's allocations with the gauge `/me` already renders).
   That is the natural home: these are allocation facts, not user facts.
2. **Filters on `GET /users`** — cheaper (the page exists) but pushes period-scoped allocation state
   into a query about the user table, where it does not really belong.

(1) is probably right. Either way the dashboard cards then link to a filtered list, which is what
#54 asked for.

## Found

Building #190. The card was shipped linking to the audit log as the closest honest destination.

kolatts:
Claiming this for the final sweep C branch (claude/final-sweep-c).

Issue #207, opened 2026-09-02
Two bUnit tests fail intermittently under parallel load (Dashboard refresh loop, Me rotate confirmation)


kolatts:
## Problem

Two Web component tests fail roughly one full-suite run in five on a loaded machine, and pass on
their own every time:

- `Web.DashboardPageTests.Stops_refreshing_once_the_page_is_gone`
- `Web.MePageTests.Rotating_asks_for_confirmation_before_it_touches_the_api`

The dashboard one is the same shape #195 already tightened once: it drives a `PeriodicTimer` at
20 ms and then asserts the count stops moving. `SettledCallCountAsync` polls for a settled count,
but the final `await Task.Delay(200)` is still a fixed window — a tick that the thread pool
delivers late lands inside it. The `/me` one waits on a MudBlazor dialog appearing.

Neither is a product bug: both assert real behaviour and both pass in isolation. But an
intermittently red suite trains people to re-run rather than read, which is the expensive part.

## Options

- Drive the dashboard loop from a controllable clock (a fake `PeriodicTimer` seam, or the injected
  `TimeProvider`) instead of a real 20 ms interval, so "the loop stopped" is a deterministic claim.
- Replace the trailing fixed delay with a bounded wait for "the count did not move across N
  consecutive samples".
- For `/me`, wait on the dialog with `WaitForElement` rather than the current sequence if it is not
  already doing so.

## Found

Running the Predeployment suite repeatedly for the #190/#191/#171/#173/#172 frontend follow-ups.
Both predate that branch.

kolatts:
Duplicate of #203 (`Rotating_asks_for_confirmation…` sibling flake of the fixed #195; `Stops_refreshing_once_the_page_is_gone` was hardened in PR #176 — if it still flakes, note it on #203). Tracking there. — orchestrator

Issue #206, opened 2026-09-02
HealthEndpointTests passes only where no SQL Server is reachable — its connection-string override never reaches the bound settings


kolatts:
## Problem

`Api/Endpoints/HealthEndpointTests` overrides the connection string to a closed port so
`/health/ready` deterministically degrades:

```csharp
builder.ConfigureAppConfiguration((_, configurationBuilder) =>
    configurationBuilder.AddInMemoryCollection(new Dictionary<string, string?>
    {
        ["ConnectionStrings:FoundryGate"] = "Server=127.0.0.1,1;...",
    }));
```

That override never lands. CONVENTIONS.md already records why, in the `ApiTestFactory` note:
with minimal hosting, `ConfigureAppConfiguration`'s sources are appended only *after* `Program.cs`
has run `Configuration.Get<AppSettings>()` — so the host keeps whatever `appsettings.local.json`
says, which is the docker SQL address `127.0.0.1,3433`.

On a machine where the repo's own `foundrygate-sql-server` container is running (i.e. any machine
where someone has run `local setup`), `/health/ready` reports **Healthy** and
`HealthReady_returns_503_rather_than_throwing_when_the_database_is_unreachable` fails. It passes in
CI only because CI has no SQL Server.

## Fix

Use `IWebHostBuilder.UseSetting` for the connection string, exactly as `ApiTestFactory` does and for
exactly the same reason. Host settings are in place before `Program.cs`'s first line.

## Found

While running the Predeployment suite for the #190/#191/#171/#173/#172 frontend follow-ups. It is
the only failure on a developer machine with docker up, and it is unrelated to those changes.

kolatts:
Duplicate of #197 (same root cause: the readiness test's `ConfigureAppConfiguration` override never reaches the bound settings, so it passes only when no local SQL Server is reachable). Tracking there. — orchestrator

PR #210, opened 2026-09-02
feat(web): dashboard hard-stop counts, audit actor type-ahead, durable sync metadata, Foundry catalogue

### LINKED ISSUE #171: Persist last-Entra-sync metadata so /users/sync can show it across sessions

### LINKED ISSUE #172: The read-only SystemConfiguration key list lives in two places (Api validator + Web config page)

### LINKED ISSUE #173: Serve the Foundry create form a real model/SKU catalogue instead of a hardcoded list

### LINKED ISSUE #190: Dashboard has no hard-stopped user count, so /dashboard shows tokens-used instead
Issue #54 asks the admin dashboard for a "hard-stopped users" stat card. `DashboardSummaryResponse` (`GET /dashboard`, #162) carries no such figure — `totalUserCount`, `activeUserCount`, `unlimitedUserCount`, `pendingQuotaIncreaseRequestCount`, `totalTokensUsedThisPeriod`, `topConsumers` — so PR #174 shipped "Tokens this period" in that slot instead, which is a useful number but not the one #54 named.

Hard-stopped is `QuotaAllocation.IsHardStopped`, set when an admin revokes a key (spec §5.3) — never by quota exhaustion. It is exactly the "who is broken right now" number an admin wants on a landing page, and today the only way to find those users is to open allocations one at a time.

Work: add `hardStoppedUserCount` to `DashboardSummaryResponse` (count of active users whose current-period allocation has `IsHardStopped`), fill it in `DashboardService`, and give it its own card on `/dashboard` linking to the affected users. The card should read as an alert when it is non-zero — a revoked key is an outage for that developer.

### LINKED ISSUE #191: C:/Program Files/Git/audit filters by actor user id, not actor name — needs a type-ahead over GET /users
Issue #55 asks the audit viewer for an actor filter that is a "type-ahead from the user list". PR #174 shipped a raw numeric "Actor user id" field instead, because `AuditLogQuery` only accepts `actorUserId` and the Web client had no search-capable `GET /users` overload at the time (#176 adds `GetUsersAsync(UserListQuery, PagedRequest)`).

An admin reading the audit log knows "Ada Lovelace", not "user 41", so today the filter is unusable without a detour through `/users` to look the id up — which is the opposite of what a filter is for.

Work, once #176 has landed its `UserListQuery` client overload:

- Replace the numeric field on `/audit` with a `MudAutocomplete<UserResponse>` searching `GET /users?search=` (debounced), resolving the selection to `actorUserId` for the query.
- Keep the id path working for a deep link — an `?actor=` query parameter on the route should still apply.
- Show the actor's display name in the grid where the API supplies `actorDisplayName`, and fall back to `user #id` (already the case).

kolatts:
Five frontend follow-ups from the #174/#176 admin-UI wave, each with the API work it needed.

## #190 — the dashboard's fourth card is hard-stopped users, not a tokens substitute

`DashboardSummaryResponse` gains `hardStoppedUserCount` and `overBudgetUserCount`, **appended** to
the positional record (the Web client deserializes it by position).

- `hardStoppedUserCount` — active users whose current-period allocation carries `IsHardStopped`. The
  flag is set by the deprovision pipeline and cleared by re-activation and the monthly reset; quota
  exhaustion never sets it, because the gateway 403s an over-budget request itself. An *active* user
  carrying it is therefore an inconsistency, and normally the count is zero.
- `overBudgetUserCount` — active users whose finite budget reconciled usage has reached or passed
  (`>=`, because the gateway refuses the request that would cross the cap). This is the number an
  admin reads day to day, which is why both were added rather than only the one #190 named.

`/dashboard` renders the hard-stopped card in the slot #54 asked for: an alert when non-zero, linking
to `audit?action=key.revoked` — so `/audit` now honours an `?action=` query parameter (an action the
Domain constants don't name is ignored). "Tokens this period" stays on the page, with the over-budget
count, as a caption above the consumers grid it describes.

## #191 — `/audit`'s actor filter is a type-ahead

The raw numeric "Actor user id" field needed a detour through `/users` before it could be used. It is
now a `MudAutocomplete<UserResponse>` over `GET /users?search=` (300 ms debounce, active and
deactivated alike — a departed employee's entries are still in the log) resolving to `actorUserId`.

The id path stays reachable two ways: an all-digits search term also asks `GET /users/{id}` and
offers that person, and `?actor=41` is a deep link that applies the id and resolves the name for the
box — still filtering when the name cannot be resolved.

## #171 — the last Entra sync is remembered

`/users/sync` showed only the run you triggered in this browser session, and nothing on first load;
#63 assumed `LastUserSync*` keys that never existed.

- Two new seeded keys, `LastUserSyncDate` (ISO-8601) and `LastUserSyncResult` (the run's
  `UserSyncResult` as camelCase JSON), empty until a fork has run a sync.
- `EntraUserSyncService` writes both in the run's own unit of work, alongside the `users.synced`
  audit row. A database that predates the keys gets them inserted rather than losing the record.
- `GET /api/v1/users/sync/last` (admin) returns `{ lastSyncDate, lastResult }`. It answers where
  `Entra:Enabled` is false — this is a question about the database, not the directory — and
  unreadable stored JSON reads as "no result", never an error.
- Both keys are **system-managed**: `SystemConfigValidator.EnsureEditable` refuses a `PUT` with a
  `409` naming the reason, and `SystemConfigEntryResponse` gains `IsReadOnly` (appended) so `/config`
  disables the field without a round trip.

## #173 — a real model catalogue for the create form

`GET /api/v1/foundry/catalog` (admin) lists `{ modelFormat, modelName, modelVersion, skuNames,
defaultCapacity }` from ARM's account-scoped model list
(`CognitiveServicesAccountResource.GetModels`), merged across accounts by (format, name, version)
with SKUs unioned, behind a 5-minute cache. A missing account is skipped with a warning.

The dialog's model, version and SKU pickers read it and the hardcoded `KnownModels` / `KnownSkus`
arrays are **gone** rather than kept as a fallback — a stale list is what #173 was filed about, so
the fallback is plain free text plus a note when the catalogue can't be read. Picking a catalogued
model fills in its newest version, first SKU and suggested capacity, never overwriting something
already filled in or a capacity that was typed. Coercion stays everywhere.

## #172 — one source of truth for read-only config keys

Verified and closed with evidence. `grep -rn "ReadOnlyReason|SystemConfigurationKeys|ApimProductId|EntraTenantId|ApimGatewayUrl" src/FoundryGate.Web` returns **nothing** — the Web page has no knowledge
of configuration keys at all. #186 had already deleted the shipped retired keys from the Domain
constants, so the map #172 asked for now has exactly one writer,
`SystemConfigurationKeys.SystemManaged` (added with #171), and two readers: the API's `409` and the
`IsReadOnly` flag the page renders.

Three tripwires keep it that way:
- `SystemConfigValidatorTests.Every_system_managed_key_is_refused_and_no_other_is` — the validator
  against the map, over every seeded key.
- `ConfigEndpointTests.The_list_flags_which_keys_the_editor_must_not_offer_to_change` — the wire's
  `isReadOnly` against the same map, over every row.
- `ConfigPageTests` — a key the page has never heard of is disabled when the API flags it, and a key
  the Domain calls system-managed is still editable when the API does not. That second case is the
  exact shape of the bug #172 named, where the page was the stricter of the two.

## Verification

- `dotnet test src/FoundryGate.Tests.Predeployment` — **1317 passed, 0 failed** (main was 1270)
- `npm run build` in `docs-site/` — 15 pages, clean
- Docs updated: `reference/api.md` (the two dashboard counts, `GET /users/sync/last`,
  `GET /foundry/catalog`, `isReadOnly`, and a corrected `/audit` filter list — it documented
  `?actor=` where the API takes `actorUserId`), `reference/web-ui.md` (all four pages),
  `reference/configuration.mdx` (a "System-managed keys" section beside "Retired keys")
- Client route tripwires extended: `GET /api/v1/users/sync/last` and `GET /api/v1/foundry/catalog`
  resolve against the API's own routing table

## Decisions worth a reviewer's eye

1. **Both dashboard counts, not one.** #190 names hard-stopped, but nothing in the codebase sets
   `IsHardStopped` on a user who stays active (`DELETE /keys/{userId}` explicitly leaves `IsActive`
   alone and does not hard-stop), so `hardStoppedUserCount` is normally zero by construction. It is
   still the right alert — it means the allocation and the account disagree — but `overBudgetUserCount`
   is the number an admin actually reads. Both are on the wire and both are on the page.
2. **The hard-stopped card links to the audit log, not to a user list.** There is no way to *list*
   these people: `GET /users` filters on search and active only, and `/quota/allocations` has no
   hard-stopped or over-budget filter and no Web page. `audit?action=key.revoked` is the closest
   honest destination. Filed as #208.
3. **No hardcoded SKU default survives #173.** `_skuName` now starts empty, so a create with an
   unreadable catalogue needs the SKU typed. Keeping `"GlobalStandard"` as a default would have kept
   exactly the hardcoded knowledge the issue asked to remove; the dialog says so instead.
4. **`?action=` on `/audit` landed in the #190 commit**, because the dashboard card links to it.
   `?actor=` landed with #191.
5. **`EnsureEditable` is a `409`, not a `403`.** The caller's permissions are fine; the row is not
   theirs to set. It is also not a `404` — unlike a retired key, the row exists and is worth reading.

## Discovered and filed

- #206 — `HealthEndpointTests` passes only where no SQL Server is reachable: its
  `ConfigureAppConfiguration` connection-string override never reaches the bound settings (the exact
  caveat CONVENTIONS.md records for `ApiTestFactory`), so on a machine with the repo's own docker SQL
  up, `/health/ready` reports healthy and the test fails. Pre-existing; unrelated to this branch.
- #207 — two bUnit tests fail intermittently under parallel load
  (`DashboardPageTests.Stops_refreshing_once_the_page_is_gone`,
  `MePageTests.Rotating_asks_for_confirmation_before_it_touches_the_api`). Both pass in isolation and
  both predate this branch.
- #208 — nothing can list the hard-stopped or over-budget users the dashboard now counts.

Closes #190
Closes #191
Closes #171
Closes #173
Closes #172


kolatts:
**REQUEST CHANGES**

Five issues, five clean landings — the sync-metadata work and the #172 tripwires are the best part of this branch, and the docs are genuinely better than the code they describe. The create dialog is where it comes apart: it now offers Claude models to a form that hardcodes `ModelFormat = OpenAI`, which routes straight past the refusal that exists to keep us out of E-007. Everything else is small.

Verified on the branch: `dotnet build -c Release` 0/0, `dotnet test FoundryGate.Tests.Predeployment -c Release` 1317 passed, `dotnet format --verify-no-changes` clean. Both Major findings below are confirmed with throwaway bUnit tests, not read off the diff.

---

**Issues**

- **Major** `src/FoundryGate.Web/Shared/FoundryDeploymentDialog.razor:167` (`SearchModelsAsync`) — The model picker is fed the whole catalogue with no format filter, so it now suggests `claude-*`, while `SubmitAsync:253` hardcodes `ModelFormat = FoundryModelFormatType.OpenAI` — so `FoundryDeploymentService`'s Anthropic refusal never fires and the create goes to ARM instead, which is precisely the failed Anthropic create E-007 says can wedge the subscription's Marketplace agreement. The file's own header says this dialog "never offers one and says why instead of letting the user find out from an error"; it does now. `GET /foundry/catalog` can keep listing Anthropic for visibility — the *dialog* is the thing that has to filter.

- **Major** `src/FoundryGate.Web/Shared/FoundryDeploymentDialog.razor:206-213` (`OnModelChangedAsync`) — Version and SKU are filled only when the field is empty, so changing your mind about the model leaves the *previous* model's version and SKU in the boxes and posts them. Picking `gpt-5-codex` then switching to `gpt-4.1-mini` submits `gpt-4.1-mini` / `2026-01-01` / `ProvisionedManaged`. Capacity, notably, does re-fill (`_capacityEdited` is still false), so the three fields disagree about whether an auto-fill counts as the admin's answer.

- **Minor** `src/FoundryGate.Api/Services/Foundry/ArmFoundryManagementClient.cs:214-232` (`MapCatalogEntry`) — ARM hands you `IsDefaultVersion`, `Deprecation`, `LifecycleStatus` and per-SKU `MaxCapacity` on `CognitiveServicesAccountModel` and the mapping drops all of them. So "newest version" is inferred from `ThenByDescending(ModelVersion, OrdinalIgnoreCase)` — which picks `turbo-2024-04-09` over `2025-04-14`, and `1106` over `0125` — and a model Azure retired next month is still offered first. That's the staleness #173 was filed about, moved server-side.

- **Minor** `src/FoundryGate.Api/Services/Foundry/ArmFoundryManagementClient.cs:230` — `DefaultCapacity` comes from the first SKU ARM lists, but the dialog auto-selects `SkuNames[0]` from an alphabetically re-sorted list, so the suggested capacity routinely belongs to a different SKU than the one pre-selected. Capacity limits are per-SKU, so the head start can be a create ARM refuses.

- **Minor** `src/FoundryGate.Api/Services/Entra/EntraUserSyncService.cs:168` — `RecordLastSyncAsync` and the save that commits it run on the caller's token, *after* the departure loop has already deleted APIM subscriptions. CONVENTIONS.md's commit-point rule wants `CommitToken.For(deactivated.Count > 0, cancellationToken)` here. The final save was already on the caller's token, granted — but this PR gives it durable state whose whole point is surviving, and a disconnect now loses the record of a run that really did revoke keys.

- **Minor** `src/FoundryGate.Web/Pages/UsersSync.razor:124` + markup — A failed `GET /users/sync/last` leaves `_lastSync` null, which renders as "This fork has no record of a previous sync." The comment says the failure is "deliberately silent", but the page isn't silent — it asserts a fact about the database it never learned. "Couldn't read the last run" and "there was no last run" want different words.

- **Minor** `src/FoundryGate.Web/Pages/Audit.razor:289` (`TryLoadUserAsync`) — The numeric fallback calls `GET /users/{id}`, the heavyweight detail endpoint (three round trips plus masked key material) to render `"Name (#41)"`, on every all-digits keystroke. `GET /users?search=<id>` or a lighter lookup does the same job.

- **Nit** `src/FoundryGate.Web/Pages/Audit.razor:244-262` — `OnParametersSetAsync` re-applies `?actor=`/`?action=` and the URL is never rewritten when the filter changes, so a same-route navigation quietly restores the deep-link value the admin just cleared. The XML doc says "applied on first render only", which `OnInitializedAsync` would give you and `OnParametersSetAsync` doesn't.

- **Nit** `src/FoundryGate.Api/Services/Entra/EntraUserSyncService.cs:252` — `Value()` re-matches with `OrdinalIgnoreCase` "so the comparison does not depend on the provider's collation", but the `Where(c => c.Key == …)` two lines up already did the comparison in the provider — on a case-sensitive collation the rows are gone before `Value` sees them. Harmless, but the comment isn't true.

- **Nit** `src/FoundryGate.Api/Services/Foundry/FoundryDeploymentService.cs:145` — An empty catalogue is cached for the full five minutes and nothing invalidates `CatalogCacheKey`, so a transient window where the accounts 404 pins "catalogue unavailable" on the dialog with no way to refresh. Consider not caching the empty answer.

- **Nit** Doc drift introduced here: `ConfigService.ListAsync:59` says "this five-row table" (it's seven now, and `GetLastSyncStatusAsync` says so two files over), and `DashboardService.cs:11,17` still say "six set-based queries" / "N × 6 aggregate queries a minute" while the method doc below was updated to eight.

- **Nit** `src/FoundryGate.Web/Pages/UsersSync.razor:98` uses `lastSyncDate.ToLocalTime()`, where `Audit.razor` goes through the injected `TimeProvider.LocalTimeZone` for the same job.

---

**Not issues, for the record**

- No CLAUDE.md / CONVENTIONS.md violations found: controllers are thin and expression-bodied, no naked `UtcNow`, `CancellationToken` last on every new async method, `AsNoTracking` on every new read path, `[DoNotUpdate]` on `Value` already covers the two new seeded rows, and a convention test pins `SystemManaged ⊆ All` and `SystemManaged ∩ Retired = ∅`. Web still references Domain only.
- The dashboard counts are right: `CurrentPeriod` is `AsNoTracking`, the `(UserId, PeriodYear, PeriodMonth)` unique index means counting allocations *is* counting users, both are scoped to active users, and the cache key is still per-period.
- `EnsureEditable` is correctly ordered after the 404 and before `Normalize`, and `IsReadOnly` is appended to the end of the positional record.
- Route tripwire rows are in for both new endpoints.
- On decision 1 in the PR body — agreed that `hardStoppedUserCount` is zero by construction, and the honesty about it is appreciated. Two smaller thoughts: the card links to `audit?action=key.revoked`, but the pipeline that *sets* `IsHardStopped` writes `user.deactivated` (with `allocationHardStopped: true` in details), and `key.revoked` is also written by `DELETE /keys/{userId}`, which explicitly does not hard-stop — so the link lands on a set that is mostly not these people. And spending the card slot on a tripwire while the number an admin reads daily is a caption is a choice worth revisiting when #208 lands.

---

**Merge order**

Recommend **#210 → #204 → #211**.

- **#210 vs #204** is the collision that matters. Both rewrite `ConfigService.UpdateAsync` — #210 inserts `EnsureEditable` and makes the closing `new SystemConfigEntryResponse(...)` six arguments; #204 rewrites the whole tail for the quota re-resolution commit point and still constructs it with five. That is a build break, not a text conflict, if it merges cleanly. Landing #210 first means #204 rebases its larger change onto a small, visible one. After the merge, the refusal order should read: 404 unknown key → 409 system-managed → 409 stale `ExpectedUpdatedDate` → 400 invalid value. They touch different records in `ConfigContracts.cs`, so that file is fine.
- **#210 vs #211** overlaps on seven Foundry files (`IFoundryManagementClient`, `ArmFoundryManagementClient`, `IFoundryDeploymentService`, `FoundryDeploymentService`, `FoundryController`, `FoundryContracts`, `FakeFoundryManagementClient`) plus both Foundry test files — but #211's `PATCH …/capacity` and #210's `GET /catalog` are purely additive members alongside each other. Mechanical either way; last one rebases.
- All three touch `reference/api.md` and two touch `configuration.mdx`, in non-overlapping sections.

Fix the two dialog Majors and this is a straightforward approve.

kolatts:
Fix pass pushed as `b59f20c`. Both Majors fixed, all five Minors fixed, four of the five Nits fixed — one answered rather than changed. Gates: build 0/0, **1333 tests pass** (was 1317), `dotnet format --verify-no-changes` clean, docs build clean.

---

## Majors

**1. The dialog offered Claude to a form that hardcodes OpenAI.** Fixed, and you were right that the endpoint is not the place for it: `GET /foundry/catalog` still lists Anthropic. The dialog now filters to OpenAI **once, on load**, before any picker sees the list — `IsCreatableHere` runs in `OnInitializedAsync`, not per-`SearchFunc` — so no field can offer a Claude model by a route the model picker didn't. The banner stays, and there is a test for it plus one for the degenerate case where the catalogue is *only* Anthropic (it reads as "no suggestions", not as an empty menu). You'd caught that the file's own header had become untrue; it now says what the filter is for.

**2. Version and SKU only auto-filled when empty.** Fixed as you framed it — the three fields now agree that a derived value follows the model. `_versionEdited` / `_skuEdited` / `_capacityEdited`, each set only by the field's own `ValueChanged`; `ApplyModelDefaults()` re-derives anything not flagged. Your exact repro is a test (`Changing_the_model_re_derives_the_version_sku_and_capacity` — pick `gpt-5-codex`, switch to `gpt-4.1-mini`, assert all four values are the mini's), and so is the other half (`A_version_or_sku_the_admin_typed_survives_a_change_of_model`).

## Minors

**ARM metadata dropped in `MapCatalogEntry`.** Fixed — this was the best catch in the review, because "staleness moved server-side" is exactly what it was. The contract gains `IsDefaultVersion`, `LifecycleStatus`, `InferenceRetiresOn` and `DefaultSkuName`. Ordering is now model name → ARM's default version → version string *only as a tiebreak where ARM flags no default*, and there's a test seeded with your own example (`turbo-2024-04-09` vs `2025-04-14`). Retired models (`Deprecated`, or past `inferenceRetiresOn`) are hidden in the dialog behind a toggle that only appears when there are any — hidden rather than removed, because deploying one is legitimate when matching an existing deployment, just not something to scroll past. `FoundryCatalogEntryResponse.IsRetiredAt(asOf)` lives on the contract so every reader answers it identically, and takes the clock rather than reading `UtcNow`.

**`DefaultCapacity` belonged to a different SKU than the pre-selected one.** Fixed. `DefaultSkuName` and `DefaultCapacity` are now taken as a pair from the same SKU — ARM's first — and `Merge` keeps them together across accounts. `SkuNames` stays sorted for the dropdown, with a comment saying sorting is a display decision and must not become a choice.

**`RecordLastSyncAsync` on the caller's token.** Fixed: `CommitToken.For(deactivated.Count > 0, cancellationToken)`, covering the config rows, the audit row and the save. Two tests, one per branch of the predicate — a run that deleted a subscription records itself with the token cancelled the instant APIM accepted (via `FakeApimManagementClient.AfterMutation`), and a run that reached nobody's subscription still stops. That needed one new hook on `FakeEntraDirectoryClient` (`AfterListAssignedUsers`), mirroring `FakeFoundryManagementClient.OnCreate`.

**A failed `GET /users/sync/last` rendered as "no record".** Fixed — separate `_lastSyncUnavailable` state, the alert goes Warning, and the text is "Couldn't read the last run: …". Still not a snackbar: the page's job is running a sync. Test asserts the two strings don't substitute for each other.

**`GET /users/{id}` per keystroke.** Fixed, but not the way you suggested, and it's worth saying why: `GET /users?search=` matches display name and email only, so it cannot resolve an id at all — there is no lighter lookup that answers this question. Instead the search does names/emails only, and the id is resolved on **blur**: type `412`, tab out, one detail read. Tests pin that three searches over `4` → `41` → `412` reach the detail endpoint zero times, that blur then reaches it exactly once, and that an id matching nobody leaves the filter alone rather than emptying the grid over a typo.

## Nits

- **Deep-link restore.** Fixed — moved to `OnInitializedAsync`, which is what the XML doc already promised. Navigating `/dashboard` → `/audit` creates the component fresh, so the deep link still lands.
- **Untrue collation comment.** Fixed by making it true rather than by editing it: the read now materializes the table and matches in memory, exactly as `ConfigService` does and for the same reason. `RecordLastSyncAsync` does the same, one query for both rows.
- **Empty catalogue cached for five minutes.** Fixed — an empty answer isn't cached. Test: an empty read, then a seed, then a populated read from the same service instance.
- **Doc drift.** Fixed: eight queries / `N × 8`, seven-row table in both `ConfigService` comments.
- **`UsersSync` uses `ToLocalTime()` where `Audit` uses `TimeProvider.LocalTimeZone`** — not changed, deliberately. `Audit` uses `TimeProvider` only to turn the picker's *wall-clock dates* into instants, which has no equivalent here; for **rendering**, `Audit`'s own grid, `Config` and `UserDetail` all use `.ToLocalTime()`. Changing this one line would make `UsersSync` the odd page out. Happy to convert all four in a follow-up if you'd rather standardise the other way.

## On the "for the record" notes

You're right about the hard-stopped link, and it was free to fix: it now points at `?action=user.deactivated` — the action whose details carry `allocationHardStopped` — rather than `key.revoked`, which `DELETE /keys/{userId}` also writes while explicitly *not* hard-stopping. The comment and the docs say why, and point at #208 for the real destination. On spending the card slot on a tripwire while over-budget is a caption: agreed that's worth revisiting, and #208 is where the list view for both lands — I'd rather re-lay-out the row once, when there's somewhere for both numbers to link to, than twice.

## Merge order

Understood, and the coordinator has since sequenced it the other way round: **#204 lands first, then #211, then this.** I'll rebase when told to, and I have your note on the `ConfigService.UpdateAsync` collision — refusal order `404` → `409` system-managed → `409` stale `ExpectedUpdatedDate` → `400` invalid, with the six-argument `SystemConfigEntryResponse`. Noted that it is a build break rather than a text conflict, so it gets a compile before it gets a push.

kolatts:
Rebased onto #204 (`62ddb83`). CI green.

`ConfigService.UpdateAsync` was the only real resolution — a build break rather than a text conflict, as you called it: #204 kept constructing the five-argument `SystemConfigEntryResponse`. Refusal order is now **404 unknown key → 409 system-managed (`EnsureEditable`) → 409 stale `ExpectedUpdatedDate` → 400 invalid value**, with system-managed deliberately outranking staleness — no view of a `LastUserSync*` row would have permitted the write, so "your copy is out of date" is the less useful of the two answers. Both still run before `Normalize`, before the actor is resolved, and before anything can reach the gateway. The response #204 builds from what its conditional UPDATE actually wrote now carries the sixth `IsReadOnly` argument, off the same `SystemManaged` map as the refusal; `ClaimAsync` is taken from main whole.

Three smaller ones: `IConfigService`'s two `ConflictException` docs merged into one ordered list; the config endpoint tests kept both sides; and `api.md`'s Admin table took my `isReadOnly` / corrected audit-filter row plus #204's `expectedUpdatedDate` note. I also corrected #204's two "nine-row table" comments to seven — the merged `All` has seven keys, and that is the same stale-count nit this review raised.

Gates after merging: build 0/0, **1405 tests pass**, `dotnet format --verify-no-changes` clean, docs site builds. Standing by for #211 — expecting the Foundry files to meet its `PATCH …/capacity` as additive members.

Issue #194, opened 2026-09-02
Move the APIM tier-move client into Core so the Functions host can sync tiers itself


kolatts:
Found in review of #187 (Functions wave).

Quota resolution moved to `FoundryGate.Core` in #187 so the Api and the Functions host share one implementation (#119, D-014). One piece did not come with it: `IGatewayTierSync`'s real implementation, `ApimGatewayTierSync`, stays in `FoundryGate.Api/Services/Quota` because it composes `IApimKeyService.MoveToProductAsync`, and the whole `Services/Keys` area is Api-only.

So the Functions host resolves quota but cannot act on the one external consequence of resolving it. Today it registers `WarningGatewayTierSync`, which logs a Warning naming the user and counts the change into the reset's audit row — honest, but it means a scheduled reset can leave SQL and the gateway disagreeing until the Api next touches that user (see #193 for the case that makes this likely).

## Ask

Move the minimum needed for a tier change into Core:

- `IApimManagementClient` and its ARM implementation (`ArmApimManagementClient`) — already options-driven off `GatewayOptions`, which is now in `Core.Configuration`, so this is mostly a namespace move.
- Whatever slice of `IApimKeyService` `MoveToProductAsync` actually needs. It reads the subscription's current product and re-scopes it (`PUT .../subscriptions/{sid}` with `scope=/products/{tier}`), and adds a `key.tier-changed` audit row via `IAuditWriter` — none of which is Api-specific. The parts that *are* Api-specific (key protection, reveal, the current-user paths) should stay behind.
- Then register `ApimGatewayTierSync` in the Functions host when `GatewayOptions.IsApimConfigured`, mirroring the Api's conditional registration, and delete `WarningGatewayTierSync`.

## Watch out for

- **Least privilege.** The Functions identity currently holds Log Analytics Reader, Key Vault Secrets User and the three Storage data roles. Moving tier syncs here means it also needs **API Management Service Contributor** on the APIM instance (`infra/modules/control-plane-rbac.bicep`) — a real widening of that identity's blast radius, and worth a moment's thought against the alternative of #193 alone.
[...]

PR #211, opened 2026-09-02
feat: Core APIM tier sync, gateway alias map, key-reveal hardening, Foundry capacity PATCH

kolatts:
Five backend issues in one sweep. Each is one commit; nothing here touches `.github/`, and the only infra edits are the alias env-var block, one output, and one role assignment (see **Concurrency** below).

---

## `Closes #194` — move the APIM management client into Core so the Functions host syncs tiers itself

`IApimManagementClient`, `ArmApimManagementClient` and `UnconfiguredApimManagementClient` move from `FoundryGate.Api/Services/Keys` to `FoundryGate.Core/Gateway`. Mostly a namespace move; the one substantive change is `StatusCodes.Status404NotFound` → a documented literal, because Core carries no ASP.NET Core dependency and that is what lets the isolated worker use it.

`ApimGatewayTierSync` moves to `FoundryGate.Core/Quota` and composes the management client directly instead of `IApimKeyService.MoveToProductAsync`, **which is deleted**. One re-scope implementation, not two — the issue's "whatever slice of `IApimKeyService` `MoveToProductAsync` actually needs" turned out to be all of it, and leaving a second copy behind in the Api would have been the drift this issue exists to prevent. Key protection stays in the Api, as the issue asks; Core does not grow `IKeyProtector`.

**The one genuine host difference is who the audit row belongs to**, so that became a seam: `IGatewayTierSyncActor`. The Api registers `CurrentUserGatewayTierSyncActor` (the admin whose request moved the quota — and it resolves via `GetRequiredUserAsync`, so the 403-for-an-unprovisioned-caller refusal still lands *before* ARM is touched); the Functions host registers `SystemGatewayTierSyncActor`, so a reset's `key.tier-changed` row is system-attributed like its own `quota.monthly-reset` row. `WarningGatewayTierSync` is deleted, and the Functions host now mirrors the Api exactly: the real sync when `GatewayOptions.IsApimConfigured`, `NullGatewayTierSync` otherwise (whose "no gateway is configured" line is then simply true).

**Decisions for the reviewer:**

- **Privilege widening, stated plainly.** The Functions identity gains **API Management Service Contributor** on the APIM instance (one resource in `control-plane-rbac.bicep`). Azure has no narrower built-in for "re-scope a subscription", so the same role also permits creating, deleting and regenerating developer subscriptions and editing named values. Two things bound it rather than remove it: the Functions host carries no key protection (it cannot read or write a developer's key), and every move writes an audit row naming the subscription. `reference/infrastructure.md` gets a section explaining the trade **and how to opt out** — delete the resource, and the reset then fails loudly on the first tier change instead of moving it silently.
- **A mid-reset ARM failure aborts the run** (the issue asks for a decision). The tier sync is called before the reset saves, so an ARM refusal propagates and the whole run rolls back. The alternative — continue and report — leaves the dashboard promising 20M while the gateway keeps 403ing at 5M, with nothing rolled back. There is a test for it.
- **#132's open question is unchanged.** "Does a re-scope preserve the key?" is still unproven against real APIM; the `reference/api.md` caution block and the #132 checklist stand. Nothing here makes the assumption more or less true — it just means the Functions host now relies on it too.

## `Closes #153` — expose the gateway alias map to the control plane

`infra/main.bicep`'s `productModelAliases` reaches both hosts as `Gateway__ModelAliases__{i}__Tier` / `__Alias` / `__DeploymentName` / `__Provider`, flattened from the same object the gateway module turns into policy. One source, two consumers. Plus a `modelAliasRows` output.

**The tier is a fourth field the brief did not name, and it is the load-bearing one.** The bicep map genuinely differs per product — `unlimited` lists `opus`, `standard` does not — and the alias map *is* the allowlist (#86). A flat union would tell a Standard developer they can use `opus` and let them find the `403 model_not_permitted` at their first request, which is worse than the empty list this replaces. So each row carries its tier and `GET /users/me` filters `cliConfig.modelAliases` to the caller's own. The issue asks for exactly this ("filtered to what the caller's tier product actually allows *if that ever differs per product*") — it does.

`GatewayOptions.ModelAliases` validates item-level at startup like `Tiers` does (`ValidateRecursively()` does not walk list items): an unknown tier id, a blank member, or the same tier+alias twice all fail fast. A tier typo is otherwise silent — the row just never matches, and developers are told the gateway has no models. An unconfigured map stays valid and returns `[]`, so a fork on an older deploy behaves exactly as before.

## `Closes #181` — make the `/keys/me` rate limits configurable

`Security:RateLimits:Reveal` / `:Rotate` carry a permit count and a window each, defaulting to today's 5 and 3 per minute, validated fail-fast. `KeysRateLimitEndpointTests` now reads the configured values instead of mirroring them in two `private const`s, and a new `KeysRateLimitWindowTests` runs its own host with a one-second window to prove the window really turns over — the half the #184 review comment asked for and constants could not give. (It polls with a 20s ceiling rather than sleeping a window, so a slow agent makes it slower rather than red. It is the only wall-clock test in the suite, and it has to be: the built-in limiters own their replenishment timer and take no `TimeProvider`.)

`ApiTestFactory` is unsealed and gains a `ConfigureSettings` hook so a derived factory can change a host setting through `UseSetting` — never `ConfigureAppConfiguration`, per CONVENTIONS.md.

## `Closes #180` — anomaly signal on repeated key reveals

`Security:RevealAnomaly` (threshold 10, window 60 minutes). The reveal that crosses the line logs a Warning and adds a `key.reveal-anomaly` row, committing with the `key.revealed` row that triggered it.

Two judgement calls worth reviewing: it counts **per key** (`TargetId`) rather than per actor — a credential being drained is the thing worth noticing, and that stays true whoever holds the token; today's only reveal route is `/keys/me` so the two coincide, and the actor is on the row anyway. And it fires **once per window** — an anomaly row per reveal buries the reveals it is trying to point at. A drain that outlives the window is reported again, which is correct: it is still happening.

Implemented in `ApimKeyService.RevealAsync` (the issue's "cheapest" option) rather than a Functions timer: one indexed `COUNT` on a route that already writes a row, and the signal arrives with the event rather than up to a day later. A cross-day sweep is still the version the issue's second option describes; nothing here forecloses it.

## `Closes #130` — `PATCH /foundry/deployments/{account}/{name}/capacity`

**PATCH is confirmed, and the confirmation is stronger than "the SDK method is named Update".** ARM's `Deployments_Update` is `PATCH .../deployments/{name}` (published REST reference), its request body schema is `{ sku, tags }` — no `model`, no `modelProviderData` — and `Azure.ResourceManager.CognitiveServices` 1.5.2's `UpdateAsync(WaitUntil, PatchResourceTagsAndSku, ct)` maps to it (the assembly references `RequestMethod.Patch` exactly once, for that request builder). So the call *structurally cannot* re-send the model. That is what makes it a different animal from the `CreateOrUpdate` PUT the create-once rule is about.

**Anthropic is still refused with a 400**, per the brief's "default to refusing unless proven". The argument above is structural, not observed: nobody has run a capacity PATCH against a live Claude deployment, and E-007's lesson is that Claude deployments punish assumptions however sound. #205 is the live check that lifts it, and the code change is deleting one guard. The refusal message says all of this and names the issue.

Two smaller decisions: asking for the capacity a deployment already has is a **no-op** — no ARM call, no audit row, because a PATCH to the same value changes nothing and a trail claiming otherwise would be worse than silence. And the audit action is `foundry.deployment.capacity-changed` with `{ before, after }`, not the issue body's `foundry.deployment.resized`: it matches `key.tier-changed`'s shape and says *what* changed, since the deployment itself is untouched. Same for `UpdateCapacityAsync` over `ResizeDeploymentAsync`. Flag it if you'd rather have the issue's names.

---

## Verification

- `dotnet build FoundryGate.sln -c Release` — **0 warnings, 0 errors** (`TreatWarningsAsErrors`, and IDE0005/IDE0044/IDE0055/CA1062 are all errors here, so this is a real gate).
- `FoundryGate.Tests.Predeployment` — **1124 passed, 0 failed** (was 1090 on `main`; +34).
  - One run hit the known `Web/MePageTests` flake (**#195**), `Rotating_asks_for_confirmation_before_it_touches_the_api`; it passed on re-run and on every run since.
  - One run also hit `HealthEndpointTests.HealthReady_returns_503_...`, which turned out to be a real latent test bug rather than a flake: it sets its connection string with `ConfigureAppConfiguration`, which never reaches the bound options, so it silently connects to the developer's docker SQL and reports healthy. Filed as **#209** with the fix (`UseSetting`). Not touched here — it belongs to whoever picks up #209, and CI has no docker SQL so it is green there either way.
- `az bicep build --file infra/main.bicep` and `az bicep lint` — clean (only the pre-existing `BCP081` notes about the preview `Microsoft.CognitiveServices` api-version).
- `npm ci && npm run build` in `docs-site/` — 15 pages built.
- No live Azure was touched.

## Docs

- `reference/infrastructure.md` — the Functions APIM role row, plus a "Why the Functions identity can write to APIM" section covering the trade and the opt-out.
- `reference/configuration.mdx` — `Gateway__ModelAliases__*`, `Security__RateLimits__*`, `Security__RevealAnomaly__*`.
- `reference/api.md` — `GET /users/me`'s alias list (the "always empty" paragraph is gone), the rate limits as configuration, the anomaly signal, and the capacity endpoint with the PATCH-vs-PUT reasoning.

## Issues filed

- **#205** — live-validate the capacity PATCH (OpenAI, then Claude), record the E-00x finding, then lift the Anthropic refusal.
- **#209** — `HealthEndpointTests` uses `ConfigureAppConfiguration`, so it passes or fails depending on whether docker SQL is running.

## Concurrency

- **Merge #204 first, then this.** #204 adds `GatewayInfraBindingTests`, which fails when a `GatewayOptions` property has no matching `Gateway__*` env var in the bicep. `ModelAliases` lands *with* its env vars so the substance holds, but the test's shape may need extending for an indexed list-of-**objects** (`Gateway__ModelAliases__{i}__Tier`, four keys per row) rather than the indexed list-of-strings `FoundryAccountNames` is. I'll rebase and adjust it once #204 lands.
- Expect adjacent-line conflicts with #204 in `GatewayOptions.cs`, `configuration.mdx` (its per-host truth table — my three rows go in) and `reference/api.md`.
- Infra edits here are deliberately narrow, to stay clear of the infra/CI agent on #137/#157/#134/#155: the alias env-var block + `productModelAliases` param in `control-plane.bicep`, one `modelAliasRows` output, and one role-assignment resource in `control-plane-rbac.bicep`.
- No overlap with frontend PR #176.


kolatts:
**REQUEST CHANGES**

The engineering here is good — `IGatewayTierSyncActor` is exactly one interface for exactly one host difference, which is a lot more disciplined than the two-implementations-of-everything version this could have been. What's blocking is that the reset's *story* didn't move with it: four separate places still tell a reader the monthly reset never touches APIM, and the "a mid-reset ARM failure aborts the run" trade-off you wrote up is describing a rollback that doesn't exist. That's one Major, and the rest are small.

I re-ran everything and it holds: build 0/0, `Tests.Predeployment` **1124/1124**, `dotnet format` clean, `az bicep build` + `lint` clean (only the pre-existing `BCP081` notes), docs-site 15 pages.

**On the PATCH claim — I went further than the assembly grep, and it holds.** I put a recording `HttpPipelineTransport` behind `ArmClient` and called `CognitiveServicesAccountDeploymentResource.UpdateAsync(WaitUntil.Started, PatchResourceTagsAndSku, ct)` for real:

```
PATCH https://management.azure.com/subscriptions/.../accounts/a/deployments/d?api-version=2025-06-01
```

One request, `PATCH`, zero `PUT`s. That's observed rather than argued, so consider #205's step 1 half-answered: what's left to prove live is ARM's *behaviour* on a Claude deployment, not the verb. Keeping the Anthropic guard until then is still right — E-007 earned that caution.

---

**Issues**

- **Major** `src/FoundryGate.Core/Quota/QuotaResetService.cs` line ~37 (with `IQuotaResetService.cs`, `MonthlyResetJob.cs` line 102, `docs-site/src/content/docs/architecture/overview.mdx` line 233) — **the reset is no longer all-or-nothing, and five places still say it is.** `ResolveManyAsync` calls `SyncAsync` per user inside the loop and the single `SaveChangesAsync` runs after the whole loop, so a failure on user *N* leaves users *1..N-1* already re-scoped at APIM with their `key.tier-changed` rows discarded with the rollback — SQL says the old tier, the gateway enforces the new one, and nothing recorded it. Worse, the next run's re-sync early-returns at the `current.ProductId == productId` check *before* the audit row is added, so those moves are permanently unaudited. And an `ApimSubscriptionNotFoundException` for one user whose subscription was deleted out-of-band is a `ConflictException` that now deterministically aborts the whole reset, on every retry, for everybody else. CONVENTIONS.md is explicit about this shape: *"the residual 'save failed after the external system accepted' orphan is logged at Error with the change's full identity and rethrown"* — that needs to happen for the users already moved. Then correct the four claims: `QuotaResetService`'s *"the reset is all-or-nothing exactly as the interface promises"*; `IQuotaResetService`'s *"It never touches APIM either ... no subscription state to change"*; `QuotaResetOutcome.TierSyncCount`'s *"zero for a scheduled reset"* (the whole premise of #194 is that it isn't); and `MonthlyResetJob`'s Information log, which still tells operators *"so no APIM state was touched"* on every run — you updated the XML doc directly above it and missed the line itself. It should report `TierChangeCount`. Fan-out is the other half — #199 already has the Api side of "one config edit, N serialized ARM round trips"; this makes the timer do it too, so it probably wants to be named on that issue.

- **Minor** `src/FoundryGate.Api/Services/Users/UserService.cs` lines 382–395 — `BuildCliConfig` now carries **two** `<summary>` blocks, and the first one still says *"The alias list is empty because the alias map lives only in bicep today; #153 exposes it to the control plane"* — in the PR that *is* #153. Looks like an incomplete edit; delete the old one.

- **Minor** `docs-site/src/content/docs/architecture/overview.mdx` line 233 and `docs-site/src/components/SystemDiagram.astro` lines 20–21, 155–156, 210 — the middle rung of the progressive ladder now states the opposite of what the code does. The reset pipeline step reads *"no APIM counter manipulation needed, and no subscription state is touched"*; the diagram's `ariaLabel` and its control-plane bridge arrow both attribute the Management API to the Api alone (*"create under a tier product, regenerate both keys, move tier, delete"*), and the `FoundryGate.Functions` node has no APIM edge at all. CONVENTIONS.md: *"Every wave that changes the architecture updates the docs-site diagram components (`docs-site/src/components/*.astro`) and the architecture pages — the public site must always show the current system."* The `reference/*` pages are genuinely excellent; it's this level that's missing, and CLAUDE.md invariant 2 asks for all of them.

- **Minor** `src/FoundryGate.Api/Services/Keys/ApimKeyService.cs` (the anomaly queries) — two `AuditLogs` queries per reveal filtering `Action` + `TargetType` + `TargetId` + `OccurredDate`, but the only nonclustered index is the single-column, non-covering `IX_AuditLogs_OccurredDate`, so each one is a range seek over *every* audit row written in the last hour plus a key lookup per row. `RevealAnomalyOptions`' justification for having no `Enabled` flag — *"the check costs one indexed `COUNT`"* — is wrong on both counts. An `(Action, TargetId, OccurredDate)` index (mirrored into `dbo/Tables/AuditLogs.sql`, per CONVENTIONS.md) makes the sentence true; the two queries could also be one grouped round trip.

- **Minor** `src/FoundryGate.Api/Services/Foundry/ArmFoundryManagementClient.cs` lines 158–178 — with `WaitUntil.Started` the operation reports `HasValue=False`/`HasCompleted=False` even on a synchronous 200 (same transport probe as above), so `if (operation.HasValue)` never fires and the fallback `GetDeploymentAsync(..., cancellationToken)` always runs — on the caller's token, after ARM accepted the PATCH. A disconnect in that gap throws before `AuditAfterCommitAsync`, which is the accepted-but-unaudited case the method's own commit-point comment claims to prevent. `CreateDeploymentAsync` has the identical shape, so this is probably one issue covering both rather than a change here.

- **Minor** `docs-site/src/content/docs/reference/infrastructure.md` — the new ``[`IKeyProtector`](/reference/configuration/)`` link is missing the `/foundry-gate` base (`astro.config.mjs` line 6), so it 404s on the published site; every other link in that file is prefixed. There's no link validator in `docs-site/package.json`, so the build won't catch it.

- **Nit** `src/FoundryGate.Tests.Predeployment/Api/Endpoints/KeysRateLimitWindowTests.cs` — with `WindowSeconds = 1`, the assertion that the *second* reveal is a 429 races the fixed window's boundary, and the first reveal is the cold path through decrypt, audit and the two brand-new EF queries whose compilation happens right there. A 2–3s window costs nothing (the replenishment half polls anyway) and takes the race off the table.

- **Nit** `CONVENTIONS.md` lines 26–28 — the contract's own worked example of the Core/host split is the arrangement this PR deletes: *"the Api registers `ApimGatewayTierSync` (it composes the APIM key service) against Core's `IGatewayTierSync`, the Functions host registers `NullGatewayTierSync`."* All three clauses are now false, and CLAUDE.md tells the next implementer to read this file first.

- **Nit** #132's checklist still has *"`IApimKeyService.MoveToProductAsync` end to end once #118 lands"*, and this PR deletes that method. Worth retargeting it at `IGatewayTierSync`/`ApimGatewayTierSync` while closing #194 — and it's now a two-host check, since the Functions path has never been exercised against real APIM at all.

---

**Things I went looking for and didn't find, for the record**

- Actor-before-ARM survives the move: `ApimGatewayTierSync` resolves `IGatewayTierSyncActor` ahead of `GetSubscriptionAsync`, and `An_actor_the_host_refuses_stops_the_move_before_the_gateway_sees_it` pins it.
- No double `key.tier-changed` — `MoveToProductAsync` was the only other writer, it's gone, and `Assert.Single` covers both attributions. `RevokeAsync`/`RevokeAsSystemAsync`/rotation compensation untouched.
- Dropping `RequireKey` for a silent skip is safe: `QuotaResolutionService` already guards on `!string.IsNullOrEmpty(user.ApimSubscriptionId)`, so the Core guard is belt-and-braces.
- Core stays ASP.NET-free — the csproj pulls only `Azure.Core`, `Azure.ResourceManager.ApiManagement`, two `Microsoft.Extensions.*.Abstractions` and `Imagile.Framework.Configuration`. The `404` literal is the right call and says why.
- The reveal-anomaly `+1` is correct (the pending row isn't visible to the query), and the once-per-window guard behaves.
- `SecurityOptions` and friends live in the Api's `AppSettings.cs`, which is what makes `ValidateRecursively()` actually walk them — that's the trap CONVENTIONS.md warns about and you didn't fall in it.
- The bicep flattening produces exactly the four documented keys per row for all ten rows; `items({})` → `range(0,0)` keeps an unconfigured map valid; both hosts get a full app-settings replace on deploy, so a shrinking map can't strand a half-row. The Functions RBAC assignment is scoped to the same `apim` resource as the Api's.
- `getting-started/cli-setup.mdx` already told developers to read the alias list off `/me` — as of this PR that sentence is finally true.

---

**Merge order — and the #204 collision, precisely**

**#204 → #211 → #210.**

You called the #204 ordering right; here's the exact failure so nobody rediscovers it. `GatewayInfraBindingTests`' matcher is:

```
name:\s*'(?<name>Gateway__[A-Za-z0-9_]+(?:__\$\{i\})?)'
```

`[A-Za-z0-9_]+` can't consume `${i}`, and the optional group must be followed immediately by the closing quote — so it matches **none** of `Gateway__ModelAliases__${i}__Tier|Alias|DeploymentName|Provider`. I ran it against your `control-plane.bicep`: it finds the same 8 names it finds on `main`. Result: `Every_GatewayOptions_member_is_either_set_by_infra_or_listed_as_deliberately_absent` fails with *"GatewayOptions member(s) ModelAliases are neither set by infra..."*. Three things on the rebase:

1. Widen the regex to allow a segment *after* the index — e.g. `Gateway__[A-Za-z0-9_]+(?:__\$\{i\}(?:__[A-Za-z0-9_]+)?)?` — and make the normalizer strip both the `${i}` and the trailing member so all four names collapse to `ModelAliases`.
2. `Every_Gateway_variable_infra_sets_binds_to_a_GatewayOptions_member` builds sentinel values from the variable name, which throws at bind time for `Gateway:ModelAliases:0:Provider` — `"Gateway__ModelAliases__${i}__Provider"` doesn't parse as a `ModelProviderType`. The sentinel generator needs an enum case. This is the bit that'll eat an hour if you don't see it coming.
3. `Both_hosts_get_the_same_gateway_configuration` only asserts `sharedAppConfig` + `foundryAccountConfig`; add `modelAliasConfig`, or a later change that wires the aliases to one host slips through.

Worth noting your own `GatewayOptionsModelAliasesTests.Infra_productModelAliases_names_only_tiers_and_providers_the_control_plane_can_bind` already guards `main.bicep`'s parameter — the gap #204 closes is the *flattening* in `control-plane.bicep`, which nothing checks today.

**#210 after this one.** It touches `IFoundryManagementClient`, `ArmFoundryManagementClient`, `FakeFoundryManagementClient`, `FoundryController` and `reference/api.md`'s Foundry table — the same files as your #130 commit, all additive and adjacent-line. Your Foundry surface is four members; #210's is a catalogue endpoint plus the UI reading it, so it's cheaper for #210 to rebase onto you than the reverse.

The privilege-widening writeup is the best part of this PR, by the way — "here is the blast radius, here is what bounds it, here is how to opt out and what breaks when you do" is the section most people skip.

kolatts:
Fix pass done — everything addressed. Rebased onto **#204 → #198 → #210**, all three now in.

Thanks for the transport probe on the PATCH. "One request, `PATCH`, zero `PUT`s" is a better answer than my assembly grep, and it changed something downstream: it also told me `HasValue` is false under `WaitUntil.Started`, which is the Minor below.

---

## Major — the reset is no longer all-or-nothing, and five places said it was

You were right that the trade-off I wrote up described a rollback that doesn't exist. Fixed properly rather than by softening the prose.

Resolution now takes a `GatewayTierSyncMode`. The reset passes `Deferred`, so the bulk pass reaches no gateway and only *reports* which developers need a move; the reset then drives each move itself and commits it immediately on `CommitToken.For(true, ct)` with the `key.tier-changed` row beside it. Developers whose tier is not moving reach nothing external, so they still share the final save with the run's audit row — the batch stays a batch for the common case.

Three details worth flagging because they were decisions, not mechanics:

**Moving developers' rows are held out of the change tracker until their own move lands.** `SaveChangesAsync` commits the whole tracker, so the first successful move would otherwise also commit the rows of developers whose move hasn't been attempted — each of which would claim a tier the gateway isn't enforcing if theirs later failed. `Hold` detaches, `Restore` re-attaches marking exactly the six columns resolution and `Stamp` write (not `Update()`/`State = Modified`, which would mark `TokensUsed` dirty and write back a value read minutes earlier).

**A refused developer gets no allocation row for the period at all.** I tried writing one recording the old tier and it's incoherent — you'd have to write the old cap and level too, i.e. deliberately persist a stale resolution. Leaving them out is the honest outcome: their previous period's row still matches the gateway, `UsersResetCount` excludes them, and their next `/quota/allocations/me` (or a manual reset) retries. Documented on the interface, because it's a contract fact rather than an implementation detail.

**A failed *save* still rethrows.** Continue-on-failure is for a refused move; a save failing after APIM accepted is the residual orphan, so it logs at Error with user id, email, subscription and target tier, then propagates — `LogOrphan`, matching `AuditAfterCommitAsync`.

**The idempotence hole.** `SyncAsync` now takes the previously recorded tier, which is what separates the two "already on the target product" cases: database and gateway agreeing (real no-op, silent) versus the database still recording something else — an earlier move whose save didn't land. The second writes its row flagged `alreadyInPlace: true` and logs a Warning. Without that, exactly as you said, the caller writes the new tier onto the row and the move is recorded nowhere.

**Tests:** failure on developer N leaves 1..N-1 moved, committed and audited (`AfterMutation` breaks the gateway after the second move); an out-of-band portal deletion no longer aborts the run; and the `alreadyInPlace` row is pinned.

All four stale claims corrected, plus `MonthlyResetJob`'s Information log, which was still telling operators on every run that no APIM state was touched — it now reports tier changes *and* refusals. Fan-out named on #199.

## Minors

- **Double `<summary>`** — deleted; the surviving one documents the tier filter, with the "alias map is the allowlist" reasoning moved to `<remarks>`.
- **`architecture/overview.mdx` + `SystemDiagram.astro`** — the reset step now describes moving tiers and what a refusal does; the diagram's `ariaLabel`, bridge arrow and Functions node all say the Functions host reaches the Management API for one operation only.
- **Anomaly queries** — one grouped count instead of two, plus a composite `(Action, TargetType, TargetId, OccurredDate)` index on `AuditLog` mirrored into `dbo/Tables/AuditLogs.sql`. `RevealAnomalyOptions`' justification now describes what it actually does.
- **`operation.HasValue`** — fixed in both `UpdateCapacityAsync` and `CreateDeploymentAsync` here rather than deferring: it's two lines and it's the commit-point rule, which this PR is already about. Both read back on `CancellationToken.None`, and the comment records your finding so the dead-looking branch isn't mistaken for the live path.
- **`infrastructure.md` link** — `/foundry-gate` base added.

## Nits

- Rate-limit window 1s → **3s**. Your reasoning was the part I'd missed: the first reveal is the cold path through decrypt, audit and the two new EF queries.
- **CONVENTIONS.md §Solution structure** rewritten to the arrangement that actually landed, with the general rule stated — prefer a narrow seam for the difference over a second implementation of the whole thing.
- **#132's checklist** retargeted onto `IGatewayTierSync` as a two-host check, with items for the never-exercised Functions path, the continue-on-refusal behaviour and the `alreadyInPlace` recovery.

## The #204 collision

Your three points, in order:

1. Regex — already widened by #204's own fix pass; nothing needed.
2. **The sentinel generator.** You called it and it would have eaten an hour. Worse than a bind-time throw, actually: `Convert.ChangeType` cannot do enums *at all*, so the fix is two-sided — `Generated` rotates through `Enum.GetNames` for enum types, and a new `Expected` helper parses by name. Had it only produced an unparseable string, the binder would have left `Provider` at its default and the test would have passed while proving nothing.
3. **`Both_hosts_get_the_same_gateway_configuration`** — `modelAliasConfig` asserted, with a comment saying it's every block the hosts are handed, not the two that existed when it was written.

Also rewrote the `DefaultMonthlyTokenQuota` prose in `infrastructure.md` and on `IQuotaResetService`: #204 makes that edit re-resolve its users, so what actually reaches the reset now is a developer with no earlier allocation plus anything a re-resolution missed.

## #198 and #210

#198 merged clean — verified rather than assumed: the alias env-var block, `modelAliasRows` output and Functions APIM RBAC assignment all sit alongside their knobs, `staticWebAppResourceId` and `swa-preview-role`. #210 conflicted only where we'd both appended to `FoundryEndpointTests`; capacity and catalogue cases now sit side by side. No Web route-tripwire row is owed — the capacity PATCH has no Blazor client method yet (#62's grid editor).

## Verification

Build **0/0**; `Tests.Predeployment` **1444/1444**; `dotnet format` clean; `az bicep build` + `lint` clean (pre-existing BCP081 only) and `build-params` clean for `dev` and `prod` with dummy `FG_*`; docs-site 15 pages. CI green.

One flake to log, a different one from #195: `Web/DashboardPageTests.Stops_refreshing_once_the_page_is_gone` failed once mid-pass and passed on re-run and every run since — same timer/disposal family. Say the word and I'll file it, or fold it into #195 if you'd rather keep the Blazor-timing flakes together.

Issue #213, opened 2026-09-02
Store the per-model, prompt/completion token split so cost estimates can stop being blended


kolatts:
> Parent: #224 — Engineering debt

Split out of #177, which landed the rate card and the estimated-cost fields but could only apply a
single blended rate.

## Why the estimate is blended today

`QuotaAllocation.TokensUsed` is **one total per user per period**. It carries:

- no prompt/completion split, so `inputPerMillion` and `outputPerMillion` cannot be applied
  separately — `RateCard.BlendedRatePerMillion` averages the two and says so;
- no per-model split, so a rate card's per-model entries (`claude-opus`, `gpt-5-codex`, …) are
  validated and stored but **never read**. Only the `*` fallback entry does anything.

Both caveats are documented on `RateCard`, `DashboardSummaryResponse.EstimatedCostThisPeriod` and
`QuotaAllocationResponse.EstimatedCost`, and every rendered figure is labelled an estimate.

## The data is already at the source

`Kql/UsageBySubscription.kql` already selects `PromptTokens` and `CompletionTokens`, and the
`ApiManagementGatewayLlmLog` rows carry `ModelName` — `UsageSyncJob` sees the model on every row it
aggregates. The reconciliation query is a `by ApimSubscriptionId, ModelName` away from producing
exactly what a per-model, per-direction cost needs. Nothing new has to be collected; it is thrown
away on the way into the database.

## The real design question

Where the split is stored. `QuotaAllocation` is one row per user per period and is the enforcement
mirror — widening it to (user × period × model × direction) changes what that table means, and the
quota reads that project from it would all have to aggregate. The likely shape is a sibling
`QuotaAllocationUsage` (or `TokenUsageByModel`) table keyed on the allocation plus the model, written
by the sync job, with `QuotaAllocation.TokensUsed` staying the total the gateway enforcement mirror
needs.

## Scope

- Decide and build the storage for the split.
- `UsageSyncJob` / `UsageBySubscription.kql` write it.
- `RateCard` grows a per-model, per-direction estimate; `BlendedRatePerMillion` becomes the fallback
  for models the card does not name, rather than the only path.
- The "estimate" labelling stays: interrupted streams still undercount (#84) and cache-token
  weighting at the gateway is still unverified (#88). Those are separate reasons the number is a
  floor, and this issue does not fix either.

## Not in scope

Enforcement. Cost stays a reporting figure — the gateway enforces tokens, never dollars.

Issue #88, opened 2026-09-01
Claude on Foundry: complete end-to-end gateway validation (blocked by Anthropic deployment provisioning failures)


kolatts:
> Parent: #81 — Claude on Foundry end to end

## What happened

During the 2026-09-01 live validation pass (PR #87, full timeline in `fable-refactor-log.md` E-007), Claude model deployment creation broke **subscription-wide** on Imagile Paid:

- The **first** `claude-haiku-4-5` create on a fresh account succeeded (reached `deploymentState: Running`).
- Every create after the first **delete** failed — async `InternalServerError` after ~5 min of `Creating`, then eventually synchronous 500s. This includes fresh deployment names, fresh accounts, and the never-churned `imagile-foundry` account. OpenAI deployments provisioned instantly throughout — the failure is Anthropic-specific.
- Strongest hypothesis: the per-subscription Anthropic Marketplace (CCU) agreement wedged when the first deployment was deleted while its purchase was still settling. No SaaS resource is visible to repair via ARM; no error detail is surfaced in the Activity Log.

## What was still validated

- Claude Code **wire compatibility is proven**: it sends `x-api-key` to `/v1/messages?beta=true` (captured), passes gateway auth and the full policy chain, and fails only at the missing backend deployment.
- Everything else in the gateway architecture was validated on the OpenAI path (monthly-quota 403, TPM 429s, isolation, real Codex session) — see PR #87.

## What this issue tracks

1. Wait ≥24h (or open an Azure support ticket referencing the InternalServerError tracking IDs in the decision log) and retry a single, careful `claude-haiku-4-5` create.
2. When it succeeds, run the remaining validation matrix: T1 (Messages through the gateway), T4-Anthropic (TPM on Anthropic schema), T8 (pool failover eus2→swc under real 429s), T9 (prompt-cache token counting vs gateway estimates), T10 full (real Claude Code session end-to-end, incl. the new managed-identity backend auth on the `/anthropic` path).
3. Record results in `plans/24-apim-genai-gateway.md` verification checklist and update `architecture/feasibility.mdx` if cache-counting diverges.

## Operational rules already adopted from this incident (in main via PR #87)

- Anthropic deployments are **create-once** under ARM: never re-PUT, never delete/recreate in a loop (`infra/modules/foundry.bicep`, `CLAUDE.md`).
- `createModelDeployments = false` on all re-runs; model lifecycle belongs to the control plane (#60/#64).

[...]

PR #215, opened 2026-09-02
fix(tests): pin the readiness probe, harden the flaky bUnit tests, and make the dashboard's counts reachable and priced

### LINKED ISSUE #177: Per-developer cost attribution: rate card in SystemConfiguration + cost on the dashboard

### LINKED ISSUE #197: HealthEndpointTests readiness test passes only when no SQL Server is reachable locally

### LINKED ISSUE #203: Flaky: MePageTests.Rotating_asks_for_confirmation_before_it_touches_the_api fails intermittently in a full-suite run

### LINKED ISSUE #208: The dashboard can count hard-stopped and over-budget users but nothing can list them

kolatts:
Final sweep C: #197, #203, #208, #177. One commit per issue, on top of `origin/main` including PR #211.

## #197 — the readiness test only passed when no SQL Server was reachable

`HealthEndpointTests` overrode the connection string through `ConfigureAppConfiguration`, whose sources minimal hosting appends only *after* `Program.cs` has bound `AppSettings`. The override never reached the DbContext, so the asserted `503` was really produced by whatever `appsettings.local.json` pointed at.

- Override via `IWebHostBuilder.UseSetting` (CONVENTIONS.md §Integration tests).
- New `HealthReady_returns_200_when_the_database_is_reachable`, on `ApiTestFactory`'s SQLite connection. Without it, a check that always failed would pass the `503` test too — which is exactly the hole this had.

**Reproduced before fixing.** With the docker SQL container up *and* the `FoundryGate` database created, the old `ConfigureAppConfiguration` form fails with the issue's exact signature (`Expected: ServiceUnavailable / Actual: OK`). The new form passes 4/4 with the container both up and down.

## #203 — the flaky bUnit tests

Two real causes, both fixed at the source rather than per test:

1. **A one-second wait budget.** bUnit's `DefaultWaitTimeout` is 1 s; `/users`, `/groups` and `/quota` debounce input for 300 ms before the request the assertion is waiting on even starts. `WebTestContext` now sets it to 10 s for the whole Web suite — the number only decides how long a *failing* wait takes to give up.
2. **A fake that was not thread-safe.** Pages call the API from the renderer's thread (a debounce timer, `/dashboard`'s `PeriodicTimer`) while `WaitForAssertion` reads the recorded calls from the test's. A `List<T>` appended to mid-enumeration throws "collection was modified" — which `WaitForAssertion` swallows and retries, making it indistinguishable from a condition that never became true. Capture lists are now `RecordedCalls<T>` (locked, snapshot reads); the counters are `ConcurrentDictionary`.

Per test, on top of that:

- The rotate tests wait for the dialog's own render (MudDialog binds through the `DialogProvider` beside the page) instead of `Find`-ing buttons that may not exist yet.
- The groups search test waits for the second `ServerData` call before asserting on its contents, so a timeout says "never re-queried" rather than "nothing matched".
- `Stops_refreshing_once_the_page_is_gone` (the sighting from #211's branch) settles on **five** consecutive identical samples, not two — a tick already inside `InvokeAsync` when `Dispose` ran still lands, and 50 ms of quiet either side of it looked like a stop.
- `A_reply_that_lands_after_disposal…` waits on a new `CompletedCallCount` instead of `Task.Delay(50)`; asserting the badge before the reply landed proved nothing.

**`WebTestContext` audited for cross-class state, as the issue asked: there is none.** `Api`, `Time`, `JSInterop`, `Authorization` and the service collection are all per-instance, and the only static is `WebTestData.Tiers` (immutable records). Recorded in the class remarks so the next flake hunt starts somewhere else.

**5/5 green** on the full Predeployment suite, twice — after that commit (1445 tests) and again at the tip (1515 tests).

## #208 — the dashboard could count hard-stopped and over-budget users but nothing could list them

- `GET /quota/allocations` takes a `[FromQuery] QuotaAllocationQuery`: `isHardStopped`, `isOverBudget`, `tier`, `search`, `isActive`. All optional; an empty query string is the list this endpoint has always served.
- `isOverBudget` uses the same `>=` `DashboardService` counts with. `tier` is canonicalized against `Gateway:Tiers` so `?tier=Standard` matches on every provider, while an id no tier claims is matched literally — a legacy `tierProductId` is exactly what someone would be hunting for.
- New `Pages/QuotaAllocations.razor` at `/quota`, in the Admin nav: server-paged `MudDataGrid`, filter chips, tier picker and search, all mirrored into the query string so a filtered view is a shareable link.
- Both dashboard cards now link into it with the filter applied, replacing the audit-log link.

Tests: seven endpoint tests over the filters (including the exactly-at-cap boundary and the tier's casing), thirteen bUnit tests over the page (asserting the query it sent, not just what it rendered), two new route-tripwire rows.

## #177 — per-developer cost attribution

Azure cannot answer this: Claude bills as a single aggregate Marketplace CCU meter, so Cost Management cannot break spend down per developer. It is computed from the fork's own prices.

- New `RateCard` `SystemConfiguration` key — one JSON array of `{ modelPrefix, inputPerMillion, outputPerMillion }`, seeded `[]`, validated by `SystemConfigValidator` and stored re-serialized. `/config` gives it a multi-line box.
- `Api/Services/Cost`: `RateCard` (parsing plus the arithmetic) and `ICostEstimator`, which reads the row behind a 30-second cache and treats a value written around the API as unconfigured rather than 500ing the dashboard.
- `EstimatedCostThisPeriod` appended to `DashboardSummaryResponse` and `TopConsumerResponse`; `EstimatedCost` appended to `QuotaAllocationResponse`. Rendered on `/dashboard` and `/quota`, each behind a tooltip that says it is an estimate and names #177.

**The caveats, stated plainly because the number sits next to a person's name.** `QuotaAllocation.TokensUsed` is a single total: **no prompt/completion split**, so the two prices collapse to their mean; **no per-model split**, so only the `*` fallback entry is ever read and the per-model entries are validated ahead of a reader. The totals are themselves a floor (#84, #88). Every figure is `null` — not zero — when no rate card is configured, because a zero reads as "free".

Follow-up filed: **#213** for the per-model, per-direction split. `Kql/UsageBySubscription.kql` already selects `PromptTokens`/`CompletionTokens` and `UsageSyncJob` sees `ModelName` on every row it aggregates, so both splits are discarded on the way into the database rather than missing at the source. Referenced from `RateCard`, both contracts and the docs.

## Verification

- `dotnet build FoundryGate.sln -c Release` — **0 warnings, 0 errors**.
- Predeployment suite **5/5 green**, 1515 tests each run.
- `npm run build` in `docs-site/` — clean, 15 pages.
- Docker: SQL container brought up (with `FoundryGate` created) to reproduce #197's failure, then stopped to confirm the fix holds both ways. Container and volume torn down afterwards.

## Decisions for the reviewer

1. **`isActive` is a fifth filter the issue did not ask for (#208).** It is what makes the links honest: both dashboard counts are scoped to active users, and the deprovision pipeline hard-stops *and* deactivates — so without it a card reading 2 would land on every developer who ever left. The cards link as `?isHardStopped=true&isActive=true`.
2. **The filters are named `isHardStopped` / `isOverBudget`**, not `hardStopped` / `overBudget` — matching `GET /users?isActive=` and the issue body's own wording, so one convention covers the whole query surface.
3. **`EstimatedCost` was appended to `QuotaAllocationResponse`** (not only the dashboard contracts): `/quota` renders allocations, and the Web references Domain only, so it had no other route to a priced figure. Appended at the end, like #190's counts, since the Web client deserializes these positionally.
4. **The blended rate is the mean of the `*` entry's two prices.** A fork that knows its real prompt/completion mix expresses it by setting both prices to its own blended number — the mean of two equal numbers is that number. A third "blended" field would only be a second way to write the same guess.
5. **`RateCard` lives in `FoundryGate.Api/Services/Cost`, not Domain.** Parsing needs `System.Text.Json`, and `DomainArchitectureTests` holds Domain to an exact three-assembly allowlist. Nothing outside the Api computes cost, and Domain gained only the `decimal?` fields.
6. **`DefaultWaitTimeout` was raised suite-wide (#203)**, not per test. It costs nothing while the suite is green and only lengthens a run that was going to fail — but it does mean a genuine Web-test failure now takes up to 10 s to report.

Closes #197
Closes #203
Closes #208
Closes #177


kolatts:
**REQUEST CHANGES**

The #197 fix is exactly right and the new 200 test closes the hole that made the old one meaningless; #208 is the best-executed piece here — filters, page, links and tripwires all agree with each other. Two things block: an admin can store a rate card that 500s `/quota/allocations/me` for every developer in the fork, and #203 is not actually fixed — the test it names still fails about a third of full-suite runs on this branch.

---

**Issues**

- **Major** `src/FoundryGate.Api/Services/Cost/RateCard.cs:88,101` — `Parse` bounds prices below (non-negative) but not above, so `PUT /config/RateCard` happily stores `inputPerMillion: 79228162514264337593543950335`; `BlendedRatePerMillion` then throws `OverflowException` on the add, and `GET /dashboard`, `GET /quota/allocations` and `GET /quota/allocations/me` all 500 until someone edits the row back by hand. `/allocations/me` is the endpoint every authenticated developer hits, so one accepted config edit takes the portal down. `CostEstimator`'s "reads must not 500 over it" catch is `ArgumentException`-only and the throw is in the caller anyway — the bound belongs in `Parse`.

- **Major** `src/FoundryGate.Tests.Predeployment/Web/DashboardPageTests.cs:259` — `Stops_refreshing_once_the_page_is_gone` failed **6 of 21** full-suite Release runs (`--no-build`) on this branch; `origin/main` failed **1 of 22** of the same runs, and the class passes 5/5 on its own. I instrumented `SettledCallCountAsync` and the count climbs at the full 20 ms tick for the entire 5 s budget after `DisposeComponentsAsync()` returns (`0:2, 75:5, 126:8, … 6121:306`), so this is not an impatient assertion — the refresh loop genuinely never stops. Worth noting the sibling test two methods down disposes deterministically with `page.InvokeAsync(dashboard.Instance.Dispose)` while this one leans on bUnit's teardown. Going from 2 to 5 settle samples made the test stricter, not the code more correct, and the PR closes #203 on the strength of it.

- **Minor** `src/FoundryGate.Api/Services/Dashboard/DashboardService.cs:117` — the rate card has its own 30 s `IMemoryCache` nested inside the dashboard's 30 s cache, and nothing evicts `CostEstimator.CacheKey` on `PUT /config/{key}`, so a corrected price can take ~60 s to appear and `?fresh=true` — the documented escape hatch — re-runs eight queries and still serves the stale rate.

- **Minor** `src/FoundryGate.Api/Services/Quota/QuotaAllocationService.cs:208,229` — `?isHardStopped=` and `?tier=` land with no `[Index]` on `QuotaAllocation` and nothing in `FoundryGate.Database/dbo/Tables/QuotaAllocations.sql`, against CONVENTIONS.md §Paging ("index the columns your default ordering and filters use … and mirror the index"). The period index probably makes them cheap residual predicates — but that reasoning isn't written anywhere, and `User.cs` sets the precedent of saying so in `<remarks>`.

- **Minor** `docs-site/src/pages/index.astro`, `docs-site/src/content/docs/getting-started/why-foundrygate.mdx` — both untouched. CLAUDE.md invariant 1 asks for the landing page to be re-read when a change alters "the pricing story", and this PR *is* the pricing story: it puts the fork's first money figure on screen. Invariant 2 wants the simplest rung updated too — `why-foundrygate` still says the reconciliation Function keeps "usage and cost figures authoritative" with no hint that cost is an operator-set `RateCard`, an estimate, and absent by default. The reference-level docs, for what it's worth, are excellent.

- **Minor** `src/FoundryGate.Tests.Predeployment/Web/WebTestContext.cs:65` — 1 s → 10 s across ~200 `WaitFor*` call sites. The stated cause is a 300 ms debounce plus runner scheduling; 3 s buys the same headroom. As written, a regression in something shared turns a ~3-minute red run into something closer to half an hour, which is exactly the run where you want the answer fast.

- **Nit** `src/FoundryGate.Api/Services/Cost/RateCard.cs` — 187 lines, and `ModelPrefix`'s "matched by prefix" contract has no matching code: `BlendedRatePerMillion` does an exact `Ordinal` lookup for `"*"` and nothing else is ever read. Honestly labelled as ahead-of-#213 rather than hidden, but today the behaviour is two decimals on a record.

- **Nit** `src/FoundryGate.Tests.Predeployment/Web/RecordedCalls.cs` — this is mostly `ConcurrentQueue<T>` with a `Last()`, and the dominant read in the new page tests (`QuotaAllocationListCalls[^1]`) compiles to `Count` then index — two lock acquisitions with a gap, so the snapshot guarantee in the class remarks is the one path those assertions don't take. Harmless as used (append-only), just not what the docs claim.

- **Nit** `src/FoundryGate.Web/Pages/QuotaAllocations.razor:290` — `_hasCostEstimates` is decided from the current page's rows, so an empty filtered page drops the "Est. cost" column and it reappears when you clear the filter.

---

**Checked and fine:** `RateCard.Parse` against a hostile-input matrix (`[null]`, `[[]]`, `["x"]`, `NaN`, `1e999999`, 5000-deep nesting, bool/null prices, `{}`, duplicate keys) — every one is an `ArgumentException` → 400, the overflow above is the only escape. `[FromQuery] QuotaAllocationQuery` cannot bind null and bad values are 400s, not 500s. No injection or XSS through `search`/`tier`. Auth on the controller, the page and the nav entry is consistent with every other admin surface. The readiness pair genuinely pins both branches. No N+1: the rate card is read once per page. `ApplyAsync` does not double-reload — the `OnParametersSetAsync` equality guard catches the `NavigateTo`. `RateCardEntry` being a positional record with `[property: …]` is fine: that rule is scoped to MVC-bound request bodies. `Value` carries `[DoNotUpdate]`, so a re-seed won't wipe a configured card. And this merges cleanly with `claude/final-sweep-d` — the only overlap is two docs files, no conflict.

kolatts:
Thanks — both Majors were right, and the second one was righter than the PR deserved: I had made the test stricter and called the bug fixed. All nine points addressed in `7640050`.

---

### Major 1 — `RateCard.Parse` bounds prices below but not above

Fixed. `MaxPricePerMillion = 1_000_000` — a guardrail, not a business rule, and the message says so. Three things changed, because the bound alone would still have left the read paths trusting a row they did not write:

1. `Parse` refuses anything above it (and the ceiling itself is accepted, tested against `long.MaxValue` tokens to show `decimal` is nowhere near its limit).
2. `Parse` now wraps `NotSupportedException` alongside `JsonException`, so `ArgumentException` is genuinely its only escape — which is what makes `CostEstimator`'s catch total rather than nearly so. You were right that it was `ArgumentException`-only; it just needed the *other* side of that contract nailed down too.
3. `BlendedRatePerMillion` and `Estimate` each answer `null` on overflow instead of throwing. The bound makes that unreachable for anything `Parse` accepts, but a row can still arrive from a seed script, a DBA or a restored backup, and a read path may not fail over a price list.

Your probe is now three tests: the validator (`RateCardTests` + a `ConfigEndpointTests` row asserting the `400`), and a new `RateCardResilienceEndpointTests` that writes `decimal.MaxValue` **straight to the table**, bypassing validation, and asserts `GET /dashboard?fresh=true`, `GET /quota/allocations` and `GET /quota/allocations/me` all still answer `200` with every cost field `null`.

### Major 2 — `Stops_refreshing_once_the_page_is_gone`

You were right, and the diagnosis is worse than "impatient assertion". I instrumented it the same way you did and then went one step further — the result is unambiguous:

| how the component is rendered | after `DisposeComponentsAsync()` |
|---|---|
| `Render<Dashboard>()` (root component type) | count freezes at 3 — disposed |
| inside a `RenderFragment` (what `RenderPage` does) | 14 → 24 → 34 at the full 20 ms tick |
| same, then `dashboard.Instance.Dispose()` by hand | freezes immediately |

So `Dispose()` always worked. **bUnit's teardown never reached the page**, because `WebTestContext.RenderPage` has to wrap it in a `RenderFragment` to put the dialog and popover providers beside it. The test was reporting a real defect the whole time: nothing had asked the page to stop. Going from 2 to 5 settle samples was, as you said, making the test stricter — I should have followed the signal instead.

Three changes:

- **Harness.** `RenderPage` records what it rendered; `WebTestContext` disposes it on demand (`DisposeRenderedPagesAsync`) and in teardown, `IAsyncDisposable` first, on the renderer's dispatcher. This affected *every* `IDisposable` page rendered through `RenderPage` — Dashboard is just the only one with a clock loud enough to hear. It was leaking a 20 ms `PeriodicTimer` into the rest of the suite: **full-suite duration went from ~12 s to ~4 s**, which I suspect is a good part of why other things looked flaky under runner load.
- **Component**, as defence in depth, since the real app does call `Dispose`: idempotent disposal, `IAsyncDisposable` alongside `IDisposable` (Blazor prefers the async one), and the loop re-checks disposal after the tick *and* inside `InvokeAsync`. Cancellation alone cannot close that window — `WaitForNextTickAsync` has already returned `true` by the time the token flips.
- **Test** disposes deterministically through the page's own `DisposeAsync`, mirroring the sibling test you pointed at, and still asserts the count stops climbing over a window 20× the tick.

**10/10 green** on the full Release suite (`--no-build`), 1523 tests, no failures of any kind.

### Minor — nested caches, and `?fresh=true` not reaching the rate card

Fixed both. `ICostEstimator.GetRateCardAsync` takes `fresh` and `GET /dashboard?fresh=true` passes its own flag through; `PUT /config/RateCard` calls `Invalidate()` **after** the commit (before it, an aborted write could open a window where the cache and the database disagree). Two service tests: the eviction happens for `RateCard`, and does not for any other key.

### Minor — no index for `?isHardStopped=` / `?tier=`

Taken, with one deviation I'd like you to sanity-check: rather than two single-column indexes I extended the existing period index to `(PeriodYear, PeriodMonth, TierProductId, IsHardStopped)`, mirrored in `QuotaAllocations.sql`. Reasoning, now written in `<remarks>` per the `User.cs` precedent: every read of this table is period-scoped first — the list, both its new filters, and every dashboard count — so the period columns have to lead; a standalone `IsHardStopped` index over two distinct values would never be chosen; and as trailing columns they cost nothing and turn both filters into seeks within a range already narrowed to one month. Same index count as before, no new write cost.

### Minor — landing page and `why-foundrygate`

You're right, this *is* the pricing story. Both updated:

- Landing page, in the register that page uses: *"You type in what your AI usage costs — Microsoft's bill arrives as one lump sum, so nobody but you knows how to split it — and the dashboard multiplies it out per person. It's an estimate, and it's labelled one everywhere it appears. Until you fill your prices in, no money figures show at all."*
- `why-foundrygate` no longer says the reconciliation Function keeps "usage **and cost** figures authoritative" — that sentence is now about usage, with cost as its own short paragraph: operator-set rate card, always an estimate, absent by default, and the gateway enforces tokens rather than dollars.

### Minor — `DefaultWaitTimeout` 1 s → 10 s

Fair, and 10 s was me over-correcting. Now **3 s** — ten times the longest debounce these pages use, with the reasoning in the comment: the number only decides how long a *failing* wait takes to give up, so it should be chosen to keep a red run readable.

### Nit — `ModelPrefix`'s contract has no matching code

Dropped the claim rather than implementing it, since the per-model split is #213. `RateCardEntry`'s remarks and the reference page now say plainly that nothing matches on `modelPrefix` today and only the `*` entry is read — and why the field is in the stored shape now anyway (so no fork has to rewrite its row when #213 lands).

### Nit — `RecordedCalls` and the `[^1]` path

Good catch: `[^1]` was lowering to `Count` then `this[int]`, two acquisitions with a gap, so the dominant read was the one the class remarks didn't cover. Added an `Index` indexer that takes the lock once, and corrected the remarks to describe what it actually does.

### Nit — `_hasCostEstimates` decided from the current page

Now sticky, with the reasoning in `<remarks>`: a rate card is a property of the fork, not of the page you happen to be on, and a column vanishing on an empty filtered page reads as the page losing track of something. New test: filter to nothing, column stays.

---

**Verification:** build 0 warnings / 0 errors, `dotnet format --verify-no-changes` clean, docs build clean, Predeployment **10/10 green at 1523 tests**.

Issue #214, opened 2026-09-02
Converge the two deprovision pipelines: lift plan 21's departure sequence into Core


kolatts:
> Parent: #224 — Engineering debt

Discovered while implementing #151 (scheduled Entra sync).

The nightly `EntraSyncFunction` deprovisions people the directory no longer lists. On the Api that goes through plan 21's one orchestrator, `IUserLifecycleService.DeprovisionAsync(EntraDeparture, ...)`. Core cannot reference the Api — the orchestrator composes `IApimKeyService` and `IQuotaRequestService`, both Api-only — so `FoundryGate.Core.Entra.DeprovisioningDepartureHandler` now does the same sequence over the pieces Core owns:

1. delete the APIM subscription via `IApimManagementClient`, clear the key fields, write a system `key.revoked` row (mirrors `ApimKeyService.RevokeCoreAsync`);
2. `IsActive = false`, hard-stop the current period allocation, close pending requests, write a system `user.deactivated` row (mirrors `UserLifecycleService.DeprovisionAsync`).

**Why it was written rather than avoided.** The alternative considered was a Functions handler that only *flags* departures. That is worse than useless: the admin UI would read "inactive" while the gateway still honoured the developer's key, which is exactly the "SQL says one thing, the gateway enforces another" class the whole tier-sync design exists to prevent.

**Why it is debt anyway.** Two implementations of the most safety-critical sequence in the product can drift, and CONVENTIONS.md 167;Solution structure says to prefer a narrow seam over a second implementation of the whole thing. One piece already converged: the pending-request rule is Core's `IQuotaRequestExpiry.CancelPendingForUserAsync`, which the Api's `QuotaRequestService` now delegates to.

## Ask

Lift the rest of the deprovision pipeline into Core so there is one definition again. Sketch:

- Move the revoke half of `IApimKeyService` into Core (it needs only `IApimManagementClient` + `IAuditWriter` + the `User` row — key *protection* stays in the Api, which is the only host that stores or reveals a key).
- Give `IUserLifecycleService.DeprovisionAsync` a Core body that both hosts call, with the Api keeping the trigger-specific audit attribution it has today.
- `DeprovisioningDepartureHandler` then becomes the thin thing `IDepartureHandler` was meant to be, or disappears entirely.

[...]

PR #216, opened 2026-09-02
feat: emit the quota tier table from infra, and schedule the Entra syncs nightly

kolatts:
Two independent issues, one commit each.

---

## `Closes #201` — infra emits `Gateway__Tiers__*` from the `quotaTiers` parameter

A developer's monthly budget **is** a gateway tier (D-013), but the tier table was the one part of the `Gateway` section infra never set: both hosts shipped it in their own `appsettings.json`, mirroring `infra/main.bicep`'s `quotaTiers` *default*. A fork that overrode `quotaTiers` at deploy time got APIM products and `llm-token-limit` caps the control plane had never heard of, and nothing detected it.

- `control-plane.bicep` takes `quotaTiers` and projects it onto the Container App and the Function App as `Gateway__Tiers__{i}__ProductId` / `__DisplayName` / `__MonthlyTokenQuota` — the same loop shape `Gateway__ModelAliases__*` uses — plus a `quotaTierRows` output surfaced from `main.bicep`, so a deploy can be checked without opening the app-settings blade.
- The shipped tables are **gone** from both `appsettings.json` files. Two sources for one table means the copy a deployment did not set silently wins, and a fork whose `quotaTiers` has fewer entries than the shipped three would keep the leftover index as a phantom tier. Local dev keeps working from `appsettings.local.json` in each project (un-ignored, with the reason in `.gitignore`); `local.settings.template.json` says where the table lives rather than carrying a third copy.
- `GatewayOptions.Validate`'s "at least one tier" message now names the env vars **and** the local file, so whoever reads it knows both places to look.

**Tests.** `GatewayInfraBindingTests` loses its `NotSetByInfra` entry (now empty — the tier table was the last holdout) and inverts `Infra_does_not_emit_the_tier_table` into one that pins the presence *and* the three fields — an infra emitting only `ProductId` would bind nameless zero-cap (= unlimited) tiers, which validates and is wrong — plus a check that the settings are derived from the parameter rather than typed out a second time. The appsettings-vs-bicep parity tests retarget to the local files; `FunctionsAppSettingsTests` gains `Neither_host_ships_a_tier_table_in_appsettings_json`; `ApiTestFactory` states the tiers via `UseSetting` in the deployed shape.

---

## `Closes #151` — nightly Entra sync (users, then groups) as a timer-triggered Function

`EntraSyncFunction` runs at 02:00 UTC daily (`RunOnStartup=false`): users sync first, then every linked group. That order is the point — group sync skips directory members with no `User` row, so running it second turns tonight's joiners into memberships tonight rather than tomorrow, and it is why the two are one job rather than two timers. Both are behind the blob-lease lock the monthly reset uses (`IResetLock` generalised to `IJobLock`, a distinct lease name per job so the two never block each other), gated by `Entra:Enabled` (disabled → one Information line, no error, no lock taken), and the run writes one system-attributed `entra.scheduled-sync` row carrying both halves' counts.

To run there, the Entra area moved to `FoundryGate.Core/Entra/` (#119's pattern): `IEntraDirectoryClient` + its two implementations, `EntraOptions`, `EntraUserSyncService`, `EntraGroupSyncService`. Two host-shaped differences became one-method seams — `IEntraSyncActor` (whose audit row a run is) and `IDepartureHandler` (what happens to someone the directory dropped).

### The decision the reviewer should look at: `IDepartureHandler` has two real implementations

The Api's delegates to plan 21's one orchestrator, `IUserLifecycleService`. Core cannot reach it — it composes `IApimKeyService` and `IQuotaRequestService`. So `DeprovisioningDepartureHandler` re-states the sequence over the pieces Core owns (the APIM management client from #194, `IAuditWriter`, the `AppDbContext`): delete the subscription, clear the key fields, deactivate, hard-stop the allocation, close pending requests — same audit actions, same details shapes, same commit-point discipline.

That is a second implementation of the most safety-critical sequence in the product, and it is **filed as #214**. It was still the right call versus the alternative in the issue (flag departures, leave the key): that would have made the admin UI report someone as deprovisioned while the gateway still honoured their key — the exact "SQL says one thing, the gateway enforces another" class the tier sync exists to prevent, except applied to offboarding. Full reasoning in **D-020**.

Two things converged rather than being duplicated: the pending-request rule is Core's `IQuotaRequestExpiry.CancelPendingForUserAsync`, which the Api's `QuotaRequestService` delegates to, and every string the trail is made of is Core's `DepartureAudit` (the Api's `EntraDepartureReason` / `DeactivationReviewNote` are `const` aliases, so the compiler enforces those two). The review caught the reason this matters: the first cut wrote `trigger = "IDepartureHandler"` where the Api writes `"EntraDeparture"`, so an operator filtering `details.trigger == "EntraDeparture"` would have seen every admin-found departure and none of the nightly ones. `DepartureHandlerParityTests` now runs both pipelines against one database on one clock and compares both details objects field for field — verified by re-introducing the divergence and watching it fail.

### Two things the move surfaced

- `EntraOptions` in Core fell out of `ValidateRecursively()` on both hosts — the D-014 trap again, caught only because an existing Api test noticed. `CoreOptionsValidation` gains `ValidateEntra` alongside `ValidateGateway`; both hosts' `AppSettings` yield it.
- The Graph client could no longer read `AzureAd:ClientId` — the Functions worker binds no `AzureAd` section, because nothing there serves a request. The app registration id travels on `Entra:ApplicationClientId`, which infra now sets on **both** hosts and the Api defaults from `AzureAd:ClientId`, so an older deployment keeps working.

### Placement that is load-bearing

`IEntraSyncActor` is resolved **late** in the user sync — exactly where `IAuditService.LogAsync` used to resolve it. The Api's accessor checks the change tracker, so an admin whose own `User` row that run just added still attributes the run to themselves instead of being 403'd by it. Group sync resolves it first, as it always did, so an unprovisioned admin's 403 lands before Graph is read.

### Owner step, not Bicep-able

The Functions identity needs the same three Graph application roles as the Api's (`Application.Read.All`, `User.Read.All`, `GroupMember.ReadBasic.All`) because it calls Graph as itself. Documented in `reference/infrastructure` §Role assignments and noted on #109 and #120 with a two-identity grant snippet. Infra deliberately still does **not** set `Entra__Enabled`.

---

## Verification

| Gate | Result |
|---|---|
| `dotnet build FoundryGate.sln -c Release` | Build succeeded, **0 warnings**, 0 errors |
| `FoundryGate.Tests.Predeployment` | **1465 passed, 0 failed** (1460 + 5 from the review's parity guard) |
| `dotnet format FoundryGate.sln --verify-no-changes` | clean |
| `az bicep build --file infra/main.bicep` | 0 errors (3 pre-existing `BCP081` warnings on `Microsoft.CognitiveServices/accounts@2026-07-01`) |
| `az bicep lint --file infra/main.bicep` | clean |
| `az bicep build-params` for `test` / `dev` / `prod` | exit 0 each |
| `docs-site` `npm ci && npm run build` | 15 pages built |

**Flakes seen, disclosed.** Mid-work, two known-flaky tests failed on individual runs and passed on re-runs: `HealthEndpointTests.HealthReady_returns_503_rather_than_throwing_when_the_database_is_unreachable` (verified pre-existing — it fails on the unmodified tree too, and `HealthEndpointTests` is in another agent's scope) and `DashboardPageTests.Stops_refreshing_once_the_page_is_gone` (the #203 family). The final full run above was green.

No live Azure or Graph was touched.

## Deferred, with issues

- **#214** — converge the two deprovision pipelines by lifting plan 21's departure sequence into Core, so `DeprovisioningDepartureHandler` becomes thin or disappears. Filed as part of this work, referenced from the code and from D-020.
- **#109 / #120** — the Graph application roles for the Functions identity are an operator step; both issues now carry the two-identity grant snippet and the extra live-validation line.

## Notes for the reviewer

- The Functions host now registers `UnconfiguredApimManagementClient` when `Gateway:*` does not address APIM, where it previously registered no client at all, because the departure handler needs one to start. **This is not graceful degradation** (an earlier draft of this body implied otherwise): its `FeatureNotConfiguredException` is in neither `DeprovisioningDepartureHandler.IsUpstreamFailure`'s allowlist nor `EntraUserSyncService`'s per-departure catch, so a departure that reached it would abort the whole nightly pass as a failed invocation — exactly what the Api does on the same configuration. It is unreachable in practice: no gateway means no `ApimSubscriptionId`, and `RevokeKeyAsync` returns `false` before the client is touched. Loud-and-unreachable is the right shape for a misconfiguration.
- Functions `appsettings.local.json` and the Api's are both un-ignored in `.gitignore` with a comment; since #201 they are the only place the tier table lives for `local`, so ignoring them would mean a fresh clone cannot start.
- The Entra tests moved from `Api/Services/{Entra,Groups}` to `Core/Entra` to follow the code. Their assertions are unchanged apart from the seam wiring — the service-level tests now construct the Api's `CurrentUserEntraSyncActor` / `LifecycleDepartureHandler`, so they still test what an admin's button does.


kolatts:
**REQUEST CHANGES**

Two clean issues, one commit each, and the infra half (#201) is genuinely well done — one parameter, two consumers, and I broke the local tier table on purpose to confirm both parity guards actually fire. The Entra half is right too, with one exception: D-020 and the PR body both promise "the same audit actions, the same details shapes," and the two pipelines already disagree on the one field that says *why* the row exists. One-word fix, but #214 named this exact drift as the risk and it happened before the PR even merged, so it shouldn't land asserting a parity it doesn't have.

---

**Issues**

- **Major** `src/FoundryGate.Core/Entra/DeprovisioningDepartureHandler.cs` line 101 — the `user.deactivated` details write `trigger = nameof(IDepartureHandler)` → `"IDepartureHandler"`, where the Api writes `"EntraDeparture"` (`UserLifecycleServiceTests` line 382 pins that literal). An operator filtering the audit log on `details.trigger == "EntraDeparture"` sees every admin-found departure and none of the nightly ones. `DeprovisioningDepartureHandlerTests` asserts `keyRevoked` and `allocationHardStopped` but never `trigger`, which is why nothing caught it.

- **Minor** `src/FoundryGate.Core/Entra/DeprovisioningDepartureHandler.cs` lines 53–56 — `Reason` and `ReviewNote` are independent literals whose only tie to `UserLifecycleService.EntraDepartureReason` / `.DeactivationReviewNote` is an XML-doc comment saying "Matches". Both are `internal` and the test project can see them; two `Assert.Equal`s would turn "we keep these in step by hand" into something the build enforces until #214 lands. The same guard would have caught the `trigger` divergence's sibling.

- **Minor** `src/FoundryGate.Functions/Services/Entra/EntraSyncJob.cs` line 99 — the `entra.scheduled-sync` row and its save run on the raw `cancellationToken`, on a pass whose users half has already deleted APIM subscriptions. Everywhere else past a commit point this codebase switches to `CancellationToken.None`; a host shutdown mid-run loses the row saying the schedule ran on a night it offboarded people. The users sync's own `CommitToken` covers its rows, not this one.

- **Minor** `docs-site/src/content/docs/reference/configuration.mdx` line 14 — the `DefaultMonthlyTokenQuota` row still links `#gateway-quota-tiers-gatewaytiers`, but the heading is now `Gateway quota tiers (Gateway__Tiers__*)`, so the slug moved to `gateway-quota-tiers-gateway__tiers__` (line 150 already uses the new one). Astro won't fail the build on a dead in-page anchor, which is the problem.

- **Nit** `src/FoundryGate.Core/Entra/DeprovisioningDepartureHandler.cs` line 68 — Core drops the Api's explicit transaction and its `.Local`-first allocation lookup. Equivalent today (one `SaveChangesAsync` is atomic; nothing adds an unsaved allocation ahead of a departure), but it's a third axis of divergence to carry in your head until #214, and unlike the other two the remarks don't mention it.

- **Nit** `.gitignore` line 38 — "They carry the docker-compose SQL password and nothing else secret" is true of the Api's file and not of the Functions one, which has no password at all. For anyone auditing later: `Temp1234!` was already committed in `docker-compose.yml`, `SetupCommand.cs`, `local.settings.template.json` and the Api's `appsettings.local.json` (already tracked on main — that un-ignore line is a no-op), so this PR exposes nothing new.

- **Nit** `src/FoundryGate.Tests.Predeployment/Api/Endpoints/QuotaEndpointTests.cs` line 92 — `Tiers_are_readable_by_any_authenticated_user_and_mirror_the_shipped_appsettings`. There is no shipped appsettings any more; the whole point of #201 is that it mirrors the factory's deployed-shape settings.

---

**On the two things the PR asked me to look hardest at**

**Was #214 the honest home for convergence, or should this PR have lifted the Api sequence into Core?** #214 is the right call. Lifting `RevokeCoreAsync` into Core drags the key protector's shape and the `IApimKeyService` surface across in a change whose actual subject is a timer trigger, and D-020's option 2 — flag the departure, leave the key live — is genuinely worse than the duplication. Line by line against `UserLifecycleService.DeprovisionAsync` on main, the sequence is faithful: APIM delete first and outside the transaction, `CommitToken.For(keyRevoked, …)`, key fields cleared and `key.revoked` committed on `CancellationToken.None`, `IsActive = false`, hard-stop, `CancelPendingForUserAsync` through the shared Core rule, one `user.deactivated`, idempotent on an already-inactive user and on a subscription APIM no longer has, and the same allowlist `IsUpstreamFailure`. It got everything right except the field naming the trigger — which is precisely the argument for pinning the constants rather than trusting a comment.

**The unconfigured APIM client on a departure.** It doesn't silently flag-only. `UnconfiguredApimManagementClient.DeleteSubscriptionAsync` throws `FeatureNotConfiguredException`, which isn't in `IsUpstreamFailure` and isn't caught by `EntraUserSyncService`'s per-departure handler, so it aborts the whole nightly pass rather than deactivating anybody. Same behaviour the Api has on main, and unreachable anyway (no gateway ⇒ no `ApimSubscriptionId` ⇒ `RevokeKeyAsync` returns `false` first). Fine as shipped. But the PR body's "the `FeatureNotConfiguredException` the one call that could reach it deserves" reads like graceful degradation, and there isn't any — it's a failed invocation. Worth a word in the registration comment.

The `IEntraSyncActor` late-resolution rationale holds. It sits exactly where `IAuditService.LogAsync` used to, the Api's accessor checks the change tracker, and group sync resolving first keeps the unprovisioned-admin 403 ahead of the Graph read. Moving it would be a silent behaviour change, and the remarks say so in the right place.

---

**Verification** (worktree at `origin/claude/final-sweep-d`, `37ff315`)

| Gate | Result |
|---|---|
| `dotnet build FoundryGate.sln -c Release` | 0 warnings, 0 errors |
| `dotnet test FoundryGate.Tests.Predeployment -c Release` | 1460 passed, 0 failed |
| `dotnet format --verify-no-changes` | clean |
| `az bicep build` / `lint` on `infra/main.bicep` | clean (3 pre-existing BCP081) |
| `az bicep build-params` — dev / prod / test | exit 0 each |
| `dotnet run --project src/FoundryGate.Api`, `ASPNETCORE_ENVIRONMENT=local` | `/health` → `200 {"status":"Healthy","environment":"local"}`; `/api/v1/quota/tiers` → 401 (auth), so startup and the always-required tier validation passed from `appsettings.local.json` |

Two probes, since the tier move has the most ways to be quietly wrong:

- Raised the Standard cap to 7,000,000 in the Api's `appsettings.local.json` and re-ran `Tiers_are_readable_by_any_authenticated_user…`: still green. So `ApiTestFactory`'s `UseSetting` block really does win over the local file, and the "deployed shape, hermetic" claim in that comment is real rather than aspirational.
- Same edit against the two parity guards: `The_local_tier_table_matches_infra_main_bicep_quotaTiers_ids_and_caps_in_order` and `The_local_Functions_tier_table_is_identical_to_the_Apis` both failed as designed. Restored afterwards.

`GatewayTier` has exactly `ProductId` / `DisplayName` / `MonthlyTokenQuota`, so the "all three fields" pin in `GatewayInfraBindingTests` is complete rather than a snapshot of today, and `NotSetByInfra` going empty is honest.

---

**Merge order vs #215**

`git merge-tree` is clean in both directions. The only shared files are `architecture/overview.mdx` and `reference/configuration.mdx`, with disjoint hunks.

Merge **#215 first** — it fixes `HealthEndpointTests` and `DashboardPageTests`, the two flakes this PR had to disclose, so #216's CI comes back meaning something. Whichever goes second needs to rebase, re-run the full suite plus `az bicep build`, and actually *read* the merged `configuration.mdx` and `overview.mdx`: git will splice those two silently and nobody will have looked at the result as one document.

kolatts:
Fix pass in `f6fae5e`. Point by point.

## Major — `trigger` divergence

Fixed, and you're right that it's the wrong bug to ship in a PR whose own body claims "the same details shapes." The comment couldn't fail a build, so it didn't.

The three strings the trail is made of now live in Core's `DepartureAudit` (`Trigger` = `"EntraDeparture"`, `KeyRevocationReason`, `ReviewNote`). Core writes them directly; the Api's `EntraDepartureReason` / `DeactivationReviewNote` became `const` aliases, so **for two of the three the compiler enforces the identity** rather than a test agreeing with it after the fact.

`trigger` is the one that can't be: the Api's value comes from `DeprovisionTrigger.ToString()`, which has to keep saying `"AdminDeactivation"` for the other trigger, and Core can't reference an Api enum. So that tie is an assertion — `Assert.Equal(DepartureAudit.Trigger, nameof(DeprovisionTrigger.EntraDeparture))` — which fails if the member is renamed without the constant.

New `DepartureHandlerParityTests` (5 tests) is the guard you asked for. Both pipelines run **against one database, on one clock, from identical fixtures**, so the two details objects compare field for field including `deactivatedDate`:

- `user.deactivated`: property *names* first (a field added to one side alone fails as shape, not as a confusing value mismatch), then `trigger`, `keyRevoked`, `allocationHardStopped`, `cancelledRequestCount`, `period`, `deactivatedDate`.
- `key.revoked`: names, `reason`, `existedInApim` — and `subscriptionName` asserted *per user* rather than skipped, so a handler writing somebody else's subscription still fails.
- Both rows system-attributed by both pipelines; closed requests carry the same note and no reviewer.

Verified the guard actually guards: I put `nameof(IDepartureHandler)` back, rebuilt, and watched exactly one test fail (`The_two_pipelines_write_identical_user_deactivated_details_for_the_same_departure`), then restored it.

## Minor — share the reason/note constants

Done, but **not the way you specified, and I want to flag the deviation** rather than have you find it. You suggested two `Assert.Equal`s. I made them aliases instead:

```csharp
internal const string DeactivationReviewNote = DepartureAudit.ReviewNote;
internal const string EntraDepartureReason = DepartureAudit.KeyRevocationReason;
```

Two reasons. The alias is strictly stronger — it's one value, checked by the compiler, not two values checked by a test. And the test as described couldn't have been written: those constants are `internal` and `FoundryGate.Api` grants no `InternalsVisibleTo` to the test project (I tried it first and it didn't compile). There's a comment in the parity class recording both facts so the next person doesn't re-derive it.

## Minor — `EntraSyncJob` commit token

Fixed. The row and its save now run on `CommitToken.For(userResult.DeactivatedCount > 0, cancellationToken)`.

One judgement call in the predicate: I did **not** fold group sync's tier moves in. `EntraGroupSyncService` owns that commit point and has already committed its rows by the time this line runs, so including it would extend `CancellationToken.None` over a save that has nothing uncommitted behind it. The comment says so, so if you disagree the reasoning is visible rather than implied.

## Minor — dead anchor

Fixed. I also checked it the way the problem demands, since Astro won't: built the site and grepped the output — heading renders as `id="gateway-quota-tiers-gateway__tiers__"`, and all three references (the two body links plus the ToC) point at it.

## Nits

- **Third divergence axis.** The Core handler's remarks now carry a numbered list of *all three* — no explicit transaction, no `.Local`-first lookup, the restated sequence — each with why it's an equivalence today and what would break it. The transaction one is only equivalent because those steps end in a single `SaveChangesAsync` and, unlike the Api's path, nothing here shares the unit of work with a caller staging other writes; a step needing a second save would have to open one. Both point at #214.
- **`.gitignore` comment.** Corrected, using your finding: nothing new is exposed, the Api's file was already tracked (that un-ignore line is a no-op that only makes it explicit), and the Functions one is a tier table with no credential in it at all.
- **Stale test name.** `Tiers_are_readable_by_any_authenticated_user_and_mirror_the_hosts_bound_tier_table`, with a comment saying "bound, not shipped" and why.

## The unconfigured-APIM wording

You're right, and thank you for tracing it — "the exception it deserves" did read like degradation and there isn't any. Both the registration comment and D-020 now say plainly: `FeatureNotConfiguredException` is in neither `IsUpstreamFailure`'s allowlist nor `EntraUserSyncService`'s per-departure catch, so a departure reaching it **aborts the whole nightly pass as a failed invocation** — same as the Api on the same configuration — and it's unreachable anyway because no gateway means no `ApimSubscriptionId` and `RevokeKeyAsync` returns `false` first. Loud-and-unreachable is what the registration buys; nothing pretends a gateway-less host can offboard anyone.

## Verification

| Gate | Result |
|---|---|
| `dotnet build FoundryGate.sln -c Release` | 0 warnings, 0 errors |
| `FoundryGate.Tests.Predeployment` | **1465 passed, 0 failed** (was 1460; +5 parity) |
| `dotnet format --verify-no-changes` | clean |
| `az bicep build` — `infra/main.bicep` | exit 0 (3 pre-existing BCP081) |
| `docs-site` `npm run build` | 15 pages; anchor slug verified in the built HTML |

Also recorded in D-020, since both corrections are decisions rather than typos: the `trigger` drift as evidence that the doc-comment approach was insufficient, and the abort semantics stated precisely.

Not rebased on #215 yet — will do that when it merges, then re-run the suite plus `az bicep build` and read the merged `configuration.mdx` and `overview.mdx` end to end, as you asked. Not merging.

kolatts:
One addition since the response above, in `08b2270`.

I said the `EntraSyncJob` commit-token fix was made; I then went to prove it with a test and found the fix's protected window is narrower than the comment implied, so I want that on the record rather than discovered later.

`CommitToken.For(userResult.DeactivatedCount > 0, ct)` only bites if cancellation lands **strictly between `SyncAllAsync` returning and the job's save** — a host shutdown at that instant. Cancelling any earlier (I tried the obvious thing: cancel the moment the fake APIM accepts the deletion) aborts inside one of the two syncs instead, because each correctly honours the raw token up to its own commit point. The job's row is then never reached and the test would be asserting the sync's behaviour, not the job's.

So: the fix stays — it is what CONVENTIONS.md asks for past a commit point and it costs nothing — but there is no test for that half, and a comment in `EntraSyncJobTests` says so and why, rather than leaving a silent gap. Reaching the real window needs a seam between the two calls that exists only for the test, which seemed a worse trade than the comment. Say the word if you'd rather have the seam.

The reachable half **is** pinned: `A_pass_that_offboarded_nobody_still_honours_cancellation` — nothing external touched, so an abandoned run stops instead of forcing its row through on `CancellationToken.None`.

Suite is **1466 passed, 0 failed**; build 0 warnings; format clean. Still not rebased on #215, still not merged.

kolatts:
**Rebase note** — merged `origin/main` (with #215) in `5792c5b`, plus one follow-up in `2e23ede`.

`git merge origin/main` was **conflict-free**. Checked the four things you flagged:

- **`ApiTestFactory`** — #215 never touched it, so "keep both sides" was moot; my `Gateway:Tiers:*` `UseSetting` block is intact and is still the only edit to that file since the base.
- **`WebTestContext`** — took main's version whole: `DefaultWaitTimeout = WaitTimeout` (3 s), `_renderedPages` list, `DisposeRenderedPagesAsync` / `DisposeComponentAsync`. I touched none of it.
- **`QuotaAllocation`** — main's widened index is present: `[Index(PeriodYear, PeriodMonth, TierProductId, IsHardStopped)]` alongside the unique `(UserId, PeriodYear, PeriodMonth)`.
- **The two spliced docs** — read end to end, and one of them genuinely needed fixing (below).

## What the splice broke, and the fix

`reference/configuration.mdx` came out fine: #215's `RateCard` row and `### Cost estimates` section sit cleanly beside my rewritten tiers section, and **both** in-page anchors resolve in the built HTML — `id="gateway-quota-tiers-gateway__tiers__"` (3 refs) and `id="cost-estimates-ratecard"` (2 refs). No stale "shipped appsettings" tier claims survive anywhere in the repo.

`architecture/overview.mdx` did **not**. Git spliced my nightly-schedule sentence into the middle of "Bulk user sync" — between the departures paragraph and the group-assignment paragraph — cutting the "what the sync does to people" thread in half. Neither side's diff looked wrong on its own; only reading the section as a document showed it. Moved to the end of the section where it closes the topic, and trimmed, since the scheduled jobs are described in full two sections down.

I did not add a cross-link to that fuller description: I'd have had to guess the enforcement section's slug, and guessing a slug is what put the dead anchor in `configuration.mdx` earlier in this PR.

## Gates on the merged tree

| Gate | Result |
|---|---|
| `dotnet build FoundryGate.sln -c Release` | 0 warnings, 0 errors |
| `FoundryGate.Tests.Predeployment` | **1545 passed, 0 failed** (1466 mine + #215's) |
| `dotnet format --verify-no-changes` | clean |
| `az bicep build --file infra/main.bicep` | exit 0 (3 pre-existing BCP081) |
| `docs-site` `npm run build` | 15 pages; both anchors verified in the output |

Worth noting: the full suite is green **first time** now, no re-run. The two flakes this PR had to disclose (`HealthEndpointTests.HealthReady_…`, `DashboardPageTests.Stops_refreshing_…`) were #215's to fix and it fixed them, so #216's CI signal means what it says from here on.

Not merged.

PR #144, opened 2026-09-01
feat(ci): deploy pipeline — infra, destroy, api, functions, ui, docs, deploy-all workflows + GitHub Environments (#68–#75)

kolatts:
## Summary

The rest of epic #67: every deploy workflow from `plans/22-cicd-pipelines.md`, wired to the real infra contract from PR #111 and to `ci.yml` / `_deploy-database.yml` (neither modified here). OIDC only, **zero GitHub secrets**, GitHub Environments as gates, GitVersion trunk-based, one lock per environment.

**A merge to `main` runs exactly one workflow.** `deploy-all.yml` carries the `push` trigger and the ordered chain: infra → prepare-database → database → api → functions ∥ ui → postdeployment tests → summary. The four single-component wrappers have no `push` trigger at all — they keep their PR-track jobs plus `workflow_dispatch` for targeted redeploys. (Rev 2, after review: they used to share the `deploy-{env}` concurrency group, and GitHub keeps one pending run per group, so a `Domain/**` change started three workflows and silently cancelled one.)

Full reference (every workflow, trigger, gate, variable, runbooks): `docs-site/src/content/docs/reference/ci-cd.md`.

Closes #68
Closes #69
Closes #70
Closes #71
Closes #72
Closes #73
Closes #74
Closes #75
Closes #14
Closes #140

## Workflow inventory

| File | Triggers | Jobs | Gate | Reads |
|---|---|---|---|---|
| `deploy-all.yml` | **push `main`** — `infra/**`, `src/**`, `Directory.*.props`, `global.json`, `NuGet.config`, `.github/workflows/_deploy-*.yml`, `.github/scripts/**`, `.github/actions/**` · dispatch (`environment`, `create-model-deployments`, `run-seed-test`) | infra → prepare-database → database → api → functions ∥ ui → postdeployment-tests → summary, under `deploy-{env}` / `cancel-in-progress: false` | the target environment, per stage | — |
| `infra-deploy.yml` | PR `infra/**` → what-if comment · dispatch (`environment`, `create-model-deployments`) | → `_deploy-infra.yml` | `dev-plan` (PR) · `dev` / `production` (dispatch) | `vars.AZURE_*`, `FG_ENTRA_API_CLIENT_ID`, prod `FG_SQL_ADMIN_GROUP_*`, optional `AZURE_LOCATION` |
[...]

PR #227, opened 2026-09-05
infra+docs: dev owner setup — dedicated SQL admin group, owner-setup runbook

kolatts:
## What

Stands up the `dev` environment's Entra/RBAC/GitHub-Environment prerequisites so the first real `Deploy All → dev` (#105) can actually run, and writes down how it was done.

The handoff called these "owner actions that cannot be done by an agent". That turned out to be about **privileges**, not tooling — a CLI session signed in as a subscription Owner with Entra application-administration rights drives the whole list. So the wording is gone and the commands are a runbook instead.

## Repo changes (small)

- **`infra/parameters/dev.bicepparam`** — `sqlAdminGroupObjectId`/`sqlAdminGroupName` now point at a dedicated `SG_FOUNDRYGATE_SQL_ADMINS` group (`186dafe0-…`) instead of the tenant-wide `SG_IMAGILE_SQL_ADMINS` fallback. Azure SQL is Entra-only, so the group *is* the deploy identity's only way in; a dedicated group means the dev database's admin set is not "everyone who administers any Imagile SQL server".
- **`docs-site/src/content/docs/reference/owner-setup.md`** (new) — the `az`/`gh` commands that ran, in order, for a fresh environment. Includes the Graph PATCHes `az ad app` cannot express (exposed scope, app role, SPA redirect URIs, pre-authorization) and the parts that only bite once: `az ad group member list` reads `/members/microsoft.graph.user` and therefore *never* shows a service-principal member (verify from the principal's `memberOf` instead), and the managed identities' Graph app roles cannot be granted before the deploy that creates the identities.
- **`docs-site/.../reference/infrastructure.md`** — the dev SQL admin group name in the parameter table.
- **`handoff-orchestrator.md`** — records the created identifiers and drops the "cannot be done by an agent" framing.

## Created outside the repo

All identifiers, not secrets — hence GitHub Environment **variables**, and no credential anywhere (OIDC federated credentials only).

| Thing | Name | Client id |
|---|---|---|
| API app registration (`access_as_user` scope + `FoundryGate.Admin` role, v2 tokens) | `FoundryGate.Api (dev)` | `7e7d0561-0973-411d-ba62-a667cbfec1d9` |
| SPA app registration (localhost redirect URIs; SWA hostname added post-deploy) | `FoundryGate.Web (dev)` | `21b82312-f9f9-4be6-b243-34bf7256b557` |
| CI deploy identity — Owner on the subscription, `environment:dev` + `environment:dev-destroy` | `foundrygate-ci-dev` | `88f05620-03f2-408e-810f-0e25f668b6a7` |
| PR what-if identity — Reader only, `environment:dev-plan` | `foundrygate-ci-dev-plan` | `ec8f8758-43f7-4a5c-9ad4-70d71807ec76` |
| SWA preview identity — no RBAC yet (custom role is granted after the first deploy) | `foundrygate-ci-ui-preview` | `073418b1-a73f-4997-8805-5f3a9ec0fbda` |

Variables set on `dev`, `dev-destroy`, `dev-plan`, `ui-preview`. Production identities are deliberately **not** created — that stays open on #109 with a `needs-human` decision about reviewers.

## Verification

- `npm ci && npm run build` in `docs-site/` — clean, 16 pages.
- Role assignments read back: `foundrygate-ci-dev` = Owner, `foundrygate-ci-dev-plan` = Reader, subscription scope.
- `foundrygate-ci-dev`'s SP `memberOf` includes `SG_FOUNDRYGATE_SQL_ADMINS`.
- Admin-consent grants on the SPA read back as `User.Read` + `access_as_user`.
- The `FoundryGate.Admin` app-role assignment to the owner exists on the API service principal.
- The real verification is the deploy this unblocks — #105.

Refs #109, #105

### DIFF
--- a/docs-site/src/content/docs/architecture/overview.mdx
+++ b/docs-site/src/content/docs/architecture/overview.mdx
@@ -318,12 +318,16 @@ silently cancelled. They stay as `workflow_dispatch` buttons for redeploying one
 Gating it all: six GitHub Environments (`dev`, `dev-plan`, `ui-preview`, `production`,
 `dev-destroy`, `prod-destroy`); `dev` is restricted to protected branches, so no pull request
 can reach the identity that is Owner on the subscription. The two Environments a pull request
-*can* reach are deliberately tiny: `dev-plan` is Reader on the subscription and does nothing but
-a what-if, and `ui-preview` holds a single custom role — `FoundryGate SWA Preview Publisher`,
-defined in `infra/modules/swa-preview-role.bicep` — assignable only on the dev Static Web App.
-A pull request runs its own copy of the workflow files, so the identity, not the workflow, is
-the boundary. **What has not happened yet**: a run against Azure. That needs the owner-side OIDC
-setup (#109) and is the live validation tracked in #105. Badges on the diagram mark the
+*can* reach are deliberately tiny: `ui-preview` holds a single custom role —
+`FoundryGate SWA Preview Publisher`, defined in `infra/modules/swa-preview-role.bicep` —
+assignable only on the dev Static Web App, and `dev-plan` was meant to be Reader-and-nothing-else
+for the PR what-if. That second one turned out to be impossible — ARM's what-if preflight
+authorizes every resource as a write, so no read-only identity can run it — and the PR what-if is
+switched off pending a decision (#229). A pull request runs its own copy of the workflow files,
+so the identity, not the workflow, is the boundary. The owner-side OIDC setup is **done for dev**
+(the [Owner Setup Runbook](/foundry-gate/reference/owner-setup/) is the commands that did it);
+**what has not happened yet** is the first full run against Azure, tracked in #105. Badges on the
+diagram mark the
 difference; the [CI/CD reference](/foundry-gate/reference/ci-cd/) lists every workflow, trigger
 and variable.
 
--- a/docs-site/src/content/docs/getting-started/fork-guide.mdx
+++ b/docs-site/src/content/docs/getting-started/fork-guide.mdx
@@ -44,13 +44,16 @@ This guide takes you from a blank Azure subscription to a running FoundryGate de
    - Value: `FoundryGate.Admin`
    - Allowed member types: Users/Groups
 
-   **Graph API permissions are granted later, not here.** The Entra sync calls Microsoft Graph as the
-   API's user-assigned managed identity (`id-foundrygate-api-{env}`, created by the Bicep deploy) — there is
-   no client secret. After the first infra deploy, grant that identity the Graph *application* roles
-   `Application.Read.All`, `User.Read.All` and `GroupMember.ReadBasic.All` as app-role assignments on the
-   Microsoft Graph service principal (the [Infrastructure Reference](/foundry-gate/reference/infrastructure/#role-assignments)
-   lists them; [#120](https://github.com/kolatts/foundry-gate/issues/120) has a PowerShell snippet). Then set
-   `Entra__Enabled=true` on the API.
+   **Graph API permissions are granted later, not here.** The Entra sync calls Microsoft Graph as a
+   user-assigned managed identity (created by the Bicep deploy) — there is no client secret. After the
+   first infra deploy, grant the Graph *application* roles `Application.Read.All`, `User.Read.All` and
+   `GroupMember.ReadBasic.All` as app-role assignments on the Microsoft Graph service principal to
+   **both** `id-foundrygate-api-{env}` **and** `id-foundrygate-func-{env}`. Both, because the nightly
+   `EntraSyncFunction` calls Graph as the *Functions* identity — grant only the API's and every nightly
+   run fails with `Authorization_RequestDenied` while the on-demand sync in the UI works fine, which is a
+   confusing way to find out. The [Owner Setup Runbook](/foundry-gate/reference/owner-setup/#6-graph-app-roles-for-the-managed-identities)
+   has the loop that does both; [#120](https://github.com/kolatts/foundry-gate/issues/120) has a
+   PowerShell equivalent. Then set `Entra__Enabled=true` on **both** hosts.
 
 3. ### Configure APIM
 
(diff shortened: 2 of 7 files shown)


kolatts:
**REQUEST CHANGES**

The setup itself is right — and the CI already proved half of it, since the PR what-if authenticated with the brand-new `dev-plan` OIDC identity on the first try. The docs are the problem: a runbook whose entire value proposition is "these commands actually ran, copy them" has a paragraph that describes a workflow it didn't read, and it leaves two directly contradicting statements alive in `handoff-orchestrator.md`.

---

**Issues**

- **Major** `docs-site/src/content/docs/reference/owner-setup.md` — "Nothing gets hardcoded into the repo" says `_deploy-ui.yml` writes `appsettings.json` from `FG_ENTRA_WEB_CLIENT_ID` / `FG_ENTRA_API_CLIENT_ID` / `FG_API_BASE_URL`. It doesn't: `Api.BaseUrl` is derived from the infra output `container-app-fqdn`, `AzureAd.Authority` comes from `vars.AZURE_TENANT_ID` (unmentioned, and sign-in breaks without it), and `FG_API_BASE_URL` is read only by the *preview* job in `ui-deploy.yml`. Wrong in the one doc that exists to be trusted verbatim.

- **Major** `handoff-orchestrator.md` §2 — still reads "Dev currently falls back to `SG_IMAGILE_SQL_ADMINS` (`2ed4d6b7-…`)", contradicting both this PR's `dev.bicepparam` change and the status block added 20 lines above it in the same file. It's the last occurrence of the old group anywhere in the repo.

- **Major** `handoff-orchestrator.md` §1 — still says "update `src/FoundryGate.Web/wwwroot/appsettings.json` placeholders", which the new runbook explicitly overturns. Someone following the handoff commits a tenant's client ids into a public repo.

- **Major** `docs-site/src/content/docs/reference/owner-setup.md` §4 + `reference/ci-cd.md` — the `dev-plan` identity is documented as Reader-only, and that is empirically false: the first run failed with `AuthorizationFailed` on `Microsoft.Resources/deployments/whatIf/action`. What-if is read-only in effect but a POST action in ARM, and no built-in role grants it. Filed as #229 with the custom-role fix; the docs need to stop prescribing the thing that doesn't work.

- **Minor** `docs-site/src/content/docs/getting-started/fork-guide.mdx` — grants the Graph roles to `id-foundrygate-api-{env}` only, which is precisely the mistake the new runbook §6 warns produces a nightly `Authorization_RequestDenied`. The getting-started layer now teaches the failure the reference layer documents.

- **Minor** `docs-site/src/content/docs/reference/owner-setup.md` §8 — `OUT="az deployment sub show …"` followed by `$($OUT.staticWebAppHostname.value -o tsv)` only works because of bash word-splitting. In zsh it isn't split at all, all four variables come back empty, and the two `az` calls after it run with an empty `--role`/`--scope`. Three saved lines is a bad trade in a copy-paste runbook.

- **Minor** `docs-site/src/content/docs/reference/owner-setup.md` — `ROLE_ID` means the `FoundryGate.Admin` app-role GUID in §1 and the SWA role *definition id* in §8. Work the runbook top-to-bottom in one shell and scrolling back to re-run the app-role grant silently posts the wrong `appRoleId`.

- **Minor** `docs-site/src/content/docs/reference/ci-cd.md`, `reference/infrastructure.md` — both still route the reader to bare issue #109 for the steps that now live in a published page nothing links to. The new doc is reachable only via the autogenerated sidebar.

- **Nit** `docs-site/src/content/docs/reference/owner-setup.md` §6 — a numbered runbook whose step 6 opens with "do this **after** step 7" is just step 8's fourth bullet wearing a hat.

- **Nit** `docs-site/src/content/docs/reference/owner-setup.md` §5 — `set-var()` is defined and then never called; the variables arrive as a prose table instead. Either invoke it or drop it.

- **Nit** `docs-site/src/content/docs/reference/owner-setup.md` §8 — uses bare `--assignee <client id>` immediately after §4 explains why `--assignee-object-id` + `--assignee-principal-type` is the form that doesn't race a fresh principal.

<sub>Reviewed with three parallel passes (security/correctness, complexity/duplication, CLAUDE.md compliance); findings below 70 confidence dropped. `.github/` untouched, `.md`-without-MDX invariant clean (Starlight asides are markdown-native, and `reference/api.md` already uses them), docs build green.</sub>

kolatts:
Fix pass pushed as `a660360`. Point by point.

**Major — `owner-setup.md` misdescribes `_deploy-ui.yml`.** Fixed, and it was worse than "incomplete": `Api.BaseUrl` is derived from the infra output `containerAppFqdn`, `AzureAd.Authority` comes from `vars.AZURE_TENANT_ID`, and `FG_API_BASE_URL` is read only by the preview job in `ui-deploy.yml`. Replaced the sentence with the actual four-row mapping of key → source, and called out that `AZURE_TENANT_ID` is therefore not optional on `dev`. `ci-cd.md`'s variable table now says so too.

**Major — `handoff-orchestrator.md` §2 stale `SG_IMAGILE_SQL_ADMINS`.** Fixed. Rewritten to say the dev group exists with its object id, and that **production** still needs its own — which is the part of that bullet that was still true and was getting lost behind the stale half.

**Major — `handoff-orchestrator.md` §1 "update `appsettings.json` placeholders".** Fixed, and inverted to an explicit **do not**, with the reason. Following that line would have committed a tenant's client ids into a public repo.

**Major — `dev-plan` is not Reader-able.** This one grew. The custom role got past `deployments/whatIf/action`, and then the job failed again with **thirteen** `Authorization failed for template resource …/write` errors — `az deployment sub what-if` runs ARM's full preflight, which authorizes every resource in the template *as a write*. There is no role, custom or built-in, that both runs this what-if and cannot change anything; the floor is roughly `Contributor`, which is precisely the blast radius #144 created `dev-plan` to avoid.

So the premise, not the role, was wrong. `dev-plan`'s three variables are **unset again**, which returns `_deploy-infra.yml` to its designed skip-with-`::notice::` instead of a permanent red check on every PR — briefly setting them is the only reason any of this surfaced. #229 retitled, relabelled `needs-human`, and carries four options (drop the PR what-if; give it Contributor and accept it; point it at a separate empty subscription; restrict it to protected branches). `ci-cd.md`, `architecture/overview.mdx`, the runbook and the handoff now all state the finding instead of repeating the assumption.

**Minor — `fork-guide.mdx` grants Graph roles to the API identity only.** Fixed in both places (step 2 and the `Authorization_RequestDenied` troubleshooting entry), with the symptom that identifies it: the on-demand sync in the UI works fine while every nightly run fails, because the nightly one calls Graph as the *Functions* identity. Both now link the runbook's §6 loop.

**Minor — the `$OUT` word-splitting trick.** Fixed; the four assignments are spelled out. You were right about the failure mode and I'd underrated it: in zsh there is no word splitting, so all four come back empty and the two `az` calls after them run with an empty `--role`/`--scope`. Left a sentence saying why the compact form isn't there, so nobody re-compacts it.

**Minor — `ROLE_ID` reused.** Fixed: the SWA one is `SWA_ROLE_ID` now, with a note about the collision.

**Minor — nothing links the new page.** Fixed: `ci-cd.md` ("Exact steps: #109" → the runbook), `infrastructure.md`'s operator-steps preamble, `fork-guide.mdx` ×2, and `overview.mdx`.

**Nit — §6 opens with "do this after step 7".** Fixed as far as it can be: it is a `:::note` explaining the dependency, and step 8 now has a fourth bullet pointing back at it. Kept in place rather than moved, because it belongs next to the other identity work — splitting the Graph grants away from the app registrations to satisfy the numbering would trade a real grouping for a cosmetic one.

**Nit — `set-var` defined but never called.** Fixed; the loops that actually set the variables are now in the doc.

**Nit — `--assignee` vs `--assignee-object-id`.** Fixed in §8.

Two things from the review passes I did **not** act on, for the record:

- **`fork-guide.mdx` describes one combined app registration and one shared CI identity** where the runbook has two and three. Real drift, but it is a rewrite of a getting-started page rather than a fix, and the fork story genuinely may want to stay simpler than the dev story. Not filed yet because I want the #229 decision first — it changes how many CI identities a fork needs.
- **`dev-destroy` protection rules absent from the runbook.** Added a sentence in §5 instead of a section, since the Environments already exist here with their rules and only a fork creates them fresh.

Docs build green (16 pages). `.github/` untouched.

Issue #232, opened 2026-09-05
GET /users/me advertises the deploy-time alias map, not the one the gateway now enforces


kolatts:
## Why

Found while building #225. There are now two sources for "which models may this tier use", and they
can disagree:

- **What the gateway enforces** is the APIM named value `fg-model-map-{tier}`, which
  `PUT /api/v1/gateway/tiers/{tier}/models` edits live (#225) and `/models` renders.
- **What a developer is told** on `GET /users/me` comes from `GatewayOptions.ModelAliases`
  (`Gateway__ModelAliases__{i}__*`), which `infra/modules/control-plane.bicep` emits from
  `productModelAliases` at deploy time — a snapshot of the last deploy (#153).

So an admin who grants `opus` to Standard through the UI changes what the gateway permits
immediately, while every Standard developer's "Configure your CLI" panel keeps listing the old set
until the next infra deploy — and an alias removed through the UI keeps being advertised to
developers whose requests it now refuses. That second direction is the worse one: we are telling
people to use a model we will 403.

## Options

1. `GET /users/me` reads the named value for the caller's tier through `IGatewayModelService`
   (cached — every developer's `/me` hits it) and falls back to `GatewayOptions.ModelAliases` when
   APIM is not configured. One source of truth, at the cost of an APIM read on the `/me` path.
2. Keep configuration as the read model and have the write path also update it — impossible without
   a redeploy, since it is an env var on the Container App. Not viable.
3. Persist the map in SQL and treat APIM as a projection. Biggest change; would also give the map a
   history the audit trail currently carries alone.

Option 1 looks right, with a short cache keyed by tier and invalidated by the PUT.

## Human attention

None; a design call between the options is the only judgement needed and (1) is recommended.

## Done when

- [ ] A developer's `/me` alias list reflects an allowlist change made through `/models` without a
      deploy.
- [ ] An alias removed from a tier stops being advertised to developers on it.
- [ ] The `/me` path does not gain an uncached per-request APIM call.
- [ ] `GatewayOptions.ModelAliases`'s remaining role (or its removal) is documented where it is
      declared.

Issue #226, opened 2026-09-05
Live validation: PUT fg-model-map named value against real APIM + a request routing through a new alias


kolatts:
> Parent: #220 — Live validation of the dev environment

## Why

#225 added `PUT /api/v1/gateway/tiers/{tier}/models`, which edits the APIM named value
`fg-model-map-{tier}` through the Management API so an admin can grant or retarget a model for a
quota tier without redeploying any policy. Everything above the APIM seam is covered by the
in-memory fake; the seam itself has never run against a real APIM instance, and plans/25 already
lists the named-value round trip as its one mechanism with no offline proof (the map JSON is
substituted into a `set-variable` attribute inside `infra/policies/model-alias-fragment.xml`, so
embedded quotes are the risk).

## Human attention

Needs a deployed gateway (APIM + at least one Foundry account) and a developer key.

## Done when

- [ ] `PUT /api/v1/gateway/tiers/{tier}/models` updates the `fg-model-map-{tier}` named value on a
      real APIM instance and returns 200 (the API identity holds API Management Service Contributor —
      `infra/modules/control-plane-rbac.bicep`).
- [ ] `GET /api/v1/gateway/tiers/{tier}/models` reads back exactly what was written, and the
      `deploymentExists` / `missingFromAccounts` fields agree with the live Foundry accounts.
- [ ] A request naming a **newly added** alias routes and completes through the gateway with no
      policy redeploy — the claim plans/25 makes and this endpoint depends on.
- [ ] Removing an alias from a tier makes that model `403 model_not_permitted` for a developer on
      that tier, within APIM's named-value propagation delay (record what that delay actually is).
- [ ] The map JSON survives `{{...}}` substitution into the policy's `set-variable` attribute after
      an API-driven write, not only after an ARM-driven one.
- [ ] APIM refusing a write (bad value, conflict) surfaces as the API's 409, not a 500.

kolatts:
## Live validation — the central claim holds; three defects filed

Against the real `apim-foundrygate-dev-e7k2`, through the deployed API with an admin token and a real developer key. The gateway calls are real completions against `gpt-4-1-mini` (#259). The `fg-model-map-standard` named value was captured before I started and **restored byte-identical afterwards** (`diff` → identical, and `gpt` still routes `200`).

### Done when

[...]

Issue #126, opened 2026-09-01
Support Anthropic (Claude) deployment creation via POST /foundry/deployments: SDK cannot send modelProviderData (1.5.2 and 1.6.0-beta.4) + #107 permissions


kolatts:
> Parent: #81 — Claude on Foundry end to end

Discovered implementing #61. `POST /api/v1/foundry/deployments` refuses `modelFormat: Anthropic` with `400` today (the contract already has `FoundryModelFormatType.Anthropic`; `FoundryDeploymentService.CreateDeploymentAsync` short-circuits before any ARM call). Two independent blockers:

1. **SDK gap.** `Azure.ResourceManager.CognitiveServices` 1.5.2 (current stable, api-version `2025-06-01`) has no `modelProviderData` on `CognitiveServicesAccountDeploymentProperties` — and neither does 1.6.0-beta.4 (api-version `2026-07-15-preview`, checked against the SDK's public API surface on `main`, 2026-09-01). Per E-005 (`fable-refactor-log.md`) ARM only honours `modelProviderData` on api-versions ≥ 2026-xx; the repo's Bicep pins `Microsoft.CognitiveServices/accounts/deployments@2026-07-01`. A typed PUT through the SDK therefore reaches ARM without the Marketplace attestation and is rejected (`InvalidModelProviderData`), and per E-007 a failed Claude create can wedge the subscription's Marketplace agreement — so the API must not even try.
2. **Permissions** — #107: the API identity holds Cognitive Services Contributor only, not the `Microsoft.SaaS/register/action`-class permissions the attestation needs.

## Options once #107 lands

- **Wait for the SDK** to model `modelProviderData` (track the `Azure.ResourceManager.CognitiveServices` changelog); then `ArmFoundryManagementClient.CreateDeploymentAsync` adds the block when `request.ModelFormat == Anthropic` and the service drops the refusal.
- **Bypass the typed model now**: `ArmClient.GetGenericResources().CreateOrUpdateAsync` with `GenericResourceData.Properties` as raw JSON (`model` + `modelProviderData`) and `ArmClientOptions.SetApiVersion("Microsoft.CognitiveServices/accounts/deployments", "2026-07-01")`. Keeps the same `IFoundryManagementClient` seam; needs a live test (no fake can prove the attestation shape).

[...]

PR #236, opened 2026-09-05
Admin UI: provision models and manage allowed models per tier

### LINKED ISSUE #225: Admin UI: provision models and manage allowed models per tier
## Why

The owner's ask: "We also want to make it easy to provision new models and set allowed models in the
front end of the app."

Today both halves are file edits, not app operations:

- **Allowed models per tier.** The per-tier alias map (`{ alias: { deployment, backend, provider } }`)
  lives in `infra/main.bicep`'s `productModelAliases` and becomes the APIM named values
  `fg-model-map-{tier}`. The map **is** the allowlist (#86): an alias a tier does not list is a
  `403 model_not_permitted` at the gateway. plans/25 always intended the control plane to edit those
  named values through the APIM Management API — no policy redeploy — but the endpoint was left as
  "later, with #82". Granting a tier a model therefore means editing bicep and running an infra deploy.
- **Provisioning a model.** `/foundry` can create exactly one deployment in exactly one account per
  dialog trip, and gives no follow-through: no poll to a terminal state, and no path from "the
  deployment exists" to "a tier is allowed to use it".

## Sub-tasks

- [ ] **Allowed models per tier (API).** `GET /api/v1/gateway/tiers`,
      `GET /api/v1/gateway/tiers/{tier}/models`, `PUT /api/v1/gateway/tiers/{tier}/models` (full
      replace) over the `fg-model-map-{tier}` named value via the APIM Management API. Admin-only,
      audited (`gateway.models.updated`, before/after), validated: url-safe lowercase alias,
      `provider` in {anthropic, openai}, deployment must exist in at least one configured Foundry
      account — and in **every** pooled account when `pool = anthropic`, mirroring the main.bicep
      contract (a region missing the deployment turns a pool failover into a 404). Backend resolved
      from `pool` exactly as `infra/modules/ai-gateway.bicep` does.
- [ ] **Allowed models per tier (UI).** New admin page `/models` — "Models & access": a tier x alias
      overview, per-tier add / retarget / remove, and a "what developers see" preview of the aliases
      each tier permits. Linked in the nav beside Foundry deployments. bUnit + service tests.
- [ ] **Easy model provisioning.** `/foundry`'s dialog gains multi-account targeting (primary by
      default; every pooled account for a model that needs to exist in all of them), polls each new
      deployment to a terminal state, keeps the Claude/Anthropic refusal as an inline explanation
      linking #126/#107 rather than a generic error, and offers "add it to a tier's allowed models"
      afterwards, deep-linking into `/models` pre-filled.
- [ ] **Docs at all three levels** (CLAUDE.md invariant 2): `getting-started/why-foundrygate` (one
      plain sentence), `architecture/overview` (short paragraph + the page), `reference/web-ui.md`
      and `reference/api.md` (precise endpoints and page behaviour). Landing page re-read; touched
      only if a claim became false.

## Human attention

None for the implementation. Live verification against a real APIM instance — PUT of an
`fg-model-map-{tier}` named value, and a real request routing through a newly added alias — needs a
deployed gateway and is tracked separately under the `live-validation` milestone.

## Done when

- The three endpoints exist, are admin-only, are audited, and are covered by endpoint + service tests
  against the in-memory APIM fake.
- `/models` lists every tier's allowed models, and add / retarget / remove each round-trip through the
  named value.
- `/foundry` can provision a model into one or several accounts, polls it to a terminal state, and
  offers the "allow it for a tier" follow-through.
- Docs updated at all three levels; `npm run build` in `docs-site/` passes.
- `dotnet build -c Release` zero warnings, `FoundryGate.Tests.Predeployment` green,
  `dotnet format --verify-no-changes` clean.

kolatts:
Closes #225.

The owner's ask: *"We also want to make it easy to provision new models and set allowed models in the
front end of the app."* Both halves were file edits before this — granting a tier a model meant
editing `infra/main.bicep` and running an infra deploy, and `/foundry` could create one deployment in
one account with no follow-through.

## Allowed models per tier

The per-tier alias map **is** the allowlist (#86): each quota-tier product carries the APIM named
value `fg-model-map-{tier}` holding `{ alias: { deployment, backend, provider } }`, applied by a
policy fragment ahead of `llm-token-limit`, and anything a tier does not list is
`403 model_not_permitted` before it costs any quota. plans/25 always intended the control plane to
edit those named values through the Management API — no policy redeploy — and left it as "later, with
#82". This is that.

- `IApimManagementClient` (Core) gains `GetNamedValueAsync` / `SetNamedValueAsync`. The read goes
  through APIM's `listValue`, not the resource GET, because ARM omits `properties.value` for a secret
  named value and depending on which shape comes back would make the read right for only half the
  contract.
- `GET /api/v1/gateway/tiers`, `GET|PUT /api/v1/gateway/tiers/{tier}/models` — admin-only, audited
  (`gateway.models.updated`, target type `GatewayTier`, details carrying the whole map before/after).
  `PUT` is a full replace, because the named value is one document the gateway reads whole.
- `GatewayModelMap` (Domain) holds the naming contract shared with
  `infra/modules/ai-gateway.bicep`: the named-value name, and the pool↔backend mapping in both
  directions, so a map written by the API and a map written by a deploy are indistinguishable.
- Validation refuses maps the gateway would answer with a **404** rather than an honest 403: alias
  grammar (lower-case, url-safe — the policy compares without normalizing), no duplicate alias, a
  known pool, a deployment that exists in a configured account — and in **every** account for an
  `anthropic`-pool alias, mirroring the contract stated in `infra/main.bicep` (the pool fails a
  throttled request over to another region, and a region missing the deployment turns a 429 into a
  404). APIM refusing the write is a `409` carrying what APIM said. Writing the map a tier already
  holds is a no-op: no APIM call, no audit row.
- New admin page **`/models`** ("Models & access"), linked in the nav beside Foundry deployments: an
  alias × tier overview, per-tier allow / retarget / remove, and a "what developers on this tier see"
  preview. `provider` and `pool` are *derived* from the chosen deployment's ARM `model.format` rather
  than asked — getting either wrong produces a failure that looks like something else.

## Easier model provisioning

- The provision dialog can name several accounts ("Also create in"): still one
  `POST /foundry/deployments` per account, run in order and reported separately, so a second region
  failing never reads as the first one having failed too (E-007: never a loop the API drives on the
  admin's behalf). Which accounts a model belongs in is `infra/main.bicep`'s pooled / primary-only
  split, now expressed once in `ModelAliasDerivation` so it stays true when #126 lifts.
- After a create the page polls each new deployment to a terminal state, refreshing the grid, and
  gives up after ~1 minute rather than spinning.
- The Claude/Anthropic refusal stays an inline explanation with the issue links, not a generic error.
- A successful create ends with a link into `/models?deployment=…`, which that page turns into a
  pre-filled "allow a model" dialog — a deployment nobody may use is not yet useful.

## Docs

All three levels per CLAUDE.md invariant 2: one plain sentence in the `why-foundrygate` comparison
table, a paragraph in `architecture/overview` under APIM enforcement, and precise endpoint/page
behaviour in `reference/api.md` and `reference/web-ui.md`. Landing page re-read — no claim became
false, so it is untouched. plans/25's control-plane checkbox is now ticked.

## Verification

- `dotnet test src/FoundryGate.Tests.Predeployment -c Release` — **1599 passed** (1545 before)
- `npm ci && npm run build` in `docs-site/` — green

## Follow-ups filed

- **#226** (`live-validation`) — PUT of an `fg-model-map` named value against real APIM, and a real
  request routing through a newly added alias. The named-value round trip is the one mechanism here
  with no offline proof.
- **#232** — `GET /users/me` still advertises the *deploy-time* alias map (`Gateway__ModelAliases__*`)
  while the gateway now enforces the named value, so the two can diverge until the next deploy. Found
  while building this; the worse direction is advertising a model we will now 403.

### DIFF
--- a/docs-site/src/content/docs/architecture/overview.mdx
+++ b/docs-site/src/content/docs/architecture/overview.mdx
@@ -169,6 +169,21 @@ product's policy carries that tier's budget and TPM cap, and changing a develope
 quota means moving their subscription between products. Subscription *suspension* is
 reserved for deactivation and offboarding — not quota exhaustion.
 
+**Which models a tier may use is enforced in the same place.** Each tier product carries a
+model alias map — an APIM named value holding `{ alias: { deployment, backend, provider } }`
+— and a policy fragment that runs *before* `llm-token-limit` rewrites the alias a developer
+pinned (`sonnet`, `haiku`, `gpt`) to the real Foundry deployment and routes it at the right
+backend. The map **is** the allowlist: an alias a tier does not list is refused with
+`403 model_not_permitted`, ahead of the quota policy, so a blocked model costs the developer
+nothing. Because it lives in a named value rather than in policy text, admins edit it from
+the **Models &amp; access** page (`/models`) through the APIM Management API — granting a tier
+a model, or re-pointing an alias at a newer deployment, takes effect without redeploying a
+policy or an infrastructure template. The API validates before it writes: an alias must be
+lower-case and url-safe, and it must point at a deployment the configured Foundry accounts
+actually have — in *every* account when it routes at the multi-region Anthropic pool, because
+the pool sends a throttled request to another region and a region missing the deployment would
+turn a `429` into a `404`.
+
 The usage-sync Function still runs, but as **reconciliation**, not enforcement. It is a
 .NET isolated worker on a Flex Consumption plan — scale-to-zero, so it costs nothing between
 ticks — reading the gateway's own log with the Function App's managed identity. The query it
--- a/docs-site/src/content/docs/getting-started/why-foundrygate.mdx
+++ b/docs-site/src/content/docs/getting-started/why-foundrygate.mdx
@@ -69,6 +69,7 @@ That works fine when you have three engineers sharing a dev environment. It brea
 | Quota increase workflow | ❌ None | ✅ Submit → admin approves → quota applied |
 | Per-developer usage visibility | ❌ Aggregate only | ✅ Dashboard + audit log per user |
 | Model deployment management | ❌ Portal only | ✅ Claude models are set up by the infrastructure deploy; admins manage OpenAI model deployments from the FoundryGate UI |
+| Which models a team may use | ❌ Everything on the deployment, or nothing | ✅ Admins choose which models each budget tier is allowed to use, from the FoundryGate UI — anything else is refused at the gateway |
 | Deployment-level rate limits (TPM/RPM) | ⚠️ Shared ceiling, 429s passed to callers | ⚠️ Mitigated — backend pools + circuit breakers reroute around saturated deployments; the platform ceiling itself remains |
 
 ## How it works
--- a/docs-site/src/content/docs/reference/api.md
+++ b/docs-site/src/content/docs/reference/api.md
@@ -481,6 +481,30 @@ one, and every open of the create dialog asks for it. Nothing invalidates it ear
 does not change what is deployable. An **empty** answer is not cached at all, so a transient window
 where every account 404s cannot pin "catalogue unavailable" on the dialog for five minutes.
 
+## Gateway — model allowlist
+
+Each quota-tier product carries a **model alias map**: an APIM named value `fg-model-map-{tier}` holding `{ alias: { deployment, backend, provider } }`, which a policy fragment applies *before* `llm-token-limit`. It rewrites the alias a developer pinned to the real Foundry deployment and routes it at the mapped backend. **The map is the allowlist** — an alias a tier does not list is refused `403 model_not_permitted` ahead of the quota policy, so a blocked model costs the developer no quota, and a tier with no map permits nothing at all (fail loud, by design).
+
+These endpoints edit that named value through the APIM Management API, so a change takes effect **without redeploying a policy or an infrastructure template**.
+
+| Method | Path | Auth | Description |
+|---|---|---|---|
+| `GET` | `/gateway/tiers` | Admin | The gateway's quota tiers with the size of each one's allowlist: `{ tier, displayName, monthlyTokenQuota, isUnlimited, allowedModelCount }`. Tier identity and caps come from the same `Gateway:Tiers` table quota resolution uses; the counts come from APIM. |
+| `GET` | `/gateway/tiers/{tier}/models` | Admin | One tier's allowlist as APIM holds it: `{ tier, displayName, aliases: [{ alias, deploymentName, pool, provider, deploymentExists, missingFromAccounts }] }`, ordered by alias. `deploymentExists` and `missingFromAccounts` are checked live against the configured Foundry accounts, so a map that would 404 is visible before a developer finds it. `404` for an unknown tier. |
+| `PUT` | `/gateway/tiers/{tier}/models` | Admin | Replace a tier's allowlist in full. Body: `{ "aliases": [{ "alias", "deploymentName", "pool", "provider" }] }`. Returns the tier's allowlist as stored. An empty list is legal and permits nothing. |
+
+`PUT` is a **full replace**, not a patch: the named value is one JSON document and the gateway reads it whole, so add, retarget and remove are all the same call with a different list.
+
+Rules it enforces before touching APIM:
+
+- **`alias` is lower-case and url-safe** (`^[a-z0-9][a-z0-9._-]*$`) — it travels in a request's `model` field and the policy compares it without normalizing, so `Sonnet` would silently miss an allowlist that lists `sonnet`. `400` otherwise.
+- **An alias may appear once.** The gateway's map is keyed by alias, so a duplicate would let list order decide which deployment a developer actually reaches. `400`.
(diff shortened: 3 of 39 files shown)


kolatts:
## `sunny:code-review`

**REQUEST CHANGES**

The API half is genuinely good — the validation refuses exactly the maps that would 404 instead of 403, and the named-value contract is pinned to the bicep in one place instead of being retyped. The UI half has a poll loop that awaits inside the event handler and never asks for a render, so the "watch it provision" feature watches nothing.

---

**Issues**

- **Major** `src/FoundryGate.Web/Pages/Foundry.razor` ~261–299 — the poll loop is awaited inside the click handler and never calls `StateHasChanged`, so Blazor renders nothing until it returns: for up to a minute the grid doesn't refresh, the `foundry-polling` alert never appears, and `_busy` keeps every button disabled. That's the opposite of what the docs now promise.

- **Minor** `src/FoundryGate.Web/Pages/GatewayModels.razor` ~272–282 — `LoadAsync` walks the tiers serially, and each `GET /gateway/tiers/{tier}/models` re-enumerates every Foundry account server-side. Three tiers is four full ARM enumerations to open a page. What does this look like on a fork with six tiers?

- **Minor** `src/FoundryGate.Api/Services/Gateway/GatewayModelService.cs` 54–69 — `ListTiersAsync` reads every tier's whole map from APIM to return a count the page then immediately re-fetches per tier. Every map fetched twice per page load.

- **Minor** `src/FoundryGate.Web/Pages/Foundry.razor` ~280–296 — each poll iteration does a per-deployment `GET` *and* a full `LoadAsync`, even when nothing moved. The per-deployment response already carries the state the grid renders.

- **Minor** `src/FoundryGate.Domain/Gateway/Contracts/GatewayContracts.cs` — `ReplaceTierModelsRequest` / `TierModelAliasRequest` aren't in `RequestDtoBindingTests`. CONVENTIONS.md: *"posts each request record to its own production endpoint through the real pipeline and asserts 400, never 500."* The table is hand-maintained, so nothing else catches this.

- **Nit** `src/FoundryGate.Core/Gateway/ArmApimManagementClient.cs` ~213 — `ifMatch: ETag.All` on a *create-or-update* sends `If-Match: *`, which ARM reads as "must already exist". Harmless while the bicep always creates the named value first; a surprising 412-turned-409 on any fork where it hasn't.

Also worth a thought, not blocking: `ReadMapAsync` deliberately treats an unreadable map as an empty allowlist, but an entry that parses with a missing `deployment` field walks straight into a dictionary lookup with a null key.

kolatts:
Fair on all six. Point by point, all fixed in `393fe84`.

**Major — poll loop renders nothing.** Correct, and it made the feature a lie: the whole loop was awaited inside the click handler, which is exactly the case where Blazor renders only on return. Fixed three ways — the loop is now started *outside* the busy block (waiting for ARM is not a reason to disable the page for a minute), it calls `StateHasChanged` when it starts and when it finishes, and `_busy` is released before it begins. Also captured `_pageCts.Token` once at the top: `Dispose()` cancels *and* disposes, and reading `.Token` afterwards throws `ObjectDisposedException` rather than the cancellation the loop is waiting for.

**Minor — `/models` 1 + N sequential calls.** Two changes. The page now fires the per-tier reads with `Task.WhenAll` instead of walking them, and the expensive half — `ReadDeploymentPlacementAsync`, the full ARM enumeration behind "does this deployment exist" — is cached for 15 seconds in `GatewayModelService`. Six tiers is now one ARM enumeration for the page, not seven.

**Minor — `ListTiersAsync` fetches every map twice.** Same 15-second cache, keyed per tier, so the counts read and the rows read share one APIM call. `Reading_a_tier_twice_asks_APIM_once` locks it in. A successful `PUT` replaces its own cache entry rather than waiting it out — an admin who allows a model and then sees the old list would reasonably conclude the write failed — and `A_write_replaces_what_the_next_read_serves` covers that. 15 seconds is short enough that an edit made in the portal shows up while the admin is still looking at the page.

**Minor — poll loop refetches the whole grid every tick.** Now only when a deployment actually reached a terminal state. Twenty ticks of "nothing moved" cost twenty small GETs instead of twenty multi-account ARM listings.

**Minor — new request records missing from `RequestDtoBindingTests`.** Added two rows: a non-lower-case alias and an empty deployment name, both asserting the nested member name comes back in the 400.

**Nit — `ifMatch: ETag.All`.** Removed, with a comment saying why. You're right that it reads as "must already exist"; the bicep does create the named value first today, but a fork that adds a tier without redeploying would have hit a 412 turned into a 409 that named the wrong problem.

**On the non-blocking note (null `deployment`).** Real. `ReadMapAsync` now drops entries missing a deployment name, which is exactly what the policy does with them (plans/25: "absent, `null`, or missing any of the three fields = blocked") — so the tier's row count matches what the gateway will actually permit, and the projection never looks a null key up. `A_half_written_map_entry_is_dropped_rather_than_rendered` covers it.

One thing I did **not** do: the docs-site architecture diagram components. CONVENTIONS.md does ask every architecture-changing wave to update them, but another agent is reworking those visuals right now and touching them here would be a guaranteed conflict for no reader benefit today. The prose in `architecture/overview.mdx` is current.

Gates after the fixes: build 0 warnings, **1604 tests green** (1545 before this PR), `dotnet format --verify-no-changes` clean, docs build green.

kolatts:
Merging with `--admin`. `build-test`, `docs-build` and `Build Astro site` are green; the only red check is **UI - PR preview**, which dies at `azure/login` with `No subscriptions found` for the `ui-preview` OIDC principal — before it touches the diff. That is an Azure credential/role problem introduced somewhere between 2026-09-02 (last green run of that workflow) and today's owner-setup pass, not something this branch can fix. Filed as #238 (`needs-human`, v0.6).

PR #240, opened 2026-09-05
fix(infra): APIM product-API links must be products/apis, not apiLinks

### LINKED ISSUE #239: APIM products/apiLinks is not idempotent — every infra re-run fails with Conflict
## The gateway template is single-use. Every re-run of `main.bicep` fails.

Found by the first real `Deploy All → dev` (#105) — specifically by the *second* run of it, which is the first time this template has ever been applied twice to the same subscription.

```
ResourceDeploymentFailure  …/deployments/foundrygate-ai-gateway
  Conflict — "Link already exists between specified Product and Api."   × 4
```

Runs: [33972516153](https://github.com/kolatts/foundry-gate/actions/runs/33972516153) and [33972522833](https://github.com/kolatts/foundry-gate/actions/runs/33972522833) — two independent runs, sequential (not a race), identical failure.

## Cause

`infra/modules/ai-gateway.bicep` links each tier product to the two front-door APIs with **`Microsoft.ApiManagement/service/products/apiLinks`**:

```bicep
resource tierProductAnthropicLinks 'Microsoft.ApiManagement/service/products/apiLinks@2024-06-01-preview' = [
  for (tier, i) in quotaTiers: {
    parent: tierProducts[i]
    name: 'anthropic-link'
    properties: { apiId: anthropicApi.id }
  }
]
```

`apiLinks` **is not idempotent**. Its ARM identity is the *link name* (`anthropic-link`), but APIM separately enforces uniqueness on the **(product, api) pair**. So a re-PUT of an unchanged link is not seen as "this resource already exists, no-op" — it is seen as "you are trying to create a second link for a pair that already has one", and refused with `Conflict`. There is no `properties` change that makes it converge; the resource simply cannot be applied twice.

That breaks the invariant in `CLAUDE.md`: *"Infra lives in `infra/` … re-runs must pass `createModelDeployments=false`."* It also means the 2026-09-01 gateway live-validation could not have caught it — that stack was deployed once and torn down, never re-run.

## Fix

The classic child resource `Microsoft.ApiManagement/service/products/apis` names the association by the **API itself**, so a re-PUT is a real no-op.

```bicep
resource tierProductAnthropicApis 'Microsoft.ApiManagement/service/products/apis@2024-06-01-preview' = [
  for (tier, i) in quotaTiers: {
    parent: tierProducts[i]
    name: anthropicApi.name
  }
]
```

Verified live against the deployed `apim-foundrygate-dev-e7k2` (BasicV2), on a pair that **already had a link** — the exact case that 409s under `apiLinks`:

```
PUT …/products/standard/apis/foundrygate-anthropic?api-version=2024-06-01-preview   -> 200
PUT …/products/standard/apis/foundrygate-anthropic?api-version=2024-06-01-preview   -> 200   (again)
```

Two consecutive PUTs, both 200, no conflict. Same underlying association — `apiLinks` is a newer surface over the same relationship, so switching does not orphan anything or change what the gateway serves; the existing `anthropic-link` / `openai-link` entities keep describing the association the new resource now owns.

## Why this matters beyond the immediate red build

A template that can only be applied once is worse than a template that fails loudly, because it looks fine on day 0 and fails on the first fix you try to ship. Every downstream stage — database, API, Functions, UI, postdeployment tests — sits behind the infra stage, so this one conflict skips the entire pipeline.

It is also a general warning about `-preview` APIM resource types: `apiLinks` reads like the modern replacement for `products/apis` and is what the portal emits, but it carries a uniqueness rule its own resource identity does not express.

## Done when

- [ ] `ai-gateway.bicep` uses `products/apis`
- [ ] A `Deploy All → dev` run gets past `foundrygate-ai-gateway` on a subscription that already has the links
- [ ] A second consecutive infra run is a no-op (this is also an unticked box on #105)

kolatts:
## What

The gateway template could only ever be applied **once**. Every re-run of `main.bicep` died in the `foundrygate-ai-gateway` module with four of:

\\\
Conflict — "Link already exists between specified Product and Api."
\\\

`products/apiLinks` is not idempotent. Its ARM identity is the *link name* (`anthropic-link`), but APIM separately enforces uniqueness on the **(product, api) pair** — so a re-PUT of an unchanged link is not read as "already exists, no-op" but as "you are creating a second link for a pair that already has one". No property change makes it converge.

`products/apis` names the association by the API itself, so a re-PUT is a genuine no-op.

## Verification — live, on the pair that actually 409s

Against the deployed `apim-foundrygate-dev-e7k2` (BasicV2), on a product/API pair that **already had a link**:

\\\
PUT .../products/standard/apis/foundrygate-anthropic?api-version=2024-06-01-preview  -> 200
PUT .../products/standard/apis/foundrygate-anthropic?api-version=2024-06-01-preview  -> 200
\\\

Two consecutive PUTs, both 200. `az bicep build` clean; generated architecture model unchanged (`npm run generate:architecture` produces no diff).

`apiLinks` is a newer surface over the same underlying relationship, so switching orphans nothing and the gateway serves exactly what it served before.

## Why it survived to now

The 2026-09-01 gateway live-validation deployed the stack **once** and tore it down, so a second apply had never happened. It took the first real `Deploy All → dev` to find it — on the second run, [33972516153](https://github.com/kolatts/foundry-gate/actions/runs/33972516153) and [33972522833](https://github.com/kolatts/foundry-gate/actions/runs/33972522833), sequential and identical, so not a race.

Worth internalising as a general warning about `-preview` APIM types: `apiLinks` reads like the modern replacement for `products/apis` and is what the portal emits, but it carries a uniqueness rule its own resource identity does not express.

Closes #239
Refs #105

### DIFF
--- a/infra/modules/ai-gateway.bicep
+++ b/infra/modules/ai-gateway.bicep
@@ -363,23 +363,24 @@ resource tierProducts 'Microsoft.ApiManagement/service/products@2024-06-01-previ
   }
 ]
 
-resource tierProductAnthropicLinks 'Microsoft.ApiManagement/service/products/apiLinks@2024-06-01-preview' = [
+// Which APIs each tier product exposes. Deliberately `products/apis` and NOT the newer
+// `products/apiLinks`: `apiLinks` is not idempotent. Its identity is the link name, but
+// APIM additionally enforces uniqueness on the (product, api) PAIR, so a second PUT of an
+// already-linked pair fails with `Conflict — Link already exists between specified Product
+// and Api` even though nothing changed. That makes the whole template single-use, which
+// broke the first re-run of the dev deploy (#239). `products/apis` names the association by
+// the API itself, so a re-PUT is a genuine no-op. Verified against APIM BasicV2, 2026-09-05.
+resource tierProductAnthropicApis 'Microsoft.ApiManagement/service/products/apis@2024-06-01-preview' = [
   for (tier, i) in quotaTiers: {
     parent: tierProducts[i]
-    name: 'anthropic-link'
-    properties: {
-      apiId: anthropicApi.id
-    }
+    name: anthropicApi.name
   }
 ]
 
-resource tierProductOpenaiLinks 'Microsoft.ApiManagement/service/products/apiLinks@2024-06-01-preview' = [
+resource tierProductOpenaiApis 'Microsoft.ApiManagement/service/products/apis@2024-06-01-preview' = [
   for (tier, i) in quotaTiers: {
     parent: tierProducts[i]
-    name: 'openai-link'
-    properties: {
-      apiId: openaiApi.id
-    }
+    name: openaiApi.name
   }
 ]
 


kolatts:
**APPROVE**

Right diagnosis and the least-clever fix available. The one thing worth checking on a change like this is whether swapping `properties.apiId` for `name:` loses the implicit ARM dependency on the API — it does not, and I compiled it rather than reasoned about it:

\\\json
"dependsOn": [
  "[resourceId('Microsoft.ApiManagement/service/apis', parameters('apimName'), variables('anthropicApiName'))]",
  "[resourceId('Microsoft.ApiManagement/service/products', parameters('apimName'), parameters('quotaTiers')[copyIndex()].name)]"
]
\\\

Bicep's dependency analysis keys off any symbolic reference in the declaration, `name:` included, so both the API and the tier product are still ordered ahead of the association. A fresh day-0 deploy orders identically to before.

Also verified: `.name` resolves to the single last segment (both APIs are declared with `parent: apim`, so it is `foundrygate-anthropic`, not `apim/foundrygate-anthropic`) — the compiled name is a correct three-segment `format()`. Nothing else in the repo referenced the removed symbols or the `apiLinks` type: no outputs, no `dependsOn`, and `generate-architecture.mjs` only reads `bicepVar()` names out of this file, never its resource declarations, so the generated diagram is untouched. `GatewayInfraBindingTests` reads `control-plane.bicep` only.

**Issues**

- **Nit** `infra/modules/ai-gateway.bicep` 373, 380 — the old `anthropic-link` / `openai-link` entities stay in dev's APIM after this deploys, since incremental mode never deletes what the template drops. Harmless (they describe the same association the new resource owns), but a `what-if` on dev won't show them going away and the next person may wonder. Not worth a cleanup step.

- **Nit** `infra/modules/ai-gateway.bicep` 373, 380 — latent behaviour difference on an API *rename*: the link name used to be the constant `anthropic-link`, so a rename updated it in place; now the association is named after the API, so a rename creates a new one. Both names are hardcoded vars today, so it cannot bite yet.

Neither blocks. The empirical two-PUTs-both-200 in the description is the right kind of evidence for this repo — it tests the exact case that 409s rather than the happy path.

PR #242, opened 2026-09-05
fix(infra): decouple the SQL server's region — eastus2 is closed to new servers

### LINKED ISSUE #241: Azure SQL: eastus2 and eastus are closed to new servers — dev SQL moves to centralus
## What happened

The Deploy All → dev run that finally got past the APIM fix (#239) died one module later, in oundrygate-cp-sql:

\\\
Microsoft.Sql/servers/sql-foundrygate-dev-e7k2 -> Failed
  ProvisioningDisabled: "Provisioning is restricted in this region. Please choose a
  different region. For exceptions to this rule please open a support request with Issue
  type of 'Service and subscription limits'."
\\\

[Run 33973324629](https://github.com/kolatts/foundry-gate/actions/runs/33973324629).

## The part worth knowing

**This is invisible until you try.** It is not a quota you can read, not a SKU availability query, not anything `az sql server list-usages` or the SKU APIs will tell you. Azure closes individual regions to *new* Azure SQL logical servers as a capacity-management action, and the only way to discover it is to attempt a create and read the error.

Probed directly on this subscription, 2026-09-05:

| Region | Result |
|---|---|
| astus2 | RegionDoesNotAllowProvisioning — "not accepting creation of new Windows Azure SQL Database servers at this time" |
| astus | RegionDoesNotAllowProvisioning — same |
| centralus | **created successfully** (probe server deleted afterwards) |

So the primary region *and* its obvious neighbour are both shut, which is why this could not be worked around by nudging location.

Existing servers in a closed region keep running; only creation is blocked. It can reopen without notice, and it can close again.

## Fix

A sqlLocation parameter, defaulting to location, so the SQL server's region is decoupled from the rest of the stack without changing anything else's placement:

- main.bicep — new param sqlLocation string = location, threaded to the control plane
- modules/control-plane.bicep — same param, passed to the sql module instead of location
- modules/sql.bicep — comment explaining that location here is deliberately not the primary region
- parameters/dev.bicepparam — param sqlLocation = 'centralus'

Dev accepts the cross-region hop between the Container App (eastus2) and SQL (centralus). It is a real latency cost and it is the right trade for dev; production should set sqlLocation explicitly once a region is confirmed open, rather than inheriting location and discovering this on its own day 0.

## The alternative, not taken

Opening a support request (Issue type *Service and subscription limits*) to get eastus2 unblocked is the other path, and is what you would do if co-locating mattered. Not worth blocking the first dev deploy on a support turnaround.

## Done when

- [ ] sqlLocation exists and dev deploys SQL to centralus
- [ ] prod.bicepparam sets sqlLocation explicitly rather than inheriting
- [ ] 
eference/infrastructure documents the parameter and *why* it exists
- [ ] Revisit if eastus2 reopens — co-locating would remove a real cross-region hop

kolatts:
## What

The dev deploy got past the APIM fix (#239) and died one module later, in `foundrygate-cp-sql`:

\\\
Microsoft.Sql/servers/sql-foundrygate-dev-e7k2 -> Failed
  ProvisioningDisabled: "Provisioning is restricted in this region."
\\\

New `sqlLocation` param, defaulting to `location`, threaded `main.bicep` → `control-plane.bicep` → `sql.bicep`. Dev overrides it to `centralus`; everything else stays in `eastus2`.

## The part worth knowing

**This is invisible until you try.** Not a quota you can read, not a SKU availability query, not anything `az sql server list-usages` reports. Azure closes individual regions to *new* SQL logical servers as a capacity action and the only signal is attempting a create.

Probed directly on this subscription, 2026-09-05:

| Region | New SQL server |
|---|---|
| `eastus2` | refused — `RegionDoesNotAllowProvisioning` |
| `eastus` | refused |
| `centralus` | **created** (probe server deleted afterwards) |

The primary region *and* its obvious neighbour are both shut, so nudging `location` was not an option. Existing servers in a closed region keep working; only creation is blocked, and it can reopen without notice.

## Trade accepted

Dev now has a cross-region hop between the Container App (eastus2) and SQL (centralus). That is a real latency cost and it is the right call for dev, where the alternative is blocking the first deploy on an Azure support turnaround (Issue type *Service and subscription limits*).

`prod.bicepparam` deliberately gets **no** value here. Production should set `sqlLocation` explicitly after confirming a region is open, rather than inheriting `location` and rediscovering this on its own day 0 — that is a checkbox on #241.

## Verification

- `az bicep build-params --file infra/parameters/dev.bicepparam` renders `"sqlLocation": { "value": "centralus" }`; template builds clean.
- Generated architecture model unchanged (`npm run generate:architecture` — no diff).
- `npm run build` in `docs-site/` green, 17 pages.
- `reference/infrastructure` gains the parameter row and a "Why `sqlLocation` exists" section with the probe table, so the next person does not re-derive it from a red deploy.

Closes #241
Refs #105

### DIFF
--- a/docs-site/src/content/docs/reference/infrastructure.md
+++ b/docs-site/src/content/docs/reference/infrastructure.md
@@ -122,6 +122,7 @@ workflows resolve resources from these patterns — changing one is a contract c
 | `deployControlPlane` | `true` | `true` | default `false` (gateway only) |
 | `appEnvironment` | `qa` | `prod` | `ASPNETCORE_ENVIRONMENT` — lowercase `qa`/`prod`; `local` is docker-only |
 | `sqlAdminGroupObjectId` / `sqlAdminGroupName` | `SG_FOUNDRYGATE_SQL_ADMINS` | `$FG_SQL_ADMIN_GROUP_OBJECT_ID` / `$FG_SQL_ADMIN_GROUP_NAME` (required) | SQL server administrator (Entra-only auth; no SQL login exists) |
+| `sqlLocation` | `centralus` | set it explicitly | region for the SQL logical server, defaulting to `location`. Dev overrides it because **`eastus2` and `eastus` are both closed to new Azure SQL servers** on this subscription — see below |
 | `sqlDatabaseSku` | `GP_S_Gen5` ×1 — serverless, 60-min auto-pause | `GP_Gen5_2`, provisioned | serverless is derived from the SKU name (`GP_S_*`) |
 | `sqlBackupStorageRedundancy` | `Local` | `Geo` | |
 | `sqlZoneRedundant` | `false` | `true` | survives the loss of one availability zone without a restore; adds ~60% to the SQL compute meter (see [Cost & capacity](/foundry-gate/reference/cost-and-capacity/)) |
@@ -143,6 +144,33 @@ workflows resolve resources from these patterns — changing one is a contract c
 Nothing in a parameter file is a secret: SQL is Entra-only, storage is identity-based,
 registry pulls are identity-based, and the App Insights connection string is a module
 output. Entra object ids and client ids are identifiers, not credentials.
+
+### Why `sqlLocation` exists
+
+Azure closes individual regions to **new** Azure SQL logical servers as a capacity action,
+and there is no way to see it coming: it is not a quota you can read, not a SKU
+availability query, not anything `az sql server list-usages` reports. The only signal is
+attempting a create and getting `ProvisioningDisabled` / `RegionDoesNotAllowProvisioning`.
+
+Probed on the dev subscription, 2026-09-05:
+
+| Region | New SQL server |
+|---|---|
+| `eastus2` | refused |
+| `eastus` | refused |
+| `centralus` | created |
+
+Both the primary region and its obvious neighbour were shut, which is why dev's SQL lives
+in `centralus` while everything else stays in `eastus2`. Existing servers in a closed
+region keep working — only creation is blocked, and it can reopen (or close) without
+notice. The cross-region hop between the Container App and SQL is a real latency cost,
+accepted for dev.
+
+**Set `sqlLocation` explicitly for production** rather than letting it inherit `location`,
+so a prod day-0 does not discover this the way dev did
+([#241](https://github.com/kolatts/foundry-gate/issues/241)). The alternative is a support
+request (Issue type *Service and subscription limits*) to reopen the primary region — worth
+it when co-location matters, not worth blocking a first deploy on.
 
 ## Role assignments
 
--- a/docs-site/src/content/docs/reference/infrastructure.md
+++ b/docs-site/src/content/docs/reference/infrastructure.md
@@ -172,6 +172,18 @@ so a prod day-0 does not discover this the way dev did
 request (Issue type *Service and subscription limits*) to reopen the primary region — worth
 it when co-location matters, not worth blocking a first deploy on.
 
+:::danger[`sqlLocation` is immutable once the server exists]
+`Microsoft.Sql/servers.location` cannot be changed in place. Setting it is a **one-shot
+decision per environment**: change it after the first deploy and the incremental deployment
+*fails* rather than moving anything, and the recovery is exporting the database and
+re-creating it on a new server. Pick it before an environment's first deploy — afterwards it
+is a migration, not a parameter.
+
+Two things must agree with it: `sqlZoneRedundant` requires a region that actually has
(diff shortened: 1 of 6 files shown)


kolatts:
**REQUEST CHANGES**

The threading is right and I confirmed the boring parts hold: both `location:` usages in `sql.bicep` had to move (a database must be in its server's region, so leaving the database behind would have been an ARM error, not a latency choice), no output builds an FQDN from `location`, and nothing in `_deploy-database.yml` / `export-outputs.sh` / `ArmAzureSqlServerClient` resolves SQL by region — they all go by name and resource group, and the SQL name carries no region token the way the Foundry account names do.

What's missing is the consequence, not the change.

---

**Issues**

- **Major** `infra/modules/sql.bicep` / `reference/infrastructure` — `Microsoft.Sql/servers.location` is **immutable**. "Set `sqlLocation` explicitly for production" is only true *before* prod exists; afterwards, changing it makes the incremental deploy fail rather than move anything, and recovery is an export/recreate onto a new server. That is the single most important operational property of this parameter and neither the module comment nor the doc says it.

- **Minor** `infra/parameters/prod.bicepparam` — the docs table's prod column says "set it explicitly" and the file says nothing, so the documented contract and the actual file disagree; prod's day-0 inherits `eastus2` and fails mid-deploy exactly as dev did. Leaving it genuinely unset is defensible — you can't know prod's open region today — but every other non-obvious prod choice in that file carries a paragraph. It should be a commented stub pointing at #241, not silence.

- **Minor** `infra/parameters/prod.bicepparam` — `sqlZoneRedundant = true` and its "one availability zone in eastus2" comment are now coupled to a parameter that can move the server. Point `sqlLocation` at a region without availability zones and the database create fails; `sqlBackupStorageRedundancy = 'Geo'` also starts geo-pairing away from the rest of the stack. The advice needs an AZ rider.

- **Nit** `infra/parameters/dev.bicepparam` — the comment says the eastus2 error was `RegionDoesNotAllowProvisioning`; the deploy actually failed with `ProvisioningDisabled`. Both are real (the *create* probe returns one, the ARM deploy the other) and `sql.bicep` hedges correctly, but someone grepping the string from a red log won't match.

- **Nit** `getting-started/fork-guide.mdx` — the sample param block a forker copies still shows only `param location = 'eastus2'`, with no mention of `sqlLocation` or this failure mode. Progressive-docs invariant is satisfied at the reference level only. Not blocking; a forker on another subscription will hit this wall unsignposted.

Also worth an issue rather than this PR: nothing reads the `.bicepparam` files in `FoundryGate.Tests.Predeployment`, so nothing would catch prod shipping without a deliberate `sqlLocation`. A convention test asserting each env's param file either sets it or opts out with a marker comment is what makes the prod checkbox self-enforcing instead of a thing someone remembers.

kolatts:
Fixed in `a8fe1d8`.

**Major — immutability.** You're right that this is the property that matters, and I'd stated the parameter without stating the trap. `Microsoft.Sql/servers.location` cannot change in place, so it is one-shot per environment: change it later and the incremental deploy *fails* rather than moving anything, recovery being an export/recreate onto a new server. Now said in three places, because all three are places someone decides it: the `sql.bicep` param comment, a `:::danger` aside in `reference/infrastructure`, and both param files.

**Minor — prod silence.** Agreed, and the fix is the stub you described: `prod.bicepparam` still inherits `location` deliberately, but the file now says that is a *placeholder, not a decision*, shows the line to add, and explains why probing beats finding out mid-deploy. Silence read as "considered and fine"; it wasn't.

**Minor — zone redundancy rider.** Both constraints are on the stub now: `sqlZoneRedundant = true` needs a region that actually has availability zones, and `sqlBackupStorageRedundancy = 'Geo'` pairs from the **SQL** region rather than from `location`. The `sqlZoneRedundant` comment no longer hardcodes "eastus2" — it now says "the SQL region (eastus2 unless `sqlLocation` says otherwise)", which is the thing that would have quietly gone stale.

**Nit — wrong error string.** Correct, and the distinction turned out to be real rather than sloppy: the ARM *deploy* fails with `ProvisioningDisabled`, while a direct `az sql server create` returns `RegionDoesNotAllowProvisioning`. I hit both, from different directions, and only wrote down one. Both are named now so either greps.

**Nit — fork-guide.** Not doing it here. It needs the sample param block *and* a troubleshooting entry, and that page is already drifting from the runbook in bigger ways (one app registration vs two, one CI identity vs three) — patching one line makes the page look more current than it is. Better as one honest pass over the fork guide.

**On the convention test** — filed as #243. Agreed that this is the mechanism that makes the prod checkbox self-enforcing rather than a thing someone remembers, and worth writing as a general parameter-file convention test so the next irreversible parameter inherits the guard for free.

`az bicep build-params` clean on both `dev` and `prod`; docs build green, 17 pages.

Issue #244, opened 2026-09-05
Gateway LLM logs land in AzureDiagnostics, not the tables the reconciliation KQL queries (missing logAnalyticsDestinationType)


kolatts:
## What is wrong

`infra/modules/apim.bicep` creates the diagnostic setting that feeds the gateway's token
logs into Log Analytics, but did not set `logAnalyticsDestinationType`. Azure Monitor's
default is the legacy **`AzureDiagnostics`** catch-all table, not the resource-specific
tables.

`src/FoundryGate.Functions/Kql/UsageBySubscription.kql` — the reconciliation query, the whole
billing-grade accounting story (#84/#39) — joins:

- `ApiManagementGatewayLlmLog`
- `ApiManagementGatewayLogs`

Neither table is ever created. **The reconciliation query returns an empty result forever.**

## Why it went unnoticed

Every signal says healthy:

- the diagnostic setting exists and reports enabled for both categories
- rows really do arrive in Log Analytics, promptly
- the KQL is valid and runs without error — against a table that has no rows

Verified live 2026-09-05 on `apim-foundrygate-test-e7k2` →
workspace `d6efb708-e8a3-40df-910c-7f6234e019ca`:

```kusto
search * | summarize Rows = count() by $table
```

```
AzureDiagnostics   582
AppMetrics         414
AppExceptions      147
AppRequests        176
AppDependencies     20
```

```kusto
AzureDiagnostics | summarize Rows = count() by Category
```

```
GatewayLlmLogs   289
GatewayLogs      293
```

`ApiManagementGatewayLlmLog | summarize count()` → **0**.

So the data was flowing the entire time, into a table nothing queries. The 2026-09-01
validation recorded T6/T7 as "pending ingestion" and moved on; it was not ingestion lag.

## Fix (in this PR)

```bicep
properties: {
  workspaceId: workspaceId
  logAnalyticsDestinationType: 'Dedicated'   // <- this
  logs: [ ... ]
}
```

## Follow-up worth considering

- **Existing environments need a redeploy** to switch destination type, and rows already in
  `AzureDiagnostics` stay there. Any environment deployed before this fix has no
  resource-specific history at all.
- The postdeployment test suite should assert that
  `ApiManagementGatewayLlmLog | summarize count()` is non-zero after traffic, rather than
  only that the diagnostic setting exists. A configuration check would have passed here;
  only a data check catches it.
- `scripts/cycle/measure.ps1` now polls for exactly this and reports it, so the next live
[...]

Issue #237, opened 2026-09-05
A tier whose TPM is below one agent turn deadlocks the harness and makes its monthly quota unreachable


kolatts:
## What the live cycle showed

Observed 2026-09-05 against `apim-foundrygate-test-e7k2` with a standard tier configured at
**8 000 TPM / 40 000 tokens per month**, driving the real Codex CLI at it:

| exec | codex exit | gateway answer |
|---|---|---|
| 2 | 0 | 429, `x-fg-remaining-tpm=0`, quota 40000 → 30662 |
| 3 | **1** | 429, tpm=0 |
| 4 | **1** | 429, tpm=0 |
| 5 | **1** | 429, tpm=0 |
| … | **1** | 429, tpm=0 |
| 25 | **1** | 429, tpm=0, quota back at 40000 |

One `codex exec` spends **~9–10K tokens** before it does anything useful — its system prompt
(consistent with the ~9K/exec figure from the 2026-09-01 validation). With the per-minute cap
set *below* that, the harness **deadlocks**:

1. Codex starts an exec, spends the whole minute's bucket, gets a 429 partway through.
2. Codex retries, honouring `Retry-After`.
3. The retry spends the refilled bucket and 429s again before the exec can finish.
4. Codex eventually gives up (exit 1) having consumed tokens but produced nothing.

The exec never completes, so it never spends its **monthly** budget. Twenty-five consecutive
execs left the monthly quota effectively untouched. The 403 wall is unreachable from behind a
TPM wall that is too tight.

## Why this matters beyond the demo

This is a **tier-design constraint**, not a bug in the gateway — enforcement worked exactly as
specified. But it means a tier is not well-formed unless

```
tpm  >  (tokens one agent turn costs)
```

and a plausible-looking tier can be functionally unusable for agent harnesses while looking
fine on paper and while every policy reads as configured. A developer issued such a key
experiences "Codex is broken", not "I am over my limit".

The shipped defaults in `infra/main.bicep` are fine (standard = 20 000 TPM, power = 40 000),
so nothing in production tiering is wrong today. What is missing is the *rule*, written down,
so a fork or an admin lowering a tier's TPM knows the floor.

## Suggested work

- [ ] Document the floor in `reference/` and in the quota/tier admin UI copy: a tier's TPM must
      comfortably exceed one agent turn (~10K for Codex, more for Claude Code, whose captured
      request bodies ran ~87KB).
[...]

Issue #231, opened 2026-09-05
Anthropic deployment creation fails on a fresh account (InternalServerError) — likely subscription-level Marketplace entitlement


kolatts:
## What happened

2026-09-05, subscription **Imagile Paid** (`3d0b1b83-8540-44c5-97af-fa338056f73a`),
resource group `rg-foundrygate-test`, a **completely clean starting state**: no
`rg-foundrygate-test`, nothing in `az apim deletedservice list`, nothing in
`az cognitiveservices account list-deleted`. The 2026-09-01 environment had been torn down
and purged.

The day-0 ARM deploy created a brand-new Foundry account `fgtest-e7k2-eus2` (Succeeded) and
made **one** Anthropic create attempt:

```
fgtest-e7k2-eus2/claude-sonnet-4-5   Creating (~5 min)  ->  Failed
  code:    InternalServerError
  message: Internal Server Error.
  correlationId: 46e5db9d-54ea-4b23-9ef4-b895bca3e3b5
```

It was **not** retried and **not** deleted/recreated, per the E-007 rule. The cycle
continued on the OpenAI path (`gpt-4-1-mini` provisioned synchronously and worked), which is
the primary harness path anyway.

## Why this needs a human

This sharpens E-007 in an important way. E-007 was written as "creates start failing in an
account **after** delete/recreate churn". This run shows a first attempt on a **fresh account
in a fresh resource group with nothing soft-deleted** failing the same way, with the same
opaque `InternalServerError` and no marketplace detail in the Activity Log.

That pattern is consistent with a **subscription-level** Anthropic/Marketplace agreement or
entitlement problem rather than per-account churn — which is not something an agent can fix
from the CLI. It needs one of:

- an owner checking the Anthropic offer's Marketplace agreement status for this subscription
  (Marketplace permissions are owner-only), and/or
- an Azure support ticket quoting correlation id `46e5db9d-54ea-4b23-9ef4-b895bca3e3b5`.

Claude GlobalStandard quota is present and is not the constraint (sonnet-4-5 200, haiku-4-5
80, opus-4-5 40 units; the deployment asked for 10).

## Impact while unresolved

| Affected | Status |
|---|---|
| OpenAI front door, Codex harness, TPM 429, monthly-quota 403, alias allowlist, Log Analytics reconciliation | fully exercised, unaffected |
| `/anthropic/v1/messages` 200 (T1) | cannot be verified |
[...]

PR #247, opened 2026-09-05
Automated gateway spin-up / test / spin-down cycle, with live evidence

### LINKED ISSUE #219: Automated gateway spin-up / test / spin-down cycle (scripts + skill + live evidence)
The gateway data plane was validated by hand on 2026-09-01 and torn down. Everything about
that run — the deploy invocation, the subscription creation, the 401/403/429 probes, the
Codex harness config, the Log Analytics reconciliation — lived in an agent transcript. It
needs to be a script anyone (or any agent) can run unattended.

Owner's framing: *"we want to be able to spin up, test, and spin down frequently ...
demonstrate that we are able to start codex harnesses locally pointed at the foundry
resource and that the quota and measurement works. You can set the quotas low to make sure
they get hit."*

## Sub-tasks

- [ ] **Scripts** — `scripts/cycle/{up,subscriptions,smoke,codex-test,measure,down,status,cycle}.ps1`.
      PowerShell 7, strict mode, idempotent, no prompts, gitignored JSON state under
      `scripts/cycle/.state/`. Tiny quota overrides (standard tier: 40K tokens/month,
      8K TPM) so 429 and 403 are reachable in a handful of `codex exec` runs.
- [ ] **Skill** — `.claude/skills/gateway-cycle/SKILL.md` so a future agent can run the
      whole cycle unattended: prerequisites, exact commands, expected durations, what PASS
      looks like, the Anthropic create-once rules, KeepFoundry vs Full teardown, cost per
      hour, troubleshooting. Pointer bullet in `CLAUDE.md` + `AGENTS.md`.
- [ ] **Live run evidence** — run `cycle.ps1` against Imagile Paid for real and commit the
      report to `validation/<date>-gateway-cycle.md` (checks, timings, gateway URL,
      redacted header samples).
- [ ] **Anthropic provisioning outcome** — record whether the single day-0 Claude create
      attempt succeeded on a fresh account (E-007). Exactly one attempt per fresh account;
      never retry, never delete/recreate in a loop.
- [ ] **Docs pointer** — short "Local gateway test cycle" section in
      `docs-site/src/content/docs/reference/infrastructure.md` linking the skill and the
      validation report. Nothing lands in `cli-setup.mdx` that was not verified live.

## Teardown default is KeepFoundry

Because Anthropic deployments are create-once under ARM (E-007), the default teardown
deletes everything in the resource group *except* the `Microsoft.CognitiveServices/accounts`,
so the next `up` re-runs the template with `createModelDeployments=false` over surviving
accounts. `-Mode Full` (delete RG + purge soft-deleted APIM and cognitive accounts) exists
but is labelled with the Anthropic-churn risk.

## Also closes the KQL half of #178

`measure.ps1` runs `src/FoundryGate.Functions/Kql/UsageBySubscription.kql` against the live
workspace and reports whether duplicate `CorrelationId`s actually appear — the `max()`
de-duplication assumption from D-017.

## Human attention

- **The one Anthropic create attempt is supervised.** Everything else in the cycle is safe to run unattended — that is the point of the issue — but the day-0 `claude-haiku-4-5` create on a fresh account is create-once under ARM (E-007). One attempt, never a retry loop, never delete-and-recreate. If it fails, the cycle records the failure and moves on; a person decides whether to try again, on a new account, another day. The Claude end-to-end story itself lives on #81.
- **Cost.** `up` spends real money for as long as the environment is up. The skill states the per-hour figure and the default teardown is `KeepFoundry`; `-Mode Full` (purge soft-deleted APIM and cognitive accounts) is the Anthropic-churn footgun and is labelled as one.

## Done when

- `cycle.ps1` runs end to end unattended against Imagile Paid and produces a PASS report.
- `validation/<date>-gateway-cycle.md` is committed with checks, timings, gateway URL and redacted header samples — the evidence the landing page and `architecture/overview` can point at.
- `.claude/skills/gateway-cycle/SKILL.md` exists and a fresh agent can run the cycle from it alone; `CLAUDE.md` and `AGENTS.md` carry the pointer bullet.
- The KQL de-duplication question from #178 §1a is answered with live numbers.
- Nothing lands in `cli-setup.mdx` that was not captured off the wire during the run.

### LINKED ISSUE #230: APIM product/API links race on deploy: 'Link already exists' leaves tier products with no APIs
## What happened

Clean gateway deploy against `Imagile Paid` on 2026-09-05 (`rg-foundrygate-test`,
`apim-foundrygate-test-e7k2`). The `foundrygate-ai-gateway` module failed with four
identical errors:

```
{"code":"Conflict","message":"Link already exists between specified Product and Api."}
```

Per-operation states:

| Resource | State |
|---|---|
| `unlimited/anthropic-link` | Succeeded |
| `unlimited/openai-link` | Succeeded |
| `standard/anthropic-link` | **Failed — Conflict** |
| `standard/openai-link` | **Failed — Conflict** |
| `power/anthropic-link` | **Failed — Conflict** |
| `power/openai-link` | **Failed — Conflict** |

This was a **first** deployment of those links — nothing pre-existed. ARM issued all six
`Microsoft.ApiManagement/service/products/apiLinks` writes in parallel, three of them
targeting `foundrygate-anthropic` and three targeting `foundrygate-openai`, and APIM
rejected the concurrent writes against the same API.

## Why it matters

`az apim product api list --product-id standard` came back **empty**. A tier product with
no APIs linked serves nothing: every APIM subscription issued against `standard` or `power`
would have been dead on arrival, and the whole enforcement story (`llm-token-limit` lives in
the tier product policy) silently would not apply. The deployment reported failure, but a
partial success here is the dangerous shape — the products exist and look right in the
portal.

## Fix (in this PR)

`@batchSize(1)` on both `apiLinks` loops in `infra/modules/ai-gateway.bicep`, plus a
`dependsOn` from the OpenAI loop to the Anthropic loop so the two loops cannot race each
other either. Costs a few seconds of deploy time.

## Follow-up worth considering

- The same shape may exist for other APIM child collections written per-tier. `groupLinks`
  and `tagLinks` are the obvious candidates if either is added later.
- A post-deploy assertion that every tier product has both APIs linked would have caught
  this without a human reading operation lists. `scripts/cycle/smoke.ps1` now catches it
  indirectly (a key on an unlinked product 404s), but an explicit check would name the cause.

kolatts:
Closes #219. Also fixes #230 and #244, and implements the demo-tier half of #237.

Turns the 2026-09-01 hand-run gateway validation into `scripts/cycle/` — eight re-runnable
PowerShell stages sharing one gitignored JSON state file — plus a `gateway-cycle` skill so an
agent can run the whole thing unattended, and **ran it live on Imagile Paid**.

```sh
pwsh scripts/cycle/cycle.ps1 -Subscription "Imagile Paid"
```

## Live evidence — `validation/2026-09-05-gateway-cycle.md`

Gateway `https://apim-foundrygate-test-e7k2.azure-api.net`. **22 PASS, 1 FAIL, 2 SKIP** — the
FAIL and both SKIPs are the one Anthropic provisioning failure and its two dependents.

| | |
|---|---|
| `T3a-d` | no key and bad key → 401 on **both** front doors |
| `T2a/T2b` | `/openai/v1/chat/completions` and `/responses` → 200 with `api-key` |
| `A1/A2/A3` | unknown alias, the **real deployment name**, and a Claude alias on the OpenAI door → 403 `model_not_permitted` |
| `T4a` | `429`, `Retry-After: 13`, `x-fg-remaining-tpm: 0` after 5 requests |
| `T4b` | same-tier `dev-bob` unaffected — counter keyed on subscription |
| `T5a` | `403 Token quota is exceeded. Try again in 25 days…`, `x-fg-remaining-quota: 0` |
| `T5c` | power-tier `dev-carol` still 200 while `dev-alice` is 403 |
| **`C1`** | **a real `codex exec` completed through the gateway** (`model: gpt, provider: azure`) |
| **`C2`** | Codex at the TPM wall: gateway 429 `Retry-After: 30`, codex exits 1 |
| **`C3`** | Codex at the monthly wall after 4 execs: gateway 403, codex exits 1 |
| `M1/M2/M3` | 59 log entries; per-developer reconciliation; the D-017 check |

**The measurement cross-check is the headline.** `UsageBySubscription.kql` independently
corroborates the gateway's own enforcement: `dev-alice` logged **41 089 tokens against the
40 000 cap she was 403'd on — ratio 1.03**, exactly the shape expected when the gateway stops
a developer on the request that crosses the cap.

## Two infra bugs the live run found, both silent

**#244 — the token logs were going to the wrong table all along.** The APIM diagnostic setting
lacked `logAnalyticsDestinationType: 'Dedicated'`, so both categories went to the legacy
`AzureDiagnostics` catch-all instead of `ApiManagementGatewayLlmLog` /
`ApiManagementGatewayLogs` — the two tables `UsageBySubscription.kql` joins. Every signal read
healthy (setting enabled, rows arriving, KQL valid and error-free) while the reconciliation
query returned empty forever. Live proof: 582 rows in `AzureDiagnostics`, zero in either
resource-specific table. **The 2026-09-01 validation recorded this as "pending ingestion" — it
was never ingestion lag.**

**#230 — `apiLinkId` is unique per APIM *service*, not per product.** All three tier products
declared their link as `anthropic-link`, so the first won the name and the rest failed with
`Conflict: Link already exists`. The dangerous part is the partial success: the tier products
existed and looked right in the portal *with no APIs attached*, so every key issued against
`standard` or `power` would have been dead on arrival while every policy still read as
configured.

## #237 — a tier's TPM has a floor, and it is one agent turn

`codex exec` spends ~9–10K tokens on its system prompt before doing anything. Below that, the
harness **deadlocks**: codex 429s, retries, keeps the bucket empty, gives up without finishing,
and therefore never spends its monthly budget — the 403 is unreachable from behind a tight 429.
Verified across 25 consecutive execs at both 8 000 and 12 000 TPM. So the two demo tiers now
have deliberately opposite shapes, each proving one meter: `standard` (40K/mo, 12K TPM) proves
the 429 wall, `power` (60K/mo, 100K TPM) proves the 403 wall. Shipped defaults in `main.bicep`
are unaffected; #237 tracks writing the rule down.

## Anthropic — one attempt, not retried (#231, `needs-human`)

The single day-0 Claude create failed with an opaque `InternalServerError` on a **completely
fresh account, fresh resource group, nothing soft-deleted** (correlation id
`46e5db9d-54ea-4b23-9ef4-b895bca3e3b5`). It was **not** retried and **not** deleted/recreated,
per E-007. E-007 described *post-churn* failures; a first attempt on a clean account failing the
same way points at a subscription-level Marketplace entitlement problem, which no agent can fix
from the CLI. The demo continued on the OpenAI path — which is the primary harness path anyway.
`T1` and `C4` are SKIP, not FAIL, in consequence.

## Second-cycle bugs a hand-run validation could never have hit

The whole point is running this *frequently*, and everything below only appears on the second
cycle:

- **APIM soft-delete blocks the next spin-up.** Deleting an APIM service reserves the name, so
  the next deploy dies with `ServiceAlreadyExistsInSoftDeletedState`. Without the purge the
  cycle works exactly once.
- **The monthly quota counter cannot be reset.** It is keyed on the APIM subscription, so
  reusing `dev-alice` means the second cycle in a calendar month starts her at 403 and can
  demonstrate neither wall. Subscription ids now carry a cycle stamp.
- **A freshly created key is not immediately usable** — it must propagate to the gateway nodes,
  and until it does the gateway 401s a key that works a minute later.
- **Log Analytics ingestion outlasts the cycle**, so a teardown that deletes the workspace
  destroys the measurement evidence minutes before it arrives. `KeepFoundry` now keeps the
  telemetry stores too; neither bills at rest.

## Teardown: KeepFoundry by default

Deletes everything except the Foundry accounts and the telemetry stores, and purges the
soft-deleted APIM. Foundry accounts survive because Anthropic deployments are create-once per
account (E-007) — destroying them spends the next spin-up's only Claude create attempt. APIM,
the only meaningful idle cost (~$0.28/hr), always goes. Confirmed by `status.ps1`: **~$0.00/hr**
after teardown.

## Verification

- `pwsh scripts/validate-policies.ps1` — passes (run by `up.ps1` every cycle)
- `cd docs-site && npm ci && npm run build` — passes
- All ten cycle scripts parse clean
- The live cycle above

## Issues filed

- #230 APIM product/API link race (fixed here)
- #231 Anthropic create fails on a fresh account — **`needs-human`**, Marketplace/support
- #237 TPM floor / tier-sizing guidance
- #244 diagnostic destination type (fixed here)


kolatts:
## `sunny:code-review`

**REQUEST CHANGES**

The cycle itself is good work and the live evidence is the real deliverable — 22 PASS with a
reconciliation ratio of 1.03 against the enforced cap is about as convincing as this gets, and
finding #244 (token logs in the wrong table the whole time) justifies the whole exercise on its
own. But `down.ps1` will cheerfully delete the dev control plane if you typo one flag, and that
blocks.

---

**Issues**

- **Major** `scripts/cycle/down.ps1` — `-Environment` is an unvalidated free string and
  `KeepFoundry` is the *default* mode with no confirmation, so `down.ps1 -Environment dev`
  deletes the dev SQL server, Key Vault, ACR, Container App and Static Web App. Those exist in
  Imagile Paid right now. Nothing about the name "KeepFoundry" hints at that.

- **Major** `scripts/cycle/subscriptions.ps1` — `SUB-1` is recorded as an unconditional `PASS`
  even when the propagation probe just printed "still 401" for every key, so the evidence report
  gets a green check for a stage that failed. A check that cannot fail is worse than no check.

- **Minor** `scripts/cycle/report.ps1` — `[bool]$state.claudeAvailable` is dereferenced without
  a `ContainsKey` guard, and `up.ps1` only writes that key after a successful deploy. Under
  StrictMode a failed `up` therefore makes the report throw from inside `cycle.ps1`'s `finally`
  — losing the report in precisely the case it exists for.

- **Minor** `scripts/cycle/subscriptions.ps1` — the 180s propagation deadline is computed once
  outside the `foreach`, so all three keys share one budget; if the first key eats it, the other
  two are reported as failed without ever being probed.

- **Minor** `scripts/cycle/up.ps1` — the what-if path slices `$deployArgs` from a magic index 3
  and then re-appends `--name`/`--location`, duplicating both. az takes the last value so it
  works by luck, and it breaks silently the moment anyone reorders the array above.

- **Minor** `scripts/cycle/codex-test.ps1` — `Start-Process -FilePath 'claude'` uses the bare
  name, which is the exact bug `Resolve-CliPath` was written to fix and whose comment even
  describes the `claude` shim. It also has no SKIP path for "Claude deployment exists but the
  CLI isn't installed".

- **Minor** `docs-site/.../reference/infrastructure.md` — the new section quotes the demo tier as
  8 000 TPM, which is both stale (the default is 12 000) and specifically the value this same PR
  proves deadlocks the harness. It also omits the `power` tier, which is the entire point of the
  opposite-shapes design, and understates what `KeepFoundry` keeps.

- **Minor** `docs-site/.../reference/infrastructure.md` — `logAnalyticsDestinationType:
  'Dedicated'` is the line that makes billing-grade accounting work for the first time, and it is
  documented nowhere. A fork enabling the diagnostic setting by hand hits the identical silent
  failure.

- **Nit** `scripts/cycle/cycle.ps1` — `$LASTEXITCODE -ne 0 -and $null -ne $LASTEXITCODE` tests
  before it guards; under StrictMode the first operand throws on an unset value, so the guard can
  never run.

- **Nit** `scripts/cycle/report.ps1` — the header promises keys are redacted "on the way into the
  state file and again on the way out", but `down.ps1` drops `apimSubscriptions` before the report
  renders, so the second pass has nothing to match on. The claim oversells the control.

---

**Not flagged, deliberately:** `Assert-Check` being defined three times, the single-element
`foreach` in `up.ps1`, and `Invoke-Stage`'s unused return value all came up. They are real, they
are small, and folding them now would churn ten scripts that just proved themselves against live
Azure. Worth a tidy pass later, not worth blocking on.

kolatts:
## Response — all ten addressed in b2e250a

**Major — `down.ps1` would delete the dev control plane.** Fixed, and thank you, this was the
right thing to block on. `down.ps1` now enumerates the resource group before deleting anything
and refuses outright if it finds `Microsoft.Sql/servers`, `Microsoft.App/containerApps`,
`Microsoft.Web/staticSites`, `Microsoft.KeyVault/vaults` or
`Microsoft.ContainerRegistry/registries`, unless `-AllowControlPlane` is passed. Verified live:

```
$ pwsh scripts/cycle/down.ps1 -Environment dev -Subscription "Imagile Paid"
  Microsoft.Web/staticSites/stapp-foundrygate-dev
  Microsoft.KeyVault/vaults/kv-fg-dev-e7k2
  Microsoft.ContainerRegistry/registries/crfoundrygatedeve7k2
  Microsoft.Sql/servers/sql-foundrygate-dev-e7k2
  Microsoft.App/containerApps/ca-foundrygate-api-dev
rg-foundrygate-dev holds 5 control-plane resource(s), listed above. scripts/cycle/ is a
gateway-only test harness and will not delete a control-plane environment. If you genuinely
mean to, pass -AllowControlPlane.
```

The check is on resource *type* rather than on the environment name on purpose — a name
allow-list would have to be kept in sync with whatever anyone names an environment next, and the
thing that actually makes deletion unsafe is what is in the group.

**Major — `SUB-1` could not fail.** Fixed. The propagation loop now collects the keys that never
stopped 401ing, and `SUB-1` is `PASS` only when that list is empty; otherwise it FAILs naming
them. You are right that this was the worst one on the list after the deletion bug — a green
check printed directly beneath red console output is actively misleading, and the evidence report
is the whole deliverable here.

**Minor — `report.ps1` throws on a failed `up`.** Fixed, guarded with `ContainsKey`, and the
missing case now renders as "not determined (the deploy did not get that far)" rather than
silently reading as "no Claude". This one stung: the report is generated from a `finally` block
specifically so a failed cycle still produces evidence, and the unguarded dereference defeated
exactly that.

**Minor — shared propagation deadline.** Fixed; `$deadline` moved inside the loop so each key
gets its own 180s.

**Minor — what-if argument slicing.** Fixed. It now swaps the verb and reuses the tail
unchanged instead of slicing from a magic index and re-appending. You are right that it worked
only because az takes the last value of a duplicated flag.

**Minor — `claude` launched by bare name.** Fixed, routed through `Resolve-CliPath`, and I added
the missing SKIP branch for "a Claude deployment exists but the CLI is not installed" — that
would previously have been an unhelpful FAIL on a machine that simply lacks the tool.

**Minor — stale demo tier numbers in the docs.** Fixed, and this was an embarrassing one: the
section quoted 8 000 TPM, which is the exact value the same PR proves deadlocks the harness. It
now carries both tiers as a table with the reason they are shaped oppositely, and states the TPM
floor explicitly.

**Minor — `Dedicated` undocumented.** Fixed. `reference/infrastructure` gets a short subsection
explaining that without it the rows land in `AzureDiagnostics`, that every signal reads healthy
while reconciliation returns nothing, and the two-line KQL to tell the two cases apart. Called
out for forks enabling the diagnostic setting by hand, which was your point 3.

**Nit — `$LASTEXITCODE` guard ordering.** Fixed, guard first.

**Nit — the redaction claim in `report.ps1`'s header.** Fixed by making the comment honest rather
than by adding a second control: the real guarantee is that `Add-CycleCheck` redacts on the way
*in*, and the header now says so and explains why the outbound pass has nothing left to match by
the time a full cycle renders.

---

**On the three you consciously did not flag** — agreed on all three, and agreed on the reasoning.
`Assert-Check` in particular is a copy waiting to drift, but these ten scripts just spent an
afternoon being debugged against live Azure and I would rather not re-open them for a tidy. Filed
as **#251** so it does not evaporate.

**Also worth flagging separately:** while rebasing, main landed #240, which fixes the same APIM
product/API link failure I hit — and fixes it better. `products/apiLinks` is not idempotent *at
all* (APIM enforces uniqueness on the (product, api) pair, so a re-PUT is refused however the link
is named); my tier-prefixed ids would have fixed the first deploy and still broken every re-run.
The merge takes main's `products/apis` version wholesale and I have closed **#230** as a duplicate
of #239.

PR #248, opened 2026-09-05
fix(ci): the API smoke test must supply Gateway__Tiers, which is never optional

### LINKED ISSUE #246: API image smoke test starts the container with no config, but the app validates config at startup
Blocking the API/Functions/UI stages of the first dev deploy (#105). [Run 33975323485](https://github.com/kolatts/foundry-gate/actions/runs/33975323485), job *API - Build, push and deploy (dev)*.

The stage fails **before any Azure interaction** — in the local `Build and smoke-test the image` step, which does a bare `docker run` and polls `/health`:

\\\
##[error]/health did not return 200 within 30s for foundrygate-api:0.1.0-613e50d
Unhandled exception. Imagile.Framework.Configuration.Exceptions.ConfigurationValidationException:
  Gateway.Tiers: Tiers must contain at least one tier. On a deployed host infra sets them as
  Gateway__Tiers__0__ProductId / …__DisplayName / …__MonthlyTokenQuota from infra/main.bicep's
  quotaTiers parameter; locally they come from appsettings.local.json's Gateway:Tiers. Neither is
  present here, and quota resolution has no meaning without a tier table.
   at ConfigurationValidationExtensions.ValidateRecursively(Object obj)
   at Program.<Main>$ in /src/src/FoundryGate.Api/Program.cs:line 48
\\\

## The contradiction

Two deliberate design choices that cannot both hold:

1. **The app fails fast on missing configuration.** Program.cs calls ValidateRecursively at startup, and the Gateway.Tiers message is a *good* error — it names the env vars, the bicep parameter and the local file. That is the behaviour you want on a real host.
2. **The smoke test starts the image with no configuration at all** and expects /health to return 200.

The container is doing exactly what it was built to do. The test is asserting something the app was explicitly designed not to do.

Note this is not caught by anything else: the image builds fine, FoundryGate.Tests.Predeployment is green (1545+), and the deployed Container App *does* get Gateway__Tiers__* from control-plane.bicep. Only a bare docker run hits it — which is why it survived until the first real deploy.

## Options

1. **Give the smoke test a minimal config** — one tier via -e Gateway__Tiers__0__…, plus whatever else ValidateRecursively requires. Keeps the test meaningful (the app really does start and serve /health) at the cost of a fixture the smoke test has to keep in step with the options classes.
2. **Smoke-test for the right thing.** What this step actually protects against is a broken image — missing entrypoint, wrong base, bad publish layout. Asserting *"the process starts and fails with a configuration error"* proves that, and proves the validation works, without pretending an unconfigured container should be healthy. /health on a configured container is already verified later against the real Container App (Verify new revision and /health).
3. Point the smoke test at /health only, with the app's config validation relaxed for a hermetic health endpoint — rejected on sight: it would weaken a real production guard to satisfy a test.

Option 2 is the honest one, and it is cheaper. Option 1 is defensible if the intent is genuinely to catch runtime startup regressions before pushing to ACR.

## Current dev state

Everything upstream of this succeeded: infra (7m3s), database prepare (56s), **dacpac deploy (2m17s)**. The Container App is still running the bootstrap placeholder mcr.microsoft.com/k8se/quickstart:latest. Functions/UI/postdeployment-tests never ran because they sit behind the api stage.

Refs #105

kolatts:
## What

The API stage of the first dev deploy died **before touching Azure**, in the local `docker run` smoke test:

\\\
/health did not return 200 within 30s for foundrygate-api:0.1.0-613e50d
ConfigurationValidationException: Gateway.Tiers: Tiers must contain at least one tier.
\\\

The smoke step **already** passes placeholder configuration — `AzureAd__*`, the connection string, `KeyProtection__Provider` — precisely because `Program.cs` validates at startup and the host exits without them. `Gateway__Tiers` was just missing from that list. So this is the smaller change: extend a list that already exists, in the pattern the file already documents.

## Which of the two options, and why

The coordinator's framing was right that a hermetic `/health` should not need SQL/APIM config — and it doesn't. `GatewayOptions`' *addressing* members (`ApimName`, `FoundryAccountNames`, `LogAnalyticsWorkspaceId`…) are all optional; the features that need them fail with a clean `503`. `Tiers` is the deliberate exception:

> `Tiers` is *always* required — quota resolution has no meaning without it.

So **placeholder config in the smoke step**, not weakened validation. That error was a *good* error: it named the env vars, the bicep parameter and the local file. Making `/health` reachable by relaxing a real production guard would trade a genuine fail-fast for a green check.

**Two tiers, not one.** Validation insists on exactly one unlimited tier (`MonthlyTokenQuota = 0`) **and** at least one finite tier for finite quotas to map onto — a one-entry table fails as loudly as an empty one. Product ids must be real (`GatewayTiers.All`) because unknown ids are rejected by name; the caps are deliberately fake, since this proves the host starts and serves `/health`, not what dev's quotas are.

## Why it survived until the first real deploy

The image used to ship `appsettings.local.json` past a mis-cased `.dockerignore` pattern, and that file carried `Gateway:Tiers` into the container. The *Verify the image ships no local configuration* step directly above — added to close that hole — removed the accidental supply without anyone adding the deliberate one. **This smoke test only ever passed because of the bug that step now catches.**

## Verification — real image, both directions

Built `src/FoundryGate.Api/Dockerfile` locally and ran the exact env from the action:

\\\
old env -> curl 000; ConfigurationValidationException: Gateway.Tiers … in docker logs
new env -> /health 200 after 2s
           {"status":"Healthy","version":"0.1.0-local","environment":"local"}
           /health/ready -> 503   (no SQL — hermetic/ready split intact)
\\\

The `/health/ready` 503 is the part worth noting: it confirms the fix does **not** paper over the readiness probe. `/health` is hermetic and answers; `/health/ready` still tells the truth about the database.

Touching `.github/` here is the documented exception — the workflow is the thing that is actually broken.

Closes #246
Refs #105

### DIFF
--- a/.github/actions/build-api-image/action.yml
+++ b/.github/actions/build-api-image/action.yml
@@ -91,6 +91,16 @@ runs:
         # /health itself is hermetic, so placeholder values are the point.
         # KeyProtection__Provider=DataProtection is the documented local-only substitute for a
         # Key Vault wrapping key (the real one is Gateway__KeyEncryptionKeyUri, an infra output).
+        #
+        # Gateway__Tiers__* is REQUIRED — GatewayOptions.Tiers is the one member that is never
+        # optional, because quota resolution has no meaning without a tier table. The image no
+        # longer ships appsettings.local.json (the .dockerignore fix above), so nothing supplies
+        # it here unless this step does. Two tiers, not one: validation insists on exactly one
+        # unlimited tier (MonthlyTokenQuota = 0) AND at least one finite tier for finite quotas to
+        # map onto, so a single-tier table fails just as loudly as an empty one. The product ids
+        # must be real (GatewayTiers.All / infra's quotaTiers) — unknown ids are rejected by name.
+        # The CAPS are deliberately not the real ones: this proves the host starts and serves
+        # /health, not what dev's quotas are.
         docker run -d --name "$container" -p 8080:8080 \
           -e 'KeyProtection__Provider=DataProtection' \
           -e 'AzureAd__Instance=https://login.microsoftonline.com/' \
@@ -99,6 +109,12 @@ runs:
           -e 'AzureAd__TenantId=00000000-0000-0000-0000-000000000000' \
           -e 'AzureAd__ClientId=00000000-0000-0000-0000-000000000000' \
           -e 'AzureAd__Audience=api://00000000-0000-0000-0000-000000000000' \
+          -e 'Gateway__Tiers__0__ProductId=standard' \
+          -e 'Gateway__Tiers__0__DisplayName=Standard' \
+          -e 'Gateway__Tiers__0__MonthlyTokenQuota=1000000' \
+          -e 'Gateway__Tiers__1__ProductId=unlimited' \
+          -e 'Gateway__Tiers__1__DisplayName=Unlimited' \
+          -e 'Gateway__Tiers__1__MonthlyTokenQuota=0' \
           "$IMAGE"
         for i in $(seq 1 "$TIMEOUT"); do
           code="$(curl -s -o /dev/null -w '%{http_code}' http://localhost:8080/health || true)"
--- a/src/FoundryGate.Api/Dockerfile
+++ b/src/FoundryGate.Api/Dockerfile
@@ -11,8 +11,19 @@
 #     -e AzureAd__Audience=api://00000000-0000-0000-0000-000000000000 \
 #     -e KeyProtection__Provider=DataProtection \
 #     -e AzureAd__Instance=https://login.microsoftonline.com/ \
+#     -e Gateway__Tiers__0__ProductId=standard \
+#     -e Gateway__Tiers__0__DisplayName=Standard \
+#     -e Gateway__Tiers__0__MonthlyTokenQuota=1000000 \
+#     -e Gateway__Tiers__1__ProductId=unlimited \
+#     -e Gateway__Tiers__1__DisplayName=Unlimited \
+#     -e Gateway__Tiers__1__MonthlyTokenQuota=0 \
 #     foundrygate-api:local
 #   curl -i http://localhost:8080/health          # 200 (hermetic liveness; no DB needed)
+#
+# Two tiers, and both are needed: GatewayOptions requires exactly one unlimited tier
+# (MonthlyTokenQuota = 0) and at least one finite tier, so a single entry fails as loudly as
+# none. The product ids must be real ones (FoundryGate.Domain.Constants.GatewayTiers); the caps
+# here are placeholders.
 #
 # Those settings are required by Program.cs's startup validation and MUST come from the
 # environment — the image ships no appsettings.local.json (.dockerignore, enforced by


kolatts:
**REQUEST CHANGES** — one line, and it's the twin of the thing being fixed.

The tier table itself is right: I checked it against all seven rules in `GatewayOptions.Validate` and it satisfies every one, including the `StringComparer.Ordinal` product-id match that would have rejected `Standard`. The env set is complete for `local`, `api-deploy.yml` shares the same composite action so the PR-only check gets the fix too, and no predeployment test asserts on this file. Verifying against a real image build rather than reasoning about it is the right instinct.

---

**Issues**

- **Major** `src/FoundryGate.Api/Dockerfile` 6–20 — the header documents the exact `docker run` a developer should copy, and it does **not** include `Gateway__Tiers__*`. Following the repo's own instructions now produces the identical `ConfigurationValidationException` this PR exists to fix. That recipe and the smoke step are two copies of one thing; fixing one and not the other just moves the trap from CI to the next person building locally.

- **Minor** `.github/actions/build-api-image/action.yml` 112–117 — this is now a **fourth** hand-maintained copy of the tier table (bicep `quotaTiers`, Api `appsettings.local.json`, Functions `appsettings.local.json`, here). The first three are pinned to each other by `GatewayOptionsTiersTests` / `GatewayTiersTests`; this one is pinned to nothing. Rename `standard` in `GatewayTiers` and bicep and every test stays green while the API image build starts failing on the next PR that touches the Dockerfile — a confusing failure a long way from its cause.

- **Minor** `.github/actions/build-api-image/action.yml` 100 — `ASPNETCORE_ENVIRONMENT=local` is *correct* here (`KeyProtectorFactory` hard-refuses `DataProtection` outside `local`, and both protectors are eagerly resolved), but it means the smoke test proves "boots in the local shape", not "boots in the shape infra deploys". Two eager qa-only startup gates are never exercised: `KeysServiceCollectionExtensions`' `AddResolveOnStartup<IApimManagementClient>` (throws outside `local` without `Gateway:SubscriptionId`/`ResourceGroup`/`ApimName`) and `KeyProtectorFactory`'s `KeyVault` branch. A change that breaks qa-only registration passes this check and fails during the actual rollout — which is the exact failure mode this PR closes.

- **Nit** `src/FoundryGate.Functions/FoundryGate.Functions.csproj` 33–35 — `appsettings.local.json` has no `<CopyToPublishDirectory>Never</CopyToPublishDirectory>`, unlike `local.settings.json` two items below it, so it ships in the Functions package. Inert (tier table only, no secret, and `DOTNET_ENVIRONMENT` isn't `local` there) but it is precisely the leak class the API's "Verify the image ships no local configuration" step exists to catch, and the Functions package has no equivalent check.

- **Nit** `.github/workflows/_deploy-functions.yml` 126–133 — the Verify step asserts `state == Running`, which a Flex Consumption app reports whether or not the isolated worker threw during startup validation. The Functions host binds the same `[Required]` `GatewayOptions` plus its own `Entra.ApplicationClientId` rule, so a violation surfaces at 02:00 inside a timer trigger rather than in the pipeline.

The last three aren't this PR's job — issues, not changes. Only the Dockerfile blocks.

<sub>Checked: tier validity against `GatewayOptions.cs`; every `[Required]`/`IValidatableObject` rule reachable from `Program.cs` line 48; whether the Functions host has the same-shaped gap (it has no smoke test at all — `dotnet publish` → `functions-action`, so no); `api-deploy.yml` sharing the composite; predeployment tests touching `.github/` (none do).</sub>

kolatts:
Fixed in `60c21d0`.

**Major — the Dockerfile recipe.** Right, and it's the more embarrassing half of the same bug: the CI copy was fixed while the copy a human is told to paste was left broken. Added the six vars to the header, plus two sentences on *why* it's two tiers (one unlimited + one finite are both mandatory) and that the ids must be real while the caps are placeholders — so the next person editing it knows which parts are load-bearing.

Verified by running the header's recipe **verbatim** against a real image rather than assuming symmetry with the action:

```
Dockerfile header recipe -> /health 200 after 2s
{"status":"Healthy","version":"0.1.0-local","environment":"local"}
```

**Minor — fourth unpinned copy of the tier table** → **#249**. Agreed, and your rename example is the right way to see it: every test stays green while the image build breaks on an unrelated PR. Noting there that the test should assert *membership* in `GatewayTiers.All` rather than equality, since the caps are deliberately not the real ones.

**Minor — `local` proves the wrong startup shape** → also **#249**. This is the sharpest finding in the review and I'd underrated it: `ASPNETCORE_ENVIRONMENT=local` is *required* for this env set to work, which is exactly what makes it misleading. `AddResolveOnStartup<IApimManagementClient>` and `KeyProtectorFactory`'s `KeyVault` branch are both eager and both qa-only, so the check cannot catch the class of break it exists for. Your suggested second `docker run` at `qa` with placeholder addressing is recorded verbatim — both clients being lazy at construction is what makes it viable without network.

**Nits — Functions `appsettings.local.json` + the `state == Running` assertion** → **#250**. The csproj one is worth more than its severity suggests: it is the same leak class that shipped a SQL password in every API image, and the Functions package has no equivalent content check. The Verify-step point stands on its own — asserting a state that cannot fail is not a verification.

**Not acted on:** the `set -uo pipefail` note (pre-existing, and the loop dumps `docker logs` anyway, so the cost is 30s rather than a lost diagnosis) and the `Ordinal` vs `OrdinalIgnoreCase` inconsistency between `Tiers` and `ModelAliases` (pre-existing; worth knowing that `ProductId=Standard` would be rejected, which is why the casing here is deliberate).

Issue #250, opened 2026-09-05
Functions package ships appsettings.local.json, and 'Verify' passes on a worker that failed startup


kolatts:
Two Functions-side gaps noticed while fixing the API smoke test (#246/#248). Neither is urgent; both are the Functions equivalent of guards the API already has.

## 1. `appsettings.local.json` ships in the Functions package

`src/FoundryGate.Functions/FoundryGate.Functions.csproj` copies `appsettings.local.json` to output with **no** `<CopyToPublishDirectory>Never</CopyToPublishDirectory>` — unlike `local.settings.json` two items below it, which has exactly that. So `publish/functions/appsettings.local.json` is deployed to the Function App.

**It is inert**: the file holds only the tier table (no secret), and `DOTNET_ENVIRONMENT` on the deployed host is not `local`, so it is never loaded. But it is precisely the leak class the API's *Verify the image ships no local configuration* step exists to catch — that step was added after a mis-cased `.dockerignore` pattern shipped `appsettings.local.json` **with a SQL password** into every API image. The Functions package has no equivalent check.

Two things: add the missing `CopyToPublishDirectory`, and add a package-content assertion to `_deploy-functions.yml` so a pattern that silently matches nothing is caught on the artefact rather than trusted.

## 2. `Verify` asserts on a state that cannot fail

`.github/workflows/_deploy-functions.yml` verifies the deployment by asserting `state == Running`. A Flex Consumption app reports `Running` regardless of whether the isolated worker threw during startup — the state describes the *host*, not the worker.

The Functions host binds the same `[Required]` `GatewayOptions` the API does, plus its own `Entra.ApplicationClientId` rule. A violation of either surfaces at 02:00 UTC inside `EntraSyncFunction`, not in the pipeline — and #246 is the proof that config-validation failures are a real class of deploy break, not a hypothetical one.

Worth deciding what a meaningful post-deploy signal is: an admin-endpoint probe, a log query for the worker's startup line, or a one-shot function invocation.

Refs #246, #248

PR #253, opened 2026-09-05
fix(ci): read the Function App state from ARM, not az functionapp

### LINKED ISSUE #252: az functionapp show returns an empty state on Flex Consumption — the Functions Verify step can never pass
The Functions stage of the dev deploy **deployed successfully** and then failed its own verification:

\\\
X Function App func-foundrygate-dev-e7k2 is '', expected Running.
\\\

[Run 33977269265](https://github.com/kolatts/foundry-gate/actions/runs/33977269265) — *Functions - Deploy (dev)*: every step green (`Deploy to Azure Functions (Flex Consumption, identity-based)` included), only `Verify` red.

## The app was fine. The query was not.

`az functionapp show --query state` returns **empty** for a Flex Consumption app. The ARM resource carries `properties.state = "Running"` the whole time — the `az functionapp` command's projection just drops it for this SKU. Verified three ways against the live app:

| Command | `state` |
|---|---|
| `az functionapp show --query state` | *(empty)* |
| `az webapp show --query state` | `Running` |
| `az resource show --resource-type Microsoft.Web/sites --query properties.state` | `Running` |
| raw ARM GET, api-version 2024-04-01 | `Running` |

So the assertion `[ "" != "Running" ]` could **never** pass on this SKU. The Functions stage was guaranteed to fail on its first real run, and the error message blamed the app for what was a query problem.

Independent confirmation the deploy actually worked — all three timers are registered and the runtime is what the bicep asked for:

\\\
az functionapp function list -> EntraSyncFunction, MonthlyQuotaResetFunction, UsageSyncFunction
runtime: dotnet-isolated 10.0   sku: FlexConsumption   state: Running
\\\

(That also closes the **.NET 10 isolated accepted by Flex Consumption in eastus2** checkbox on #105 — no fallback to 9.0 needed.)

## Fix

Read the ARM property directly with `az resource show`, which applies no command-specific projection and so cannot develop the same blind spot. `az webapp show` also works today, but it is the same command family that just got this wrong.

Also: distinguish *"could not read the state"* from *"the state is wrong"*. The original message — is '', expected Running — sent the reader looking at the Function App instead of at the query.

## Related

This is the concrete instance of the second bullet on #250, and sharper than that issue guessed: I filed it as "`state == Running` passes whether or not the worker started", but on Flex Consumption the property is not populated for `az functionapp` at all. The broader point on #250 still stands — a state check does not prove the isolated worker survived startup validation, which is the failure mode #246 was.

Refs #105, #250

kolatts:
## What

The Functions stage of the dev deploy **deployed successfully** and then failed its own verification:

```
X Function App func-foundrygate-dev-e7k2 is '', expected Running.
```

[Run 33977269265](https://github.com/kolatts/foundry-gate/actions/runs/33977269265) — every step of *Functions - Deploy (dev)* green, only `Verify` red.

`az functionapp show --query state` returns **empty** on Flex Consumption. The ARM resource carries `properties.state = "Running"` the whole time; the `az functionapp` projection just drops it for this SKU. The assertion `[ "$state" != "Running" ]` therefore could **never** pass, and the Functions stage was guaranteed to fail on its first real run.

## Verified four ways against the live app

| Query | `state` |
|---|---|
| `az functionapp show --query state` | *(empty)* |
| `az webapp show --query state` | `Running` |
| `az resource show --resource-type Microsoft.Web/sites --query properties.state` | `Running` |
| raw ARM GET, api-version 2024-04-01 | `Running` |

Switched to **`az resource show`**: it applies no command-specific projection, so it cannot develop the same blind spot. `az webapp show` also works today, but it is the same command family that just got this wrong — preferring the generic reader is the point, not a style choice.

Confirmed the replacement passes the real assertion before committing:

```
state=[Running]
ASSERTION PASSES
```

## Second change: an unreadable state is its own failure

The old message — `is '', expected Running` — blamed the app for what was a query problem, and sent me to the portal before the CLI. Empty now reports as *"could not read the state … the deploy itself may have succeeded"*.

## The deploy was fine, which is the point

```
az functionapp function list -> EntraSyncFunction, MonthlyQuotaResetFunction, UsageSyncFunction
runtime: dotnet-isolated 10.0    sku: FlexConsumption    state: Running
```

All three timers registered, on the runtime the bicep asked for. That also settles the **".NET 10 isolated accepted by Flex Consumption in eastus2"** checkbox on #105 — no fallback to `9.0` needed.

## Scope

This is the concrete instance of #250's second bullet, and sharper than that issue guessed: I filed it as "`state == Running` passes whether or not the worker started", but on Flex Consumption the property is not populated for `az functionapp` at all. #250's broader point stands and is **not** addressed here — a state check still does not prove the isolated worker survived startup validation, which is exactly the failure mode #246 was. That deserves its own design decision (admin-endpoint probe, log query, or a one-shot invocation) rather than being smuggled into a one-line query fix.

Closes #252
Refs #105, #250


kolatts:
**REQUEST CHANGES**

Core fix is right and the split error branch earns its keep. But the PR declares `az functionapp show` untrustworthy for this SKU and then keeps using it, two lines down, for the hostname — and that hostname is already sitting in an environment variable the previous step exported.

---

**Issues**

- **Major** `.github/workflows/_deploy-functions.yml` 134 — `hostname` still reads through the exact projection this PR just stopped trusting, and unlike `state` it has no guard, so if `defaultHostName` ever goes the way of `state` it degrades silently. It is also a redundant round-trip: `.github/scripts/infra/export-outputs.sh` already exports `functionAppHostname` as `FG_FUNCTION_APP_HOSTNAME`, and *Resolve infra outputs* runs before this step. `hostname="$FG_FUNCTION_APP_HOSTNAME"` removes the risk and the API call together.

- **Minor** `.github/workflows/deploy-all.yml` 269–271 — an empty `FUNC_HOST` collapses `${FUNC_HOST:+https://${FUNC_HOST}}` to a blank cell, so a skipped or failed Functions stage renders as an empty link rather than saying so. The API row two lines up already uses `:-n/a`; the other three should agree. Directly downstream of the above: this is what an empty hostname actually costs.

- **Minor** `.github/workflows/_deploy-functions.yml` 132 — `az resource show` without `--api-version` makes the CLI ask the provider for its newest version at run time, so the response shape is chosen by whatever `Microsoft.Web` serves that day. That is a milder form of the same bug being fixed. `properties.state` has been stable across every version, so this is insurance rather than a defect — but the insurance is one flag, and `modules/function-app.bicep` already names the version to use (`2024-04-01`).

- **Minor** `docs-site/src/content/docs/reference/ci-cd.md` 173 — the workflow table still documents the check as "`az functionapp show` state must be `Running`", which this PR makes false. Should have been in the same commit.

- **Nit** `.github/workflows/_deploy-functions.yml` 137–139 — the comment reads as though the `-z` branch is the general "something went wrong" path, but under the default `bash -e` a hard `az` failure exits at the assignment and never reaches it. The branch only catches the narrow "az exits 0 and prints nothing" case, which is exactly the bug — worth saying precisely, since the next person may otherwise trust it to catch auth failures.

- **Nit** `.github/workflows/_deploy-functions.yml` 149 — `az functionapp function list` is the one remaining use of that command family, and it is already `2>/dev/null || true`-guarded with a comment explaining that indexing lags. Correct as-is; noting it so the "no more `az functionapp`" claim isn't overstated.

<sub>Also checked: `--resource-type "Microsoft.Web/sites"` in full `Namespace/type` form is valid (the CLI splits on `/` when `--namespace` is absent); `_postdeployment-tests.yml` re-runs `export-outputs.sh` itself and never consumes `function-app-hostname`, and `functions-deploy.yml` ignores the output, so an empty hostname breaks nothing functionally — only the two summaries and `environment.url`; no other workflow, composite action or script under `.github/` calls `az functionapp` or `az webapp`.</sub>

kolatts:
Fixed in `db78d16`.

**Major — hostname still going through the distrusted projection.** Fair hit, and the inconsistency is the embarrassing part: the PR argued `az functionapp show` can't be trusted for this SKU and then used it two lines later, ungarded. Now `hostname="$FG_FUNCTION_APP_HOSTNAME"` — `export-outputs.sh` already exports it and *Resolve infra outputs* runs first, so this removes the risk and an Azure round-trip in one move. Guarded it as well: an empty hostname is not fatal to the deploy, which is precisely why it's worth failing on — it silently ships a green run with a dead `environment.url` and a blank summary cell.

**Minor — `deploy-all.yml` blank cell.** Fixed for all three rows. Worth flagging that my first attempt at this was wrong: I wrote `${X:+https://${X}}${X:-n/a}`, which *looks* right and isn't — both halves expand when `X` is set, so a working host rendered as `https://h.exampleh.example`. Spelled it out instead and checked all three branches (all set / all empty / mixed) before committing. Left a comment saying why the compact form is not used, since it's an inviting simplification.

**Minor — unpinned `--api-version`.** Taken, at `2024-04-01`, matching `modules/function-app.bicep`. You scored it 65 and I agree it's unlikely to bite — but the argument that it's a milder form of the same bug is exactly right, and the fix is one flag.

**Minor — `ci-cd.md:173`.** Fixed, and it now says *why* rather than just naming the new command, so the next person doesn't "simplify" it back to `az functionapp show`.

**Nit — the `-z` comment overstates its reach.** Correct, and worth being precise about: under the default `bash -e` a hard `az` failure exits at the assignment and never reaches the branch, so it only catches "exits 0, prints nothing". Reworded.

**Nit — `az functionapp function list` remains.** Deliberate. It's already `2>/dev/null || true`-guarded with a comment about indexing lag, which is the right posture for a genuinely unreliable call. Not claiming the file is free of `az functionapp`.

Thanks for checking the downstream consumers — knowing that `_postdeployment-tests.yml` re-resolves outputs itself and `functions-deploy.yml` ignores the output is what made it clear an empty hostname costs two summaries rather than a broken pipeline, which is the right severity for it.

Issue #192, opened 2026-09-02
Live-validate the admin UI pages against a deployed environment and a real Entra tenant


kolatts:
> Parent: #220 — Live validation of the dev environment

## Why

`plans/17-admin-management-ui.md` and `plans/20-foundry-provisioning.md` both carry a
"live validation" box that no automated test can tick, and the plan file pointed at
**#132** for it — which is the APIM key lifecycle and Key Vault wrapping checklist and
covers none of these pages. This is the right issue for that box.

The bUnit suite renders every page against an in-memory client, so it proves the pages
call the right routes with the right arguments and gate the right mutations. It cannot
prove the deployed app signs in, that the role claim arrives in the shape
`[Authorize(Roles = "FoundryGate.Admin")]` expects, or that ARM and Graph behave the way
the fakes do.

## Checklist

Against a deployed environment with a real Entra tenant:

**Auth and routing**
- [ ] A developer (no app role) signing in lands on `/me`; an admin lands on `/dashboard`.
- [ ] A developer navigating directly to `/users`, `/groups`, `/foundry` or `/users/sync`
      gets the AccessDenied page, not a blank grid or a redirect loop.
- [ ] An admin's role claim actually satisfies `RoleNames.Admin` — this is the one that
      silently fails if the app registration emits roles in an unexpected claim type.

**Users (#51, #63)**
- [ ] `/users` pages and searches against a directory with more than one page of people.
- [ ] Deactivating a real user deletes their APIM subscription; their key stops working
      at the gateway.
- [ ] Re-activating mints a new subscription and the developer can reveal the new key.
- [ ] Provision / rotate / revoke each do what the confirmation said they would.
- [ ] `POST /users/sync` against a tenant where developers are assigned **through a
      group**: `skippedGroupAssignmentCount` is `0` on a healthy run, and non-zero with
      the expected copy once `GroupMember.ReadBasic.All` is removed from the identity.

**Groups (#52)**
- [ ] Creating a group linked to a real Entra group, then syncing, pulls the right roster.
- [ ] "Sync all Entra-linked groups" renders a failed group distinctly — force one by
      unassigning the Graph role, and confirm the `GraphRead` wording appears.
- [ ] An Entra-linked group really refuses membership edits (the UI hides them; confirm
      the API's 409 is what backs that).

**Requests (#53)**
- [ ] Approving from the drawer moves the user's tier at the gateway, not just in the UI.
[...]

Issue #139, opened 2026-09-01
Make postdeployment tests gating in deploy-all.yml once Playwright scenarios exist


kolatts:
> Parent: #222 — Owner setup and production readiness

`deploy-all.yml` (#75) runs `_postdeployment-tests.yml` after the api/functions/ui stages, but the run is `continue-on-error: true` and only reports (`n/n passed` in the summary) because `FoundryGate.Tests.Postdeployment` still holds the scaffold smoke test only.

The reusable workflow already passes `FG_API_BASE_URL` (Container App FQDN) and `FG_UI_BASE_URL` (SWA hostname) as environment variables and waits for `/health` before running.

**Done when:**
- Playwright (+ optional Reqnroll) scenarios exist per CONVENTIONS.md §Testing (desktop + mobile viewports, `@LocalOnly` for mutating scenarios, screenshots/traces to TestResults) and read those two variables;
- the `playwright.ps1 install --with-deps` step is added to `_postdeployment-tests.yml`;
- `continue-on-error` is removed and `deploy-all.yml`'s summary job treats a test failure as a failed run;
- consider a `postdeployment-tests` step in `api-deploy.yml` / `ui-deploy.yml` too (today only deploy-all runs them).

PR #99, opened 2026-09-01
feat(api): foundation — Entra auth, options, health, OpenAPI, ProblemDetails (#26, #27)

kolatts:
## Summary

Bootstraps `FoundryGate.Api` on top of #92's data layer (`AddFoundryGateData` already wired) with the infrastructure every feature-endpoint epic will build on:

- **Options + fail-fast validation**: `Configuration/AppSettings.cs` (`AzureAd`, `ConnectionStrings`, `Cors`, `OpenTelemetry`, `Azure.KeyVaultUrl`), bound and validated via `Imagile.Framework.Configuration`'s `ValidateRecursively()` at startup — throws `ConfigurationValidationException` with every violation aggregated, not just the first.
- **Key Vault**: `@KeyVault(SecretName)` reference resolution wired via `Imagile.Framework.Configuration`'s `AppTokenCredential`/`ReplaceKeyVaultReferences`, skipped entirely when `Azure:KeyVaultUrl` is absent — local dev needs zero Azure connectivity.
- **Entra ID bearer auth**: `Microsoft.Identity.Web`'s `AddMicrosoftIdentityWebApiAuthentication`; `PolicyNames.AdminOnly` (role `RoleNames.Admin`, from #91's Domain contracts) registered and ready for the first admin-only controller.
- **CORS**: named policy from `Cors:AllowedOrigins` config, applied globally (see Deviations below).
- **Global error handling**: one `GlobalExceptionHandler : IExceptionHandler` — `KeyNotFoundException`→404, `ArgumentException`→400, new `FoundryGate.Domain.Exceptions.ConflictException`→409, `UnauthorizedAccessException`→403, everything else→500 — all as RFC 7807 `ProblemDetails` with an `X-Correlation-Id` header. `UseStatusCodePages()` covers the 404 catch-all for unmatched routes. No per-controller try/catch.
- **Health checks**: `GET /health` (liveness, anonymous, no dependencies) and `GET /health/ready` (readiness, `AddDbContextCheck<AppDbContext>`) — both return JSON with app version + environment.
- **OpenAPI**: .NET 10's built-in `Microsoft.AspNetCore.OpenApi` (matches imagile-app's precedent, not Swashbuckle) + a `BearerSecuritySchemeTransformer` + a dev-only Scalar UI at `/scalar/v1` (the built-in generator ships no UI of its own).
[...]

PR #258, opened 2026-09-05
fix(api): accept the v2 token audience, and assert 401/403 against a real deployment

### LINKED ISSUE #102: E2E assertions for 401/403 auth behavior on /api/v1 (needs deployed environment)

kolatts:
## The bug this found

The deployed dev API rejected **every token Entra could issue for it**. Live, with a valid admin token:

```
GET /api/v1/users/me
401 Unauthorized
www-authenticate: Bearer error="invalid_token", error_description="The audience '(null)' is invalid"
```

The app registration sets `requestedAccessTokenVersion: 2`, so a v2 access token's `aud` is the bare client id `7e7d0561-…`. `infra/modules/control-plane.bicep` deploys `AzureAd__Audience` as `api://{clientId}`, and setting `Audience` pins Microsoft.Identity.Web's validation to that single string. Nothing in between could ever match. The SPA sign-in in #192 would have hit exactly this.

`EntraAudiences.Resolve` offers both forms — plus whatever `AzureAd:Audience` names, so a fork with a custom application ID URI keeps working — and a `PostConfigure<JwtBearerOptions>` registered after `AddMicrosoftIdentityWebApiAuthentication` makes that the last word. Confirmed by flipping the live Container App's `AzureAd__Audience` to the bare client id: the same request went 401 → 200.

## Why the tests are where they are

`WebApplicationFactory` cannot mint an Entra token, so the predeployment suite can prove the `[Authorize(Policy = AdminOnly)]` attributes are attached and nothing more. It cannot prove the *deployed* configuration accepts a token Entra actually signed — which is the thing that was broken. So the assertions deferred from PR #99 land in `FoundryGate.Tests.Postdeployment`:

- every admin route, no token → **401**
- every admin route, valid token without `FoundryGate.Admin` → **403** (not 401 — a 401 here is precisely the failure above)
- every admin route, admin token → **200**
- developer routes (`/users/me`, `/quota/allocations/me`) → 401 unauthenticated, 200 with either token
- a malformed bearer value → 401, not 500
- `GET /users/me` self-provisions and is idempotent; the key it reports is masked

Configuration comes from `FG_API_BASE_URL` / `FG_ADMIN_TOKEN` / `FG_NONADMIN_TOKEN`. A test whose variable is unset reports as **skipped**, with the reason — never as passed, so an unconfigured run cannot be read as a verified one. `_postdeployment-tests.yml` already exports `FG_API_BASE_URL`, so the unauthenticated half starts running in the pipeline immediately; that is real movement on #139. The token halves stay developer-driven on purpose — a CI principal signs in with OIDC as an application and cannot mint a delegated user token, and an access token expires in an hour, so there is nothing durable to put in a secret.

`EntraAudiencesTests` covers the audience list offline, including the exact deployed shape that failed.

## Docs

- `reference/owner-setup.md` — the `requestedAccessTokenVersion: 2` note now says what it implies for the audience, and a new **§9** documents pre-authorizing the Azure CLI's public client and getting a token. Both are wire-verified: `--resource api://…` fails `Status_InteractionRequired` and `--scope api://…/access_as_user` succeeds, and `preAuthorizedApplications` is replaced wholesale, so the SPA must be listed alongside the CLI.
- `reference/configuration.mdx`, `reference/infrastructure.md` — both audience forms are accepted.

## Verification

- `dotnet test src/FoundryGate.Tests.Predeployment -c Release` — **1617/1617**
- `dotnet test src/FoundryGate.Tests.Postdeployment -c Release` against dev with real tokens — **29/29**; with no variables set — 1 passed, 8 skipped
- `npm run build` in `docs-site/` — clean

Closes #102
Refs #139


kolatts:
**REQUEST CHANGES**

The fix itself is right and the diagnosis is excellent — an audience mismatch that made the whole API unreachable is exactly the class of thing only a live run finds, and the postdeployment suite is the honest place for it. What blocks it is that three docs surfaces still describe the old world, and this repo's CLAUDE.md is unusually explicit that every level has to move together.

---

**Issues**

- **Major** `docs-site/src/content/docs/reference/ci-cd.md` line 175 — Still says "The project holds only the scaffold smoke test, and the summary says so"; this PR deletes that exact sentence from the workflow and rewrites the summary line, so the reference page now describes a workflow that does not exist.

- **Major** `docs-site/src/content/docs/architecture/overview.mdx` line 397 — The project table bills `FoundryGate.Tests.Postdeployment` as "Playwright against a deployed environment, desktop + mobile", which after this PR is an `HttpClient` auth-contract suite with no Playwright and no viewports.

- **Minor** `docs-site/src/components/DeployPipeline.astro` line 70 — `'scaffold smoke test · reporting only until #139'` — the first half is now false, and CONVENTIONS.md asks every wave that changes the pipeline to move the diagram components with it.

- **Minor** `src/FoundryGate.Tests.Postdeployment/DeployedFactAttributes.cs` line 25 — `NonAdminTokenFactAttribute` has no call sites; it's there to make the 3×2 grid symmetric. The other five each earn their keep, this one doesn't.

- **Nit** `src/FoundryGate.Tests.Postdeployment/ApiAuthenticationTests.cs` line 92 — `A_developer_route_with_a_valid_non_admin_token_is_200` on `quota/allocations/me` assumes the non-admin principal already has a `User` row; for a genuinely fresh principal it depends on the `users/me` case running first, which xUnit doesn't promise. Fine today because both tokens are the same person — worth a comment saying so.

- **Nit** `src/FoundryGate.Tests.Postdeployment/ApiAuthenticationTests.cs` line 25 — The comment claims "every admin-only route in the API, one per controller/policy site", but Groups, Requests and the admin half of Keys aren't in the list. Either add them or soften the claim, otherwise the stated safety net reads stronger than it is.

- **Nit** — Two of these tests now write to a live deployment (a `User` row, a `QuotaAllocation`, an APIM subscription) while `reference/ci-cd.md` still calls stage 6 "read-only smoke tests". The token gate keeps them out of CI today, so it isn't urgent, but the new attributes don't express "mutating" at all. Worth an issue rather than a fix here.

---

One question rather than an issue: did you try simply *not* setting `AzureAd__Audience`? Microsoft.Identity.Web defaults `ValidAudiences` to both forms when it's absent, which is the same outcome for free. I assume the answer is that you wanted to keep the knob live for a fork with a custom application ID URI — worth a sentence in `EntraAudiences`' summary if so, because the next reader will ask.

For the record, I did chase the scary version of this and it's clean: with `ValidAudience = null` and an empty `ValidAudiences`, Microsoft.IdentityModel 8.22 falls through to `IDX10214` and rejects, so an empty list fails closed rather than skipping audience validation — and `[Required]` on `ClientId`/`Audience` plus `ValidateRecursively()` means the host never starts in that state anyway.

kolatts:
Point by point. All seven addressed in 0e6199f.

**Major — `reference/ci-cd.md:175` stale.** Fixed. The row now reads "the `/api/v1` 401 contract (#102) plus the scaffold smoke test; the 403/200 half needs a delegated Entra token and reports as *skipped* in CI, and the summary says so". You were right that this is the worst of the three — it is the page someone reads to find out what the pipeline proves.

**Major — `architecture/overview.mdx:397` claims Playwright.** Fixed. Now "xUnit over HTTP against a deployed environment — today the `/api/v1` 401/403/200 auth contract (#102); Playwright scenarios, desktop + mobile, land with #139." That keeps the aspiration on the page without letting it masquerade as the present tense.

**Minor — `DeployPipeline.astro:70`.** Fixed: `'401 auth contract + smoke test · reporting only until #139'`.

**Minor — dead `NonAdminTokenFactAttribute`.** Deleted. Guilty as charged — the sixth cell existed because the grid had six cells.

**Nit — `quota/allocations/me` ordering.** Not a code change, because the fix is to say what the fixture actually requires: the `DeveloperRoutes` summary now states that the route answers 403 for a principal with no `User` row, and that `FG_NONADMIN_TOKEN` must point at an established principal. Making the test self-provision first would hide the ordering rather than remove it, and would add a second writing test to a suite that already has one too many (see below).

**Nit — "every admin-only route" overclaims.** Reworded to "one admin-only GET per controller that has an addressable admin route with no path parameters", naming Groups, Requests and the admin half of Keys as the deliberate omissions and why (they need id fixtures, which is #139's job).

**Nit — mutating tests billed as read-only.** Filed as **#262** and referenced from the two tests that write. The point that made it worth an issue rather than a comment is the one you made: nothing about `AdminTokenFact` says "this writes" — the tests are out of CI by accident of credentials, and #139 removing that accident is exactly when it would bite. #262 asks for a marker that fails safe by default. `ci-cd.md`'s stage table now says what stage 6 does in the meantime.

**Your question — why not just delete `AzureAd__Audience`?** I did consider it, and you are right that Microsoft.Identity.Web then defaults to both forms for free. I kept the knob because a fork can expose a custom application ID URI (`api://foundrygate.contoso.com`) that is derivable from nothing, and honouring it costs one entry in a three-element list. That reasoning is now in `EntraAudiences`' summary, where the next reader will find it before asking.

Thanks for chasing the empty-`ValidAudiences` case to the actual `Validators.ValidateAudience` source rather than assuming — "fails closed via IDX10214" is exactly the sentence I wanted written down somewhere.

Gates after the fix pass: `dotnet build -c Release` 0 warnings, predeployment **1617/1617**, `dotnet format --verify-no-changes` clean, `npm run build` in `docs-site/` clean, `docs-site/src/generated` unchanged (this diff touches no bicep, csproj, timer or controller route).

Issue #261, opened 2026-09-05
Every Function timer invocation on dev fails: Worker.Core 2.52 vs Worker.Grpc 2.1 MissingMethodException


kolatts:
> Parent: #220 — Live validation of the dev environment. Blocks the second half of #178.

## What happened

Every timer invocation on the deployed dev Function App fails, and has been failing since the app was
deployed. From `func-foundrygate-dev-e7k2`'s App Insights:

```
17:45:00.006  Executing 'Functions.UsageSyncFunction' (Reason='Timer fired at 2026-09-05T17:45:00Z')
17:45:00.039  AppExceptions: Method not found:
              'Void Microsoft.Azure.Functions.Worker.DefaultTraceContext..ctor(System.String, System.String)'.
17:45:00.039  Executed 'Functions.UsageSyncFunction' (Failed, Duration=32ms)
```

`AppRequests` for the app, six-hour window — **5 invocations, 5 failures, 0 successes**:

| TimeGenerated | Name | Success | Duration |
|---|---|---|---|
| 17:45:00 | UsageSyncFunction | **False** | 36ms |
| 17:30:00 | UsageSyncFunction | **False** | 48ms |
| 17:15:00 | UsageSyncFunction | **False** | 35ms |
| 17:00:00 | UsageSyncFunction | **False** | 53ms |
| 16:45:00 | UsageSyncFunction | **False** | 39ms |

It dies at *binding* time, ~35ms in, before any user code runs. So `UsageSyncJob` has never executed on
dev, `QuotaAllocation.TokensUsed` has never moved, and no `usage.synced` audit row has ever been
written. `MonthlyQuotaResetFunction` and `EntraSyncFunction` are on the same worker and would fail the
same way.

The failure is silent in every place anyone looks: the host is `Running`, the functions are enabled,
the schedule log prints the next five occurrences correctly, and `/health/ready` on the API is green.

## Root cause

A NuGet resolution split, confirmed with `dotnet list package --include-transitive`:

| | before | after |
|---|---|---|
| `Microsoft.Azure.Functions.Worker` | 2.1.0 | 2.52.0 |
| `Microsoft.Azure.Functions.Worker.Core` | **2.52.0** | 2.52.0 |
| `Microsoft.Azure.Functions.Worker.Grpc` | **2.1.0** | 2.52.0 |

`Microsoft.Azure.Functions.Worker` pins Core **and** Grpc to its own version. But
[...]

Issue #260, opened 2026-09-05
A tier's TPM above the backing deployment's capacity makes the developer's own meter unreachable


kolatts:
> Parent: #220 — Live validation of the dev environment. Sibling of #237 (the TPM *floor*); this is the *ceiling*.

## What the live run showed

Driving dev's `standard` tier (20 000 TPM) with the cycle's burn request, the gateway answered 429 —
but not with its own meter:

```
HTTP 429  Retry-After: 30  x-fg-remaining-tpm: 6413
{ "error": { "message": "Your requests to gpt-4.1-mini for gpt-4-1-mini in eastus2 have exceeded rate limit.", ... } }
```

`x-fg-remaining-tpm` still had **6 413 tokens of headroom** on the refusal. That 429 is Azure OpenAI's
deployment saturating, passed straight through by the OpenAI API policy (which deliberately does not
retry a single backend). `gpt-4-1-mini` was 10 capacity units ≈ **10 000 TPM**, behind a tier whose cap
is **20 000 TPM**.

The consequence is not cosmetic: **the developer's own per-minute budget can never be reached.** Every
throttle a standard-tier developer on dev experienced was the shared deployment, not their allocation —
so `x-fg-remaining-tpm` was permanently misleading, and one developer's traffic could throttle another's
even though the meters are correctly isolated per subscription.

Raising the deployment to 60 capacity units made the gateway's own wall reachable immediately:

```
attempt 1 -> HTTP 200, remaining-tpm=11915
attempt 2 -> HTTP 200, remaining-tpm=3787
attempt 3 -> HTTP 200, remaining-tpm=0
attempt 5 -> HTTP 429, remaining-tpm=0     <- the gateway's meter
```

## The rule

#237 established the floor:

```
tier.tpm  >  one agent turn (~10K for Codex)
```

This is the ceiling, and it is about a resource the tier does not mention:

```
sum(capacity of the deployments an alias routes to) * 1000  >  max(tpm of every tier that can reach it)
```

A tier is well-formed only between the two. Dev shipped outside the second one and nothing said so —
the policy was correct, the deployment was healthy, and the two were simply never compared.

## Suggested work

- [ ] Write the ceiling down next to #237's floor, in `reference/` and in the tier admin copy.
- [ ] `scripts/cycle/smoke.ps1` already reports it: `T4a` now returns SKIP naming both numbers when the
      backend wall sits in front of the gateway wall, instead of a PASS on someone else's 429. Consider
[...]

Issue #259, opened 2026-09-05
dev's gateway 404s on every model: the alias map names gpt-4-1-mini, which no Foundry account has


kolatts:
> Parent: #220 — Live validation of the dev environment

## What happened

Attaching the gateway cycle to **dev** on 2026-09-05, every request through the gateway came back:

```
HTTP 404
{"error":{"type":"invalid_request_error","code":"DeploymentNotFound",
 "message":"The API deployment for this resource does not exist..."}}
```

`fgdev-e7k2-eus2` held exactly one model deployment — `claude-sonnet-4-5`, in **Failed** (#231) — and
`fgdev-e7k2-swc` held none. There was no `gpt-4-1-mini` anywhere, but the alias map named values
(`Gateway__ModelAliases__*`, and the rendered `model-alias-fragment`) advertise `gpt → gpt-4-1-mini`
on all three tiers.

So dev's gateway was answering **404 on every model on every tier**, while every resource in the
environment reported healthy and `Deploy All` was green.

## Why nothing caught it

`infra/parameters/dev.bicepparam` sets `createModelDeployments = false` permanently, with the comment
*"Model lifecycle after day 0 belongs to the control plane, not ARM."* That is the right call for
Anthropic (create-once, E-007). But **nothing** then owns creating the OpenAI deployment: not ARM, and
not the control plane, which has `POST /api/v1/foundry/deployments` but no reconciliation that notices
an alias pointing at a deployment that is not there.

The alias map and the deployment list are two independent facts that are never compared.

## Unblocked by hand

Created out of band, which is the documented-safe exception (E-007e: OpenAI deployments provision
synchronously and reliably; `scripts/cycle/up.ps1` already does exactly this for its own environment):

```sh
az cognitiveservices account deployment create \
  --name fgdev-e7k2-eus2 --resource-group rg-foundrygate-dev \
  --deployment-name gpt-4-1-mini --model-name gpt-4.1-mini --model-version 2025-04-14 \
  --model-format OpenAI --sku-name GlobalStandard --sku-capacity 60
```

Dev has worked since. **No Anthropic create was attempted** (#231, E-007).

## Suggested work

- [ ] A startup or health check that compares the tier alias map against the deployments that actually
      exist on the Foundry accounts, and fails loudly. `GET /health/ready` is the obvious home; it
      currently only checks the database.
- [ ] Decide who owns creating a **missing OpenAI** deployment on a non-day-0 environment. The
[...]

Issue #251, opened 2026-09-05
Tidy pass over scripts/cycle: dedupe Assert-Check and the other deferred review items


kolatts:
Deliberately deferred during the #247 review rather than churning ten scripts that had just been
debugged against live Azure. None of these are correctness bugs; all of them are the kind of thing
that drifts.

## Items

- [ ] **`Assert-Check` is defined three times, byte-identical** — `smoke.ps1`, `codex-test.ps1`
      and `measure.ps1`. It closes over `$state` and `$script:failures`, both of which exist in
      all three, and dot-sourcing puts a `_common.ps1` definition in the caller's scope, so a
      single definition taking `-State` (as `Add-CycleCheck` already does) works unchanged. This
      is the one most likely to drift.
- [ ] **`up.ps1` — `foreach` over a hard-coded single-element array** for the OpenAI
      create-if-missing step. A plain `if` says the same thing in a third of the lines and makes
      the "only OpenAI, never Anthropic" invariant visible rather than implied.
- [ ] **`cycle.ps1` — `Invoke-Stage` returns `$ok` and all six call sites pipe it to
      `Out-Null`.** `$stageResults` is the real channel; drop the return and the noise.
- [ ] **`measure.ps1` — every column is read twice through `Get-KqlValue`**, once to build the
      hashtable and again for the `Write-Host` format string, so each column name is spelled
      twice and the two lists can diverge.
- [ ] **`smoke.ps1` — `New-ChatBody` and `New-MessagesBody` are identical except the default
      `$Model`**, and both already take `-Model`.
- [ ] **`down.ps1` — local `$args` shadows the automatic variable.** Legal at script scope,
      silently wrong the moment those lines move into a function. Rename to `$deleteArgs`.
- [ ] **`up.ps1` — the pre-flight APIM wait/purge partially duplicates the deploy retry loop.**
      The retry loop handles both `ServiceLocked` and `ServiceAlreadyExistsInSoftDeletedState`,
      and the code's own comment says a pre-flight poll cannot see what is about to reject the
      deploy. Decide which layer owns it. (Keep in mind the pre-flight purge *is* load-bearing on
      a clean second run where the deploy would otherwise fail once before retrying.)
- [ ] **`smoke.ps1` — the T5 quota-burn loop only sleeps on 429.** A 401, a 5xx or a transport
      failure falls through with no backoff and no attempt cap for the full 20 minutes, unlike T4
      which is bounded at 30 attempts.
[...]

PR #263, opened 2026-09-05
Run the gateway cycle against dev, and fix the three things that found

kolatts:
The gateway cycle could only ever test a gateway it had just deployed. This teaches it to
**attach** to one that is already there, and then runs it against **dev** — which turned up
three things dev was quietly getting wrong, one of them the reason #178's second half could
never have passed.

## 1. Attach mode

`up.ps1 -AttachOnly` reads the addresses out of the existing `foundrygate-<env>`
subscription deployment's outputs and deploys nothing. The interesting part is the tiers: the
outputs carry `quotaTierRows` but **not** `tpm` — the per-minute cap exists only inside the
rendered `llm-token-limit` element — so the limits are read off the **live product policy**,
which is also the authority, because it is the thing the gateway runs.

`dev` and `prod` are treated as CI-managed (`Test-CycleManagedEnvironment`): `-AttachOnly` is
implied and cannot be turned off, `cycle.ps1` forces `-Teardown None`, and `down.ps1` refuses
outright, pointing at `infra-destroy.yml` with its typed confirmation and environment approval.
`-IKnowThisIsDev` exists only so that overriding is a decision rather than a wall.

Fixture keys are named `fgcycle-*` and `subscriptions.ps1 -Cleanup` removes exactly those —
**by prefix, not from the state file**, so an interrupted run's keys go too. That prefix is
load-bearing: dev's APIM also carries the real `foundrygate-{UserId}` keys the control plane
issues.

## 2. The walls that are out of reach, said as arithmetic

Dev ships the product tiers: standard is 5 000 000 tokens behind a 20 000 TPM cap. Exhausting
that is **250 minutes** of continuous full-rate traffic, several dollars of tokens, and a
counter APIM offers no way to reset — left spent for everyone on that tier until the month
rolls over. So `T5a/b/c` and `C3` report **SKIP by design** with that arithmetic in the detail,
and `T5d` stands in: `x-fg-remaining-quota` read twice and observed decrementing. The wall is
out of reach; the meter is demonstrably live.

Editing dev's policy to shrink a tier would test a gateway nobody runs, so it is not done.

## 3. T4a was passing on somebody else's 429

Two 429s live on that path and `T4a` asserted on neither — it checked only that `Retry-After`
was present. On dev that produced a green check for:

```
HTTP 429  Retry-After: 30  x-fg-remaining-tpm: 6413
{ "error": { "message": "Your requests to gpt-4.1-mini for gpt-4-1-mini in eastus2 have exceeded rate limit." } }
```

6 413 tokens of headroom on the refusal — Azure OpenAI's deployment saturating, not the
developer's budget. `T4b` had always made that distinction; `T4a` now does too, and reports SKIP
naming both numbers when the backend wall sits in front of the gateway wall (#260).

## 4. The Functions timer had never once succeeded (`Closes #261` after verification)

`UsageSyncFunction` fires every 15 minutes on dev and **failed every single time since the app
was deployed** — five for five, ~35ms each, dying at binding time before any user code ran:

```
Method not found: 'Void Microsoft.Azure.Functions.Worker.DefaultTraceContext..ctor(System.String, System.String)'
```

`Microsoft.Azure.Functions.Worker` 2.1.0 pins Core **and** Grpc to 2.1.0;
`Worker.OpenTelemetry` 1.2.0 independently requires Core ≥ 2.52.0. Highest-wins lifted **Core**
to 2.52.0 and left **Grpc** at 2.1.0, and Grpc 2.1.0 calls a constructor Core 2.52.0 removed.
Confirmed both ways with `dotnet list package --include-transitive`. One line, with the coupling
written down so the next `Worker.OpenTelemetry` bump does not re-open it.

The build never said a word: clean, 0 warnings, 1604 tests green — the break is between two
transitive assemblies at runtime, and the host reports `Running` while printing its next five
scheduled occurrences.

## What #178 now has

**The KQL half is proved to the token.** A key issued through the real control-plane flow
(`foundrygate-9`) driven through the gateway, read three independent ways:

| Source | Requests | Tokens |
|---|---|---|
| `x-fg-tokens-consumed` summed | 12 | 1 095 |
| `x-fg-remaining-quota` delta | 12 | 1 095 |
| `UsageBySubscription.kql` | 12 | 1 095 |
| …plus one 54 118-byte request | 13 | 13 124 |

`ApimSubscriptionId` renders as the subscription **name** — the checkbox that decided whether
`ApimSubscriptionNames.TryGetUserId` works at all. The >32 KB request #178 asks for produced a
**single** row (`SequenceNumber = 0`, `TotalTokens = 12029`), so even the shape D-017 is about
does not chunk here.

**The Function half is blocked, not disproved.** Every input the job needs is correct on the
deployed app — the workspace GUID, the enabled timers, and `logAnalyticsDestinationType:
Dedicated` (the #244 fix is deployed, and `UP-3` now checks it on every attach). The job has
simply never executed. `QuotaAllocation.TokensUsed` for user 9 is 0 while the gateway and Log
Analytics both say 13 124, and no `usage.synced` row exists.

So #178 stays open on one verification after the next `Deploy All`.

## Also found on dev, filed not fixed

- **#259** — dev's gateway was answering `404 DeploymentNotFound` on **every model on every
  tier**: the alias map advertises `gpt → gpt-4-1-mini` and no Foundry account had it. Nothing
  owns creating a missing OpenAI deployment after day 0. Unblocked out of band (E-007e, the
  documented-safe exception); no Anthropic create was attempted (#231).
- **#260** — the tier-TPM *ceiling*, sibling of #237's floor: a tier whose TPM exceeds the
  backing deployment's capacity makes the developer's own meter permanently unreachable.

## Gates

- `pwsh scripts/validate-policies.ps1` — clean (14 renders)
- `dotnet build` — 0 warnings; `FoundryGate.Tests.Predeployment` — 1604 passed
- All 9 cycle scripts parse-checked; run live against dev end to end
- `fgcycle-*` fixtures removed; `master` and the real `foundrygate-9` are all that remain on
  dev's APIM. Nothing torn down.

Refs #220, #178, #237, #251
Refs #259, #260, #261


kolatts:
## `sunny:code-review`

**REQUEST CHANGES**

Good PR — the diagnosis work is genuinely excellent and the "say the arithmetic instead of faking the test" instinct is right. But two of the new helpers fail *open* in exactly the situation they exist to prevent, and the docs level this change lives at was skipped.

---

**Issues**

- **Major** `scripts/cycle/_common.ps1` `Get-CycleQuotaBurnMinutes` — `if ($tpm -le 0) { return 0 }` reads as "the monthly wall is reachable *right now*", so the one function whose job is to stop a doomed 20-minute burn against a 5M-token tier starts one instead. Infinity is the fail-safe answer.

- **Major** `scripts/cycle/_common.ps1` `Get-CycleTierFromPolicy` — a regex miss on `tokens-per-minute` silently yields `tpm = 0`, which combined with the above turns "I could not parse this policy" into "burn away". `tokens-per-minute` is mandatory on every rendered `llm-token-limit`, so a miss is never a legitimate "no cap".

- **Major** `scripts/cycle/up.ps1` attach branch — the shape-tolerant unwrap handles the array and the `{value:[...]}` envelope but not the `-AllowFailure` `$null`. `@($null).Count` is 1, so the next `Where-Object` throws under strict mode and aborts the whole attach instead of recording UP-3 as FAIL — and if it didn't throw, the count would report "1 diagnostic setting(s)" for an environment that returned none.

- **Minor** `scripts/cycle/subscriptions.ps1` `-Cleanup` — the list call ignores `nextLink`. ARM pages APIM subscriptions, and dev's service also carries every real `foundrygate-{UserId}` key, so past one page the sweep silently stops finding fixtures and leaves held the counters it exists to release.

- **Minor** `scripts/cycle/subscriptions.ps1` `-Cleanup` — `cycle.ps1`'s `finally` fires this automatically and it deletes *every* `fgcycle-*` on the service, including a concurrent run's freshly stamped fixtures. Two people on dev at once is precisely the scenario attach mode enables; the other run gets 401s mid-stage.

- **Minor** `scripts/cycle/smoke.ps1` new `T5d` — the two probes don't retry, unlike `T4b` and `T5c` which retry precisely because the shared deployment throttles. T5d runs seconds after T4 hammered it with 1500-max-token bodies, so one transient 429 records a FAIL on the check whose whole purpose is not reporting environmental facts as failures.

- **Minor** `docs-site/src/content/docs/reference/infrastructure.md` — CLAUDE.md invariant 2 names `reference/*` as the full-detail level. That section still says the cycle "deploys the template", "tears the gateway down", and that "teardown defaults to `KeepFoundry`" — none of which is true for `dev`/`prod` after this PR. The new mode was documented only in `SKILL.md`, which is agent-facing.

- **Nit** `docs-site/src/pages/index.astro` — says evidence is "one file per cycle"; `validation/README.md` changes to "one file per run, two shapes" in this same diff, and the file added here is explicitly not a spin-up/spin-down cycle.

- **Nit** `scripts/cycle/_common.ps1` — the comment above the `token-quota` regex describes a negative lookahead the pattern doesn't contain. The code is right for a different reason than the comment claims, which is the kind of comment that gets "fixed" into a real bug.

- **Nit** `scripts/cycle/subscriptions.ps1` — a `-Prefix` parameter whose own help text says changing it breaks `-Cleanup`'s guarantee and there is no good reason to. A constant carries the same guarantee with none of the surface. Same file: `Get-ShortName` is a named function wrapping one `-replace` for two call sites, only needed because the table still keys on the old `dev-*` names.

- **Nit** `scripts/cycle/codex-test.ps1` — the C3 skip wraps the 25-line burn loop in an `else`, which forced `$state.codexExecCount = $totalExecs` to be duplicated in both branches.

- **Nit** `scripts/cycle/_common.ps1` / `up.ps1` — the `-DisplayName` parameter, the `displayName`/`description` fields it fills, and the `quotaTierRows` lookup built to feed it are read by nothing; `report.ps1` prints only `name`, `monthlyTokenQuota`, `tpm`.

---

Checked and clean: no secrets leak — `Protect-CycleSecret` still scrubs on the way into the state file, the new SKIP/attach strings carry no key material, and the committed evidence file has only workspace GUIDs, subscription *names* and header counters. The `down.ps1` dev/prod refusal fires before any read or delete and leaves the control-plane guard in force. `fgcycle-` and `foundrygate-` are provably disjoint (checked against `ApimSubscriptionNames.Prefix`). The generated architecture diagram does not need regenerating — `generate-architecture.mjs` reads `ProjectReference` edges, never `Directory.Packages.props`.

kolatts:
## Review response — all twelve addressed in e5c467d

Every fix was re-verified against the live dev gateway, before and after.

### Major

**1. `Get-CycleQuotaBurnMinutes` returned 0 for a zero-TPM tier.** Fixed — `PositiveInfinity`
for either `quota <= 0` or `tpm <= 0`. You are right that this was the worst one: the single
function whose job is to prevent a doomed burn was the one that would start it, and the comment
now says *why* infinity is the answer rather than just what it returns.

**2. `Get-CycleTierFromPolicy` silently yielded `tpm = 0` on a regex miss.** Fixed — it throws
now, naming the product and the reason:

> Could not read tokens-per-minute out of the 'standard' product policy on … The enforcement
> checks are sized from it, so guessing is not an option.

Also worth noting the second blast radius you did not mention: `tpm = 0` collapses T4's burn
body to its 200-repeat floor, so the 429 test would have quietly gone back to hovering.

**3. `up.ps1` unwrap didn't cover the `-AllowFailure` `$null`.** Fixed — three explicit branches
(`$null` → `@()`, envelope, bare array) instead of a two-way ternary. The `@($null).Count == 1`
trap is #251's last bullet arriving in new code within a day of being written down.

### Minor

**4. `-Cleanup` ignored `nextLink`.** Fixed — it pages through the whole list now. Verified on
dev: `-Cleanup` after the run found and removed exactly the three fixtures, and a second sweep
correctly found nothing.

**5. Automatic cleanup would delete a concurrent run's fixtures.** Fixed with a scope split
rather than by weakening the sweep, because both properties are worth keeping:

| | scope | who fires it |
|---|---|---|
| `-Cleanup -OnlyThisRun` | the ids in this state file | `cycle.ps1 -CleanupSubscriptions`, automatically |
| `-Cleanup` | every `fgcycle-*` on the service | a human, deliberately, to catch orphans |

Both verified live on dev.

**6. `T5d` didn't retry.** Fixed — same shape as `T4b`/`T5c`: up to four attempts, honouring
`Retry-After`. It genuinely does run seconds after T4 empties that deployment.

**7. `reference/infrastructure.md` still described a cycle that always deploys and tears down.**
Fair, and the miss was mine — I documented the new mode in `SKILL.md` (agent-facing) and
stopped. That page now has an attach-mode section covering the forced teardown refusal, the
`fgcycle-*` fixtures, the SKIP-by-design arithmetic spelled out, and the backend-vs-gateway 429
distinction. Register kept at full detail, per invariant 2.

### Nits — all taken

**8.** Landing page now says "one file per run".

**9.** The lookahead comment was describing a pattern that isn't there. Rewritten to say the
true reason (`"(\d+)"` is digits-only, so `token-quota-period="Monthly"` cannot match) and to
say that a missing `token-quota` is legitimate while a missing `tokens-per-minute` is not.

**10.** `-Prefix` is now a `$prefix = 'fgcycle'` constant with the reasoning attached — *"a knob
here would turn a guarantee into a convention"*. `Get-ShortName` is gone: the developer table
keys on `alice`/`bob`/`carol` and derives the `dev-$name` state key, which is what the other
stages have always asked for.

**11.** `codex-test.ps1` C3 — the assignment is hoisted out of both branches.

**12.** `-DisplayName`, `displayName`, `description` and the `quotaTierRows` lookup are gone. An
attached tier is now `{ name, monthlyTokenQuota, tpm }`, which is exactly what is read.

---

### One thing I did not do

I left #251's other items alone — `Assert-Check` is now defined three times *and* bypassed twice
by the new SKIP branches, which is that issue's own prediction coming true on schedule. It
belongs in its own PR with a full `test`-gateway cycle as the verification, and I have said so
[on #251](https://github.com/kolatts/foundry-gate/issues/251#issuecomment-5553790976) with an
item-by-item status after this diff.

### Re-verified live after the fixes

```
attempt 1 -> HTTP 200, remaining-tpm=11837
attempt 2 -> HTTP 200, remaining-tpm=3749
attempt 3 -> HTTP 200, remaining-tpm=0
attempt 5 -> HTTP 429, remaining-tpm=0
  [PASS] T4a — TPM cap returns 429 with Retry-After          <- the gateway's own meter
  [PASS] T4b — Same-tier dev-bob unaffected while dev-alice is 429
  [PASS] T5d — Monthly quota counter is live and decrementing
```

Gates: `validate-policies.ps1` clean (14 renders) · `dotnet build` 0 warnings ·
1604 predeployment tests pass · `dotnet format --verify-no-changes` clean ·
`npm run build` in `docs-site/` clean (17 pages) · all 9 cycle scripts parse-check ·
dev's APIM back to `master` + the real `foundrygate-9`, nothing else.

Issue #266, opened 2026-09-05
A Function timer that fails every tick is invisible: every health signal stayed green for hours


kolatts:
> Parent: #220 — Live validation of the dev environment. Split out of #261, which fixed the cause but not this.

## What happened

`UsageSyncFunction` fired on schedule on dev every 15 minutes and **failed every single invocation
since the Function App was deployed** — a `MissingMethodException` at binding time, ~40 ms in, before
any user code ran. It was found by accident, hours later, while gathering evidence for something else
(#178).

Every signal an operator would look at was green the whole time:

| Where you would look | What it said |
|---|---|
| `az functionapp show` | `state: Running` |
| `az functionapp function list` | all three functions, `isDisabled: false` |
| Host startup traces | `Job host started`, `Host started (258ms)` |
| The schedule log | `The next 5 occurrences of the 'UsageSyncFunction' schedule (Cron: '0 0,15,30,45 * * * *') will be:` followed by five correct timestamps |
| API `/health/ready` | `Healthy` (it checks the database, which was fine) |
| `Deploy All` | success, including `Functions - Deploy (dev)` |

The only place the truth existed was `AppRequests | where Success == false`, which nothing was watching.

**A timer that fails every time is indistinguishable from a timer that has nothing to do.** Both write
no business logs, move no data, and leave a healthy-looking host. That is the failure mode worth
engineering against, and it is not specific to this bug: any future exception inside `UsageSyncJob`,
`MonthlyResetJob` or `EntraSyncJob` produces the identical silence.

The blast radius here was real but quiet: `QuotaAllocation.TokensUsed` stayed `0` for every developer
while the gateway and Log Analytics both had the true numbers, so every dashboard and every admin view
of usage was wrong, with nothing anywhere saying so.

## Suggested work

- [ ] **An alert on failed timer invocations.** The minimum viable version is an Azure Monitor alert on
      `AppRequests | where AppRoleName == 'func-…' and Success == false`, at any count > 0 in an hour.
      Cheap, and it would have caught this on the first tick.
- [ ] **Surface sync freshness where it is already being looked at.** The API's `/health/ready` checks
      the database; it could also read the most recent `usage.synced` audit row and degrade when it is
      older than a small multiple of the timer interval. That turns "the reconciliation is dead" into a
      symptom on the endpoint the deployment pipeline already probes.
      - Worth deciding deliberately: the idle guard (D-016) means a *successful* pass writes no audit
[...]

Issue #220, opened 2026-09-05
Live validation of the dev environment


kolatts:
## Why

Nothing in this repo has ever run against a deployed environment. The predeployment suite (1545 tests) proves the code calls the right seams with the right arguments; it cannot prove that ARM, Microsoft Graph, Azure SQL, Key Vault and Log Analytics behave the way the fakes do. Every issue below is a checklist that only a live `dev` deployment can tick, and they are cheapest to run in one session, in dependency order, against one environment.

Run order: **#105 → #142 → #132 → #120 → #125 / #205 → #178 → #192 → #102 → #183**, with #226 whenever #225 has shipped.

## Sub-issues

- [ ] #105 — Live-validate the control-plane Bicep deploy (SQL/Container Apps/SWA/Functions/Key Vault/RBAC)
- [ ] #142 — Live-validate CLI `ip setup` / `ip cleanup` / `db grant-identities` against the dev Azure SQL server
- [ ] #132 — Live-validate APIM key lifecycle and Key Vault key wrapping against real APIM + Key Vault
- [ ] #120 — Live-validate Entra user sync against a real tenant
- [ ] #125 — Live-validate `/foundry` deployment endpoints against a real Foundry account
- [ ] #205 — Live-validate the capacity PATCH, then allow it for Anthropic deployments
- [ ] #178 — Live-validate the reconciliation KQL and both Function timers against a deployed gateway
- [ ] #192 — Live-validate the admin UI pages against a deployed environment and a real Entra tenant
- [ ] #102 — E2E assertions for 401/403 auth behavior on `/api/v1`
- [ ] #183 — Graph: verify `$select`/`$top` on `/transitiveMembers` without `ConsistencyLevel: eventual`
- [ ] #226 — Live validation: PUT `fg-model-map` named value against real APIM + a request routing through a new alias (the live half of #225)

## Human attention

- **#105 — the day-0 deploy itself.** `Deploy All` with `create-model-deployments=true` is a one-time, money-spending, owner-dispatched run, and it cannot start until the `dev` GitHub Environment carries the OIDC variables from the owner-setup parent. Everything downstream is blocked on it.
- **#120 — Graph application-role grants.** `Application.Read.All`, `User.Read.All`, `GroupMember.ReadBasic.All` on **both** `id-foundrygate-api-dev` and `id-foundrygate-func-dev` need a Privileged Role Administrator. Once granted, the whole checklist is agent-runnable (`automated-ok` is on the issue for that reason).
[...]

PR #271, opened 2026-09-05
Record the dev reconciliation path actually running

kolatts:
Docs-only. #263 fixed the Worker/Grpc split that had `UsageSyncFunction` failing every tick on dev, but
it merged before `Deploy All` could land the fix — so the committed evidence file says *"verification is
still outstanding"*. It isn't any more.

`Deploy All` [run 33983870847](https://github.com/kolatts/foundry-gate/actions/runs/33983870847) landed
the new build at 18:44Z. The very next tick:

| TimeGenerated | Success | Duration |
|---|---|---|
| 17:45 / 18:00 / 18:15 / 18:30 | **False** | 37 / 42 / 61 / 71 ms |
| **18:45** | **True** | **2 943 ms** |
| 19:00 | True | 5 512 ms |

The duration is its own evidence — tens of milliseconds is an exception at binding time, seconds is the
job actually querying Log Analytics and writing to SQL.

**Four readings of the same traffic, agreeing to the token** for `foundrygate-9`, the key issued through
the real control-plane flow:

| Source | Tokens |
|---|---|
| `x-fg-tokens-consumed` headers, summed | 13 124 |
| `x-fg-remaining-quota` delta | 13 124 |
| `UsageBySubscription.kql` | **13 220** |
| `QuotaAllocation.TokensUsed`, reconciled | **13 220** |

(The last two include six requests the pipeline's own postdeployment tests made in between.) The
database value was `0` an hour earlier, and both moved together again on the next pass, to `13 252`.

One `usage.synced` row, `allocationsUpdated: 3, unknownSubscriptions: 6` — exactly the three real
`foundrygate-*` keys and the six `fgcycle-*` cycle fixtures, so `ApimSubscriptionNames.TryGetUserId`
partitions real developers from everything else correctly on real data. That was the mapping question
#178 existed to answer.

## What this deliberately does not claim

**The idle guard.** The 19:00 pass wrote a *second* audit row — but with `allocationsUpdated: 1`, i.e.
something genuinely moved, which is D-016 working rather than D-016 being tested. Dev has had traffic on
every pass so far, so "two hours of ticks with no traffic produce ~1 row, not 8" stays an open item on
#178. The report says so in those words.

Also still unproved and named as such: `MonthlyQuotaResetFunction` and `EntraSyncFunction` are unblocked
(same worker) but unobserved on their daily/nightly schedules, and #178's sections 3 and 4 are untouched.

## Filed alongside

**#266** — the finding worth more than the version bump. Five consecutive failed invocations produced no
alert and no symptom anywhere an operator looks: host `Running`, functions enabled, schedule log printing
its next five occurrences, `/health/ready` green, and `Deploy All` — postdeployment tests included —
passing, while `QuotaAllocation.TokensUsed` sat at `0` for every developer. A timer that fails every tick
is indistinguishable from a timer that has nothing to do.

Refs #178, #220, #266

### DIFF
--- a/validation/2026-09-05-dev-gateway.md
+++ b/validation/2026-09-05-dev-gateway.md
@@ -162,8 +162,9 @@ Highest-wins lifted **Core** to 2.52.0 and left **Grpc** at 2.1.0, and Grpc 2.1.
 Core 2.52.0 removed. The build was clean, 0 warnings, all tests green — the break is between two
 transitive assemblies at runtime.
 
-Fixed here by bumping `Microsoft.Azure.Functions.Worker` to 2.52.0 so the two move together.
-**Verification that reconciliation then runs is still outstanding** — it needs the next `Deploy All`.
+Fixed by bumping `Microsoft.Azure.Functions.Worker` to 2.52.0 so the two move together, and **verified
+live**: the first tick after `Deploy All` landed the new build succeeded, and reconciliation ran end to
+end. Numbers below.
 
 ## #178 — the reconciliation path, end to end
 
@@ -196,9 +197,9 @@ Three independent readings of the same traffic agreeing to the token. Against #1
   checkbox is satisfied. `Entries > 1` remains unobserved; the de-duplication is defending against
   something this workspace has not yet produced.
 
-### The Function half: blocked, not disproved
+### The Function half: blocked for hours, then proved in one tick
 
-Everything the job needs is correct on the deployed app:
+Everything the job needs was already correct on the deployed app:
 
 - `Gateway__LogAnalyticsWorkspaceId` = `062aeba1-151c-44c8-a554-1456d6d9fcc8` — the **GUID**, not the
   ARM id (which is separately present as `Gateway__LogAnalyticsWorkspaceResourceId`). The host starts,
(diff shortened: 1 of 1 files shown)


kolatts:
## `sunny:code-review`

**APPROVE**

Docs-only, and it does the hard thing: it draws a sharp line between what the tick actually proved and what it didn't. The `allocationsUpdated: 1` catch — refusing to bank the idle guard because the pass wasn't idle — is exactly the discipline that makes the rest of the file worth trusting.

---

**Issues**

None blocking. Two things I checked and want on the record:

- **Nit** `validation/2026-09-05-dev-gateway.md` — the four-reading table mixes measurement times (13 124 at 18:00, 13 220 at 18:45/18:47) and leans on prose underneath to explain the gap. The prose does explain it, and the timestamps are now in the row labels, so this is fine — but the reason the numbers differ is a third party's traffic, and that is the kind of detail a reader six months out will trip on before they reach the sentence explaining it.

- **Nit** The file's header still says the sections after *Reproducing this* "would be lost by a re-render". Still true, and still worth keeping, but this is now the second commit to hand-edit that region — if a third arrives, the hand-written half probably wants to be its own file that `report.ps1` can never clobber.

Checked and clean: no secrets — the audit JSON, the workspace GUID, the subscription *names* and the token counters are all non-sensitive, and no key material appears. Every number in the diff is reproducible from the queries quoted next to it. No `src/` change, so no build/test surface, and no behavioural claim that needs a code change to back it. Docs invariants: this is `validation/`, which the landing page links as a folder and does not enumerate, so invariants 1, 2, 3 and 5 are untouched; invariant 4 (`docs-site` build) is unaffected by a file outside `docs-site/`.

Issue #254, opened 2026-09-05
what-if reports 39 phantom changes on a clean re-run — the no-op invariant cannot be checked


kolatts:
Filed against the last unticked box on #105 — *"Re-run of the same template is a no-op (`what-if` shows no changes)"*. It is **not** a no-op, and the reason is worth writing down: the **deploy** is idempotent, but `what-if` is not honest about it.

Third consecutive `Deploy All → dev` on an unchanged template ([run 33979330341](https://github.com/kolatts/foundry-gate/actions/runs/33979330341)) reports:

```
Resource changes: 6 to create, 33 to modify, 21 no change, 15 unsupported, 1 to ignore.
```

Every one of those 39 is phantom. Two distinct causes, both verified against the live resources.

## 1. `reference()` cannot be evaluated at plan time — 33 "modify"

Every APIM backend and API shows a diff like:

```
~ .../apis/foundrygate-anthropic
  ~ properties.serviceUrl:
      "https://fgdev-e7k2-eus2.cognitiveservices.azure.com/anthropic"
   => "[format('{0}anthropic', reference('.../accounts/fgdev-e7k2-eus2', '2026-07-01').endpoint)]"
```

The left side is the deployed value; the right side is the **unevaluated ARM expression**. `what-if` cannot resolve `reference()` before the deployment runs, so it compares a string against a template expression and calls it a change. Verified the deployed value is unchanged and correct:

```
az apim api show --api-id foundrygate-anthropic --query serviceUrl
-> https://fgdev-e7k2-eus2.cognitiveservices.azure.com/anthropic
```

This is a known ARM `what-if` limitation, not drift. It affects every resource whose property is built from `reference().endpoint` — the two APIs, their policies, and all four Foundry backends.

## 2. `products/apis` has no readable representation — 6 "create"

The six product↔API associations are reported as `+ Create` on every run:

```
+ .../products/{standard,power,unlimited}/apis/{foundrygate-anthropic,foundrygate-openai}
```

They exist:

```
GET .../products/standard/apis -> foundrygate-anthropic, foundrygate-openai
```

`Microsoft.ApiManagement/service/products/apis` is an association, not a resource with a comparable GET body, so `what-if` has nothing to diff and defaults to "create". The PUT is a genuine no-op (200, verified in #239/#240 with two consecutive PUTs) — which is precisely why this resource type was chosen over `apiLinks`. The irony is that the fix for a non-idempotent resource produced a permanently noisy `what-if`.

[...]

Issue #205, opened 2026-09-02
Live-validate the capacity PATCH, then allow it for Anthropic deployments


kolatts:
> Parent: #220 — Live validation of the dev environment

Follow-up to #130, which shipped `PATCH /api/v1/foundry/deployments/{accountName}/{deploymentName}/capacity` for **OpenAI-format deployments only**.

## Why Anthropic is refused today

The refusal is not a limitation like the create/delete ones (#107/#126) — those are blocked by real, observed limits. This one is blocked purely because nobody has run a capacity PATCH against a live Claude deployment.

The reasoning says it is safe, and it is a good argument:

- The ARM operation is `Deployments_Update` — **`PATCH .../deployments/{name}`**, verified against the published REST reference and against `Azure.ResourceManager.CognitiveServices` 1.5.2 (`CognitiveServicesAccountDeploymentResource.UpdateAsync(WaitUntil, PatchResourceTagsAndSku, ct)`; the assembly references `RequestMethod.Patch` exactly once, for this request builder).
- Its request body schema is **`{ sku, tags }`**. There is no `model` field and no `modelProviderData` field, so the call *structurally cannot* re-send the model or its Marketplace attestation. That is what makes it a different animal from the `CreateOrUpdate` PUT that drove a Claude deployment to `Failed` (fable-refactor-log E-006/E-007).

But E-007's lesson is precisely that Claude deployments punish assumptions, however sound, so the code refuses Anthropic with a `400` whose message says why.

## Ask

1. Live-validate on an **OpenAI** deployment first: `PATCH .../capacity` with a new value, confirm `sku.capacity` changes, `provisioningState` settles on `Succeeded`, and the model/version/name are untouched.
2. Then on a **Claude** deployment in a non-production subscription: same checks, plus confirm the deployment still serves traffic through the gateway afterwards and the subscription's Marketplace agreement is undisturbed.
3. Record the finding as an E-00x entry in `fable-refactor-log.md`.
4. If it holds: delete the `AnthropicFormat` guard in `FoundryDeploymentService.UpdateCapacityAsync` (one `if`), update its test to assert the resize instead of the refusal, and update `reference/api.md`. If it does not: keep the guard and write down what actually happened, which is more valuable than the feature.

Related live-validation work: #132, #125, #88.

Issue #125, opened 2026-09-01
Live-validate /foundry deployment endpoints against a real Foundry account (OpenAI create/409/delete, multi-account list)


kolatts:
> Parent: #220 — Live validation of the dev environment

Follow-up to #61 (PR on branch `claude/foundry-deployments-61`). The `/api/v1/foundry/*` endpoints are fully covered by hermetic tests (ARM behind `IFoundryManagementClient` with an in-memory fake — no live Azure in tests, per CLAUDE.md), so the ARM seam itself (`ArmFoundryManagementClient` over `Azure.ResourceManager.CognitiveServices` 1.5.2) has never been exercised against a real Foundry account. Needs the control plane deployed (#105) with `Gateway__SubscriptionId` / `Gateway__ResourceGroup` / `Gateway__FoundryAccountNames__{i}` set (#108) and the API identity holding **Cognitive Services Contributor** on each account.

## Manual checklist (rg-foundrygate-dev, OpenAI-format only — Claude creation is refused by design, see #107)

- [ ] `GET /api/v1/foundry/deployments` as admin lists every deployment in **every** configured account (both regions), primary account first; SKU/capacity/provisioningState/createdDate populated; matches `az cognitiveservices account deployment list`.
- [ ] `GET /api/v1/foundry/models` as a non-admin developer lists each pooled Claude deployment **once**, `Succeeded`.
- [ ] `POST /api/v1/foundry/deployments` `{ accountName: <primary>, deploymentName: "gpt-4-1-mini-live", modelFormat: "OpenAI", modelName: "gpt-4.1-mini", modelVersion: "2025-04-14", skuName: "GlobalStandard", capacity: 5 }` → `201`, `Location` header resolves, body `provisioningState` is `Creating`/`Accepted`/`Succeeded`; portal shows the deployment; `GET` on the Location reaches `Succeeded` within a minute.
- [ ] Repeat the same `POST` → `409` with the "already exists" detail; portal shows **no** change to the deployment (no re-PUT: `provisioningState` stays `Succeeded`, `systemData.lastModifiedAt` unchanged).
- [ ] `POST` with `modelFormat: "Anthropic"` → `400` mentioning #107; **no** deployment appears in the portal (E-007: a failed Claude create can wedge the subscription).
- [ ] `POST` with an `accountName` not in `Gateway__FoundryAccountNames` → `400`.
[...]

PR #273, opened 2026-09-05
fix(infra): let ARM own the OpenAI deployments, and report the tier capacity ceiling

### LINKED ISSUE #259: dev's gateway 404s on every model: the alias map names gpt-4-1-mini, which no Foundry account has

### LINKED ISSUE #260: A tier's TPM above the backing deployment's capacity makes the developer's own meter unreachable

kolatts:
Owns **#259** and **#260**, the two blind spots the dev gateway cycle found.

## #259 — who owns a missing OpenAI deployment

dev answered `404 DeploymentNotFound` on every model on every tier for its whole first day, while every resource reported healthy and `Deploy All` was green. `createModelDeployments = false` on every day-N run, so ARM did not create `gpt-4-1-mini`; the control plane can create a deployment but never reconciles one. Nobody owned it.

**The create-once rule is Anthropic's, not ARM's.** Verified against the live dev account: PUT the same OpenAI deployment body three times —

```
create   provisioningState Succeeded   createdAt 19:09:46.373   lastModifiedAt 19:09:46.373
re-PUT   provisioningState Succeeded   createdAt 19:09:46.373   lastModifiedAt 19:09:53.990
re-PUT   provisioningState Succeeded   createdAt 19:09:46.373   lastModifiedAt 19:09:55.923
```

No Conflict, no `InternalServerError`, no drive to `Failed` — the opposite of what E-007 records for Claude. (Probe deployment created and deleted for this; nothing else touched.)

So the guard in `foundry.bicep` now reads `if (d.format != 'Anthropic' || createAnthropicModelDeployments)`: ARM keeps the OpenAI deployments present on every run, and the flag covers only the deployments whose re-PUT is actually dangerous. Renamed to `createAnthropicModelDeployments` because a flag called `createModelDeployments` that does not control OpenAI deployments is a trap. The **workflow input keeps its name** (`create-model-deployments`) and its meaning — only the template parameter and the descriptions change.

## #260 — the tier capacity ceiling

#237 established the floor: a tier's TPM must exceed one agent turn. This is the ceiling, about a resource the tier does not mention:

```
sum(capacity of the deployments an alias routes to) x 1000  >=  TPM of every tier that reaches it
```

dev's `gpt-4-1-mini` was 10 units — 10,000 TPM — behind tiers allowing up to 100,000. Every throttle a developer saw was the shared deployment saturating, with `x-fg-remaining-tpm` still showing headroom on the refusal. Their own meter was unreachable, and one developer's traffic throttled another's despite the meters being correctly isolated per subscription.

Two changes:

- **Default capacity 10 → 100** for `gpt-4-1-mini`, matching the highest tier that reaches it. GlobalStandard bills per token, so capacity is a rate ceiling rather than a reservation — raising it costs nothing until the tokens are spent, and the subscription's quota for this model is 5,000 units. This also retires the by-hand `60` set during the live run: ARM now converges dev to the parameter (#260's last bullet).
- **`main.bicep` computes the shortfalls** from the parameters actually in use (`modelCapacityWarnings`), and `deploy.sh` prints one `::warning::` per offending `(tier, alias)`. **Reported, never fatal.** A shortfall is a capacity decision, and for Claude a create-once one — so the Claude defaults are deliberately left alone and simply named by the warning rather than raised by an agent. #205 owns changing a live Claude deployment's capacity.

## The offline guard

`InfraModelDeploymentTests` reads `infra/main.bicep` (the same technique `GatewayOptionsModelAliasesTests` already uses) and asserts two things about the shipped defaults:

1. every alias in `productModelAliases` names a deployment `pooledModelDeployments` or `primaryOnlyModelDeployments` declares — the comparison whose absence *is* #259;
2. every **OpenAI** alias's deployment can serve the TPM of each tier that reaches it, multiplying capacity by the region count for pooled deployments.

Reverting the capacity to 10 fails the second one with the exact sentence #260 was filed over:

> Tier 'standard' allows 20000 TPM, but alias 'gpt' routes at OpenAI deployment 'gpt-4-1-mini', which can serve 10000 TPM (10 capacity units). Developers on that tier are throttled by the deployment before their own meter is ever reached (#260).

## What to expect on the next deploy

The first run after this merge will PATCH dev's `gpt-4-1-mini` from 60 to 100 capacity units and, from then on, re-PUT it unchanged each run. `what-if` will list the OpenAI deployment as a change on every run — a real change once, a no-op after, and one more entry on #254's list of things `what-if` reports that the deploy does not do.

## Verification

- `az bicep build infra/main.bicep` — clean (only the pre-existing `BCP081` warnings for the 2026-07-01 CognitiveServices types)
- `dotnet test src/FoundryGate.Tests.Predeployment -c Release` — **1619/1619**
- `npm run build` in `docs-site/` — clean; `docs-site/src/generated/architecture.json` regenerated (capacity 10 → 100)
- `bash -n .github/scripts/infra/deploy.sh` — clean

Docs moved at every level the change touches: `reference/infrastructure.md` (both re-run invariants, plus a new **capacity ceiling** section beside #237's floor), `reference/ci-cd.md`, `reference/owner-setup.md`, `getting-started/fork-guide.mdx`, `architecture/overview.mdx`.

Closes #259
Closes #260


kolatts:
**REQUEST CHANGES**

The core call is right and the live re-PUT evidence settles it properly — but renaming a template parameter and only updating one of the two callers breaks the gateway cycle, and there's a trail of docs and one operator-facing error message still naming the old flag.

---

**Issues**

- **Major** `scripts/cycle/up.ps1` line 383 — Still passes `createModelDeployments=…` to `az deployment sub create`, and `main.bicep` no longer declares it, so every cycle `up` (and `-WhatIf`, which reuses the same arg list) now fails `InvalidTemplate`. This is the second real deployment call site in the repo and CI never touches it.

- **Major** `scripts/cycle/up.ps1` lines 468 and 481 — The out-of-band OpenAI create is justified by "that flag is all-or-nothing, so an account that came up without its OpenAI deployment can never get one back from the template", which this PR makes false. Worse, if it ever fires it creates `gpt-4-1-mini` at `capacity = 10` — the exact condition #260 exists to eliminate.

- **Major** `.claude/skills/gateway-cycle/SKILL.md` lines 87–90, 270, 303, 369 — `reference/infrastructure.md` points at this file as the authority on the create-once rules, and it still names the old parameter twice and still frames the capacity fix as outstanding work when this PR lands it.

- **Major** `CLAUDE.md` line 76 (and the `AGENTS.md` mirror) — "re-runs must pass `createModelDeployments=false`" sits in Architecture ground truths, the section every agent reads first, and now names a parameter that does not exist and a rule this PR deliberately narrowed.

- **Major** `src/FoundryGate.Api/Services/Foundry/FoundryDeploymentService.cs` line 185 — The 400 an admin gets when they try to create a Claude deployment tells them to use `createModelDeployments=true`. That is an operator instruction on the wire; following it now fails. Same string in `IFoundryDeploymentService.cs` and `reference/api.md` line 436.

- **Minor** `docs-site/src/components/DeployPipeline.astro` line 44 — "model deployments create-once", unqualified. Only the Claude ones are, now.

- **Minor** `.github/scripts/infra/deploy.sh` lines 44–60 — The reporting block is the script's last command under `set -euo pipefail`, so anything it raises turns a green deployment into a failed step. `python3` is preinstalled on `ubuntu-latest` so it doesn't bite today, but a `.format(**r)` `KeyError` the moment the output shape changes would report a successful deploy as failed. An `|| true` costs nothing.

- **Nit** — The Claude aliases are all below the ceiling on the shipped defaults, so the new warning channel ships permanently saturated with nine rows nobody can action without #205. A warning that always fires is a warning nobody reads; consider grouping by deployment rather than by `(tier, alias)`.

---

For the record, the parts I chased hardest are clean: the loop condition is right in both directions (the true-branch set is unchanged, the false-branch set strictly grows by OpenAI-only entries — no Claude deployment gets re-PUT that didn't before), `toObject`'s duplicate-key throw is unreachable with unique names across the two parameters, the region multiplication is correct against the module loop, and `az bicep build` is clean apart from the pre-existing `BCP081`s. `docs-site/src/generated/architecture.json` regenerated correctly and moved exactly the two `gpt-4-1-mini` entries.

One question rather than an issue: `_deploy-api.yml` calls the same `deploy.sh`, so an API deploy now also PUTs `gpt-4-1-mini` at 100 units. If the subscription's GlobalStandard quota for that model were under 100, a previously-inert resource becomes a hard blocker on the API deploy path. Worth a sentence in the capacity comment saying what the headroom actually is — the one above it documents the Claude quotas and says nothing about OpenAI.

kolatts:
Point by point. All eight addressed in 0fe4904.

**Major — `up.ps1:383` passes the deleted parameter.** Fixed, and thank you — this was a genuine break with nothing in CI to catch it. `up.ps1` is the repo's *other* `az deployment sub create`, and I renamed the one in `deploy.sh` without grepping for siblings. `-WhatIf` reuses the same arg list, so both paths were dead. While in there I renamed the prose in `up.ps1`, `down.ps1` and `status.ps1` and the `$state.createModelDeploymentsUsed` field (nothing else reads it), so `grep -rn createModelDeployments` over `scripts/ .claude/ CLAUDE.md AGENTS.md docs-site/src infra .github` is now empty except for one deliberate historical mention in `foundry.bicep`'s comment explaining what changed. `up.ps1` parse-checks clean.

**Major — `up.ps1:468/481` create-if-missing is contradictory and hardcodes capacity 10.** Fixed both halves. The comment now says what the block is for *after* this change — belt and braces for the paths that skip or partially apply the template (an interrupted deploy, a hand-deleted deployment, a cycle attached to an environment someone else deployed) — and the capacity matches `primaryOnlyModelDeployments`, so the out-of-band path can no longer recreate the exact condition #260 was filed over. Good catch that it would have quietly reintroduced the bug this PR removes.

**Major — `gateway-cycle/SKILL.md`.** All four spots. `:270` and `:303` renamed and scoped; the `404 DeploymentNotFound` remediation row now says "re-run the template first" instead of "create the OpenAI one out of band", which is the whole behavioural change; `:87-90` and the `T4a` row no longer frame the capacity fix as outstanding, and the `T4a` row now distinguishes OpenAI (raise it) from Claude (stop — that is #205's call, not a cycle's).

**Major — `CLAUDE.md:76` / `AGENTS.md:77`.** Fixed in both mirrors, and they now carry the scoping sentence too, since Architecture ground truths is where the next agent forms its model of this.

**Major — the API's 400 body.** Fixed in `FoundryDeploymentService.cs`, `IFoundryDeploymentService.cs` and `reference/api.md`. You were right that this is the sharpest of the doc issues: it is an instruction we hand an operator mid-failure, and following it would have failed. `api.md`'s create-once bullet now also says the 409 is the *API's* rule rather than ARM's, since this PR is precisely the discovery that ARM does not have one for OpenAI.

**Minor — `DeployPipeline.astro:44`.** Now "Claude deployments create-once".

**Minor — `deploy.sh` can fail a green deploy.** Fixed: `|| true` on the pipeline and a broad `except` inside the renderer that emits a `::notice::` instead of raising. You are right that `python3` is fine on `ubuntu-latest` today and that the `KeyError` is the one that bites later; either way, a reporting step has no business failing a deployment that succeeded.

**Nit — the warning ships permanently saturated.** Acted on, because I think you understated it. Grouped by **deployment** instead of `(tier, alias)`: the shipped Claude defaults are below the ceiling on three tiers each, so the old shape emitted nine rows on every deploy saying four things. Each row now names the deployment, the TPM it can serve, the tiers that fail against it, and the highest TPM routed at it. Still four rows on the defaults — three Claude and none for OpenAI — but four actionable ones, and #205 is the issue that clears three of them.

**Your question — the OpenAI quota headroom.** Fair, and I had the number and did not write it down. The subscription's `gpt-4.1-mini` GlobalStandard quota is **5,000 units** — read live from an `InsufficientQuota` error during the #125 validation, which helpfully reports the limit — so 100 is 2% of it. That is now in the comment beside the capacity, along with the point you raised: `_deploy-api.yml` calls the same `deploy.sh`, so an over-large value on a smaller fork's quota fails the API deploy and not just infra.

Gates after the fix pass: `az bicep build` clean (only the pre-existing `BCP081`s), `dotnet build -c Release` 0 warnings, predeployment **1619/1619**, `dotnet format --verify-no-changes` clean, `npm run build` in `docs-site/` clean, `bash -n deploy.sh` clean, `up.ps1` parses clean.

